PR #5, opened 2025-05-27
feat: add "network run" command

ghost:
Adds a command, `icp network run`, which runs the local network in the foreground until the user presses Ctrl-C or otherwise sends SIGINT.

Adds a test that dfx can deploy to this network.

Everything is hardcoded and there are various rough edges. As discussed, this is intended to be a starting point for further iteration and collaboration.

### DIFF
--- a/bin/icp-cli/src/project/structure.rs
+++ b/bin/icp-cli/src/project/structure.rs
@@ -28,8 +28,12 @@ impl ProjectDirectoryStructure {
         self.root.join("networks").join(format!("{name}.yaml"))
     }
 
+    fn work_dir(&self) -> PathBuf {
+        self.root.join(".icp")
+    }
+
     pub fn network(&self, network_name: &str) -> NetworkDirectoryStructure {
-        let network_root = self.root.join(".networks").join(network_name);
+        let network_root = self.work_dir().join("networks").join(network_name);
 
         NetworkDirectoryStructure::new(&network_root)
     }
--- a/lib/icp-network/src/config/model/connected.rs
+++ b/lib/icp-network/src/config/model/connected.rs
@@ -1,5 +1,7 @@
 use serde::Deserialize;
 
+/// A "connected network" is a network that we connect to but don't manage.
+/// Typical examples are mainnet or testnets.
 #[derive(Deserialize)]
 pub struct ConnectedNetworkModel {
     // /// The URL(s) this network can be reached at.
--- a/lib/icp-network/src/structure.rs
+++ b/lib/icp-network/src/structure.rs
@@ -29,7 +29,7 @@ impl NetworkDirectoryStructure {
     }
 
     pub fn pocketic_dir(&self) -> PathBuf {
-        self.network_root.join(".pocketic")
+        self.network_root.join("pocket-ic")
     }
 
     // pocketic expects this file not to exist when launching it.
--- a/lib/icp-network/src/config/model/managed.rs
+++ b/lib/icp-network/src/config/model/managed.rs
@@ -1,5 +1,6 @@
 use serde::{Deserialize, Deserializer};
 
+/// A "managed network" is a network that we start, configure, stop.
 #[derive(Deserialize, Default)]
 pub struct ManagedNetworkModel {
     pub gateway: GatewayModel,


bin/icp-cli/src/project.rs
```diff
@@ -0,0 +1 @@
+pub mod structure;
```
rikonor, line 1: Should this be put under `bin/icp-cli/src/project/mod.rs` instead?

bin/icp-cli/src/project/structure.rs
```diff
@@ -0,0 +1,36 @@
...
+        &self.root
+    }
+
+    #[allow(dead_code)]
+    pub fn network_config_path(&self, name: &str) -> PathBuf {
+        self.root.join("networks").join(format!("{name}.yaml"))
+    }
+
+    pub fn network(&self, network_name: &str) -> NetworkDirectoryStructure {
+        let network_root = self.root.join(".networks").join(network_name);
```
rikonor: Any concerns with a proliferation of `.<name>` directories popping up over time? What about simply nesting those under a single `.icp` directory, e.g, `.icp/networks`?

lib/icp-fs/src/fs.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+#[derive(Snafu, Debug)]
+#[snafu(display("failed to write {}", path.display()))]
+pub struct WriteFileError {
+    pub path: PathBuf,
+    pub source: std::io::Error,
+}
+
+pub fn write<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Result<(), WriteFileError> {
```
rikonor, line 57: What is the reasoning behind these thin wrappers around the std library functions?

lib/icp-fs/src/json.rs
```diff
@@ -0,0 +1,42 @@
...
+        source: serde_json::Error,
+    },
+
+    #[snafu(transparent)]
+    Write { source: WriteFileError },
+}
+
+pub fn save_json_file<T: Serialize>(path: &Path, value: &T) -> Result<(), SaveJsonFileError> {
+    let content = serde_json::to_string_pretty(&value).context(SerializeSnafu { path })?;
+    crate::fs::write(path, content)?;
```
rikonor, line 40: Same question - what is the reasoning behind having these thin wrappers?

lib/icp-network/src/config/model/connected.rs
```diff
@@ -0,0 +1,7 @@
+use serde::Deserialize;
+
+#[derive(Deserialize)]
+pub struct ConnectedNetworkModel {
```
rikonor, line 6: Can you add a comment explaining what a "connected network" is?

lib/icp-network/src/managed/pocketic/admin.rs
```diff
@@ -0,0 +1,145 @@
...
+    AutoProgressConfig, CreateHttpGatewayResponse, CreateInstanceResponse, ExtendedSubnetConfigSet,
+    HttpGatewayBackend, HttpGatewayConfig, HttpGatewayInfo, InstanceConfig, InstanceId, RawTime,
+    SubnetSpec, Topology,
+};
+use reqwest::Url;
+use snafu::prelude::*;
+use std::path::Path;
+use time::OffsetDateTime;
+
+pub struct PocketIcAdminInterface {
```
rikonor, line 11: Is the reason you are creating `PocketIcAdminInterface` due to `pocket_ic` not accepting an http client in its initializer? It might be simpler to ask Martin to support taking an optional transport.

lib/icp-network/src/managed/pocketic/native.rs
```diff
@@ -0,0 +1,19 @@
+use std::path::Path;
+
+pub fn spawn_pocketic(pocketic_path: &Path, port_file: &Path) -> tokio::process::Child {
+    let mut cmd = tokio::process::Command::new(pocketic_path);
+    cmd.arg("--port-file");
+    cmd.arg(port_file.as_os_str());
+    cmd.args(["--ttl", "2592000", "--log-levels", "error"]);
+
+    cmd.stdout(std::process::Stdio::inherit());
+    cmd.stderr(std::process::Stdio::inherit());
```
rikonor, line 10: Is this supposed to use `Stderr`?

lib/icp-network/src/structure.rs
```diff
@@ -0,0 +1,42 @@
+use std::path::{Path, PathBuf};
+
+pub struct NetworkDirectoryStructure {
```
rikonor, line 3: Is `Structure` an equivalent term to `manifest`?

bin/icp-cli/src/project.rs
ghost: No, this is using the newer approach of (in this case) having src/project.rs rather than src/project/mod.rs

bin/icp-cli/src/project/structure.rs
ghost: I think `.icp/networks` would be fine

lib/icp-fs/src/fs.rs
ghost: These add `path`, in order to avoid reporting errors like "OS error 2: file not found" without telling which path wasn't found.

And they do it in one place, so other error types can include these as `transparent` rather than each defining their own error with std::io::Error as a source, path as context, and error text.

lib/icp-fs/src/json.rs
ghost: These also add `path` context, and report a single error type that differentiates between i/o errors and json parse errors.

lib/icp-network/src/managed/pocketic/admin.rs
ghost: No, it's because the access library in pocket_ic does a bunch of things under the hood that we don't want, such as downloading the pocket-ic binary in some conditions and panicking in others. It's more geared to running tests than a user-facing application.

lib/icp-network/src/managed/pocketic/native.rs
ghost: Thanks for noticing, I didn't. I actually brought this over from https://github.com/dfinity/sdk/blob/master/src/dfx/src/actors/pocketic_proxy.rs#L259 or https://github.com/dfinity/sdk/blob/master/src/dfx/src/actors/pocketic.rs#L249. I think in this context, stdout is okay.

bin/icp-cli/src/project.rs
rikonor: Having both `project.rs` and a `project/` directory feels a bit inconsistent — usually, it's cleaner to pick one module style so the structure is more intuitive at a glance.

lib/icp-network/src/structure.rs
ghost: My intent with this struct is to provide a single, focused place to describe the structure of a network directory by returning known paths. 

We could consider renaming it to something like `NetworkLayout` or `NetworkFilesystemLayout`.

Regarding the term "manifest", I generally understand that to mean a populated list of specific contents. For example, a Cargo.toml file is a manifest: a list of packages and dependencies. A list of the fields one might find in a Cargo.toml would be more of a schema. Similarly, a shipping or flight manifest is a list of passengers or cargo. If such a manifest hasn't been filled in yet, we might call it a template rather than a manifest.

While one could argue that a known set of paths constitutes "contents" and thus a kind of manifest, I think that’s a stretch. We're describing the structure or layout of a directory. The contents are what's inside the files, not the locations of the files themselves.

bin/icp-cli/src/project.rs
ghost: This follows the [newer Rust module convention](https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames), which avoids mod.rs in favor of more descriptive filenames like `project.rs` alongside a `project/` directory for submodules.

I agree that having both a file and a directory with the same name looks a bit odd at first, but it's the idiomatic structure under the newer approach.

Since this is the first PR introducing submodules, I went with the newer style. Happy to discuss if we want to standardize on the older convention instead.

lib/icp-network/src/config/model/connected.rs
rikonor: Would "remote network" be a better terminology here?

ghost: Not necessarily. It could refer to another network that happens to be running locally, like another project.

rikonor: From the project's point of view, wouldn't that network be remote?

PR #6, opened 2025-05-28
feat: Identity creation

adamspofford-dfinity:
Three components:
- `icp-identity/`, for managing identities
- `icp identity import` and `icp identity new` for creating identities
- `icp identity default`/`--identity`/`icp identity principal`, required for useful tests

Plus scaffolding:
* `Env` for global/cached state
* `icp-dirs/` for project-wide agreement on what the dirs are
* Error reporting and JSON output (but not JSON errors yet)

### DIFF
--- a/lib/icp-dirs/src/lib.rs
+++ b/lib/icp-dirs/src/lib.rs
@@ -34,6 +34,7 @@ impl IcpCliDirs {
 pub enum DiscoverDirsError {
     #[snafu(display("user directories are non-UTF-8"))]
     NonUtf8 { source: camino::FromPathBufError },
+
     #[snafu(display("home directory could not be located"))]
     CannotFindHome,
 }
--- a/bin/icp-cli/src/main.rs
+++ b/bin/icp-cli/src/main.rs
@@ -29,6 +29,7 @@ async fn main() -> Result<(), ProgramError> {
 pub enum ProgramError {
     #[snafu(transparent)]
     Dispatch { source: DispatchError },
+
     #[snafu(transparent)]
     Dirs { source: DiscoverDirsError },
 }


bin/icp-cli/src/commands/identity/import.rs
```diff
@@ -0,0 +1,283 @@
...
+pub struct ImportCmd {
+    name: String,
+    #[arg(long, group = "import-from")]
+    from_pem: Option<Utf8PathBuf>,
+    #[arg(long, group = "import-from")]
+    read_seed_phrase: bool,
+    #[arg(long, group = "import-from")]
+    from_seed_file: Option<Utf8PathBuf>,
+    #[arg(long, requires = "from_pem")]
+    decryption_password_from_file: Option<Utf8PathBuf>,
```
ghost: Some whitespace in between these parameters will make them easier to read.

```suggestion
    name: String,

    #[arg(long, group = "import-from")]
    from_pem: Option<Utf8PathBuf>,

    #[arg(long, group = "import-from")]
    read_seed_phrase: bool,

    #[arg(long, group = "import-from")]
    from_seed_file: Option<Utf8PathBuf>,

    #[arg(long, requires = "from_pem")]
    decryption_password_from_file: Option<Utf8PathBuf>,
```

bin/icp-cli/src/commands/identity/default.rs
```diff
@@ -0,0 +1,46 @@
+use clap::Parser;
+use parse_display::Display;
+use serde::Serialize;
+use snafu::{Snafu, ensure};
+
+use crate::env::Env;
+
+#[derive(Parser)]
+pub struct DefaultCmd {
+    name: String,
```
ghost: Suggestion: `icp identity default` prints the name of the default identity. Can be a follow-up.
```suggestion
    name: Option<String>,
```

bin/icp-cli/src/commands/identity/default.rs
```diff
@@ -0,0 +1,46 @@
...
+    #[snafu(transparent)]
+    WriteDefaults {
+        source: icp_identity::WriteIdentityError,
+    },
+    #[snafu(transparent)]
+    LoadList {
+        source: icp_identity::LoadIdentityError,
+    },
+    #[snafu(display("no identity found with name `{name}`"))]
+    NoSuchIdentity { name: String },
```
ghost: Some whitespace will help to make these more readable.
```suggestion
    #[snafu(transparent)]
    WriteDefaults {
        source: icp_identity::WriteIdentityError,
    },

    #[snafu(transparent)]
    LoadList {
        source: icp_identity::LoadIdentityError,
    },

    #[snafu(display("no identity found with name `{name}`"))]
    NoSuchIdentity { name: String },
```

Cargo.toml
```diff
@@ -1,9 +1,37 @@
 [workspace]
-members = [
-    "bin/icp-cli",
-    "lib/icp-identity",
-    "lib/icp-network",
-]
+members = ["bin/icp-cli", "lib/icp-dirs", "lib/icp-identity", "lib/icp-network"]
```
ghost: Was this an autoformat thing? I put these on separate lines because I think they are easier to read that way.

bin/icp-cli/src/commands/identity.rs
```diff
@@ -0,0 +1,37 @@
...
+    subcmd: IdentitySubcmd,
+}
+
+#[derive(Subcommand)]
+pub enum IdentitySubcmd {
+    Import(import::ImportCmd),
+    New(new::NewCmd),
+    Principal(principal::PrincipalCmd),
+    List(list::ListCmd),
+    Default(default::DefaultCmd),
```
ghost: Nit: Could we alphabetize these subcommands? It might make it easier to scan and add new ones consistently.

Same goes for the dispatch method.

bin/icp-cli/src/commands/identity/import.rs
```diff
@@ -0,0 +1,290 @@
...
+    AssociatedOid, EncryptedPrivateKeyInfo, ObjectIdentifier, PrivateKeyInfo, SecretDocument,
+    der::{Decode, pem::PemLabel},
+};
+use sec1::{EcParameters, EcPrivateKey};
+use serde::Serialize;
+use snafu::{OptionExt, ResultExt, Snafu, ensure};
+use std::{fs, io};
+
+use crate::env::Env;
+
```
ghost: I recommend against splitting up `use` statements like this. We used to on dfx and moved away from it. Inevitably the IDE doesn't get the message and inconsistencies either slip through or must be addressed in CR.

bin/icp-cli/src/commands/identity/import.rs
```diff
@@ -0,0 +1,290 @@
...
+            env,
+            &cmd.name,
+            &from_pem,
+            cmd.decryption_password_from_file.as_deref(),
+        )?;
+    } else if let Some(path) = &cmd.from_seed_file {
+        let phrase = fs::read_to_string(path).context(ReadSeedFileSnafu { path })?;
+        import_from_seed_phrase(env, &cmd.name, &phrase)?;
+    } else if cmd.read_seed_phrase {
+        let phrase = Password::new()
```
ghost, line 56: This should prompt for the seed phrase twice and make sure the same seed phrase was entered both times.

lib/icp-identity/src/lib.rs
```diff
@@ -1 +1,334 @@
...
+    let list = match fs::read_to_string(&id_list_file) {
+        Ok(id_list) => serde_json::from_str(&id_list).context(ParseJsonSnafu {
+            path: &id_list_file,
+        })?,
+        Err(e) if e.kind() == ErrorKind::NotFound => {
+            let empty_list = IdentityList {
+                v: 1,
+                identities: HashMap::new(),
+            };
+            write_identity_list(dirs, &empty_list).context(WriteDefaultsSnafu)?;
```
ghost: Suggestion: `icp identity list` should not have a side effect of writing an empty identity list.
```suggestion
```

bin/icp-cli/src/commands/identity/list.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use crate::env::Env;
+
+#[derive(Parser)]
+pub struct ListCmd;
+
+pub fn exec(env: &Env, _cmd: ListCmd) -> Result<ListKeysMessage, ListKeysError> {
+    let list = icp_identity::load_identity_list(env.dirs())?;
+    let mut identities = icp_identity::special_identities();
+    identities.extend(list.identities.into_keys());
```
ghost: Suggestion: extract these three lines into a method in the icp-identity crate

bin/icp-cli/src/commands/identity/new.rs
```diff
@@ -0,0 +1,71 @@
...
+    #[arg(long)]
+    output_seed: Option<Utf8PathBuf>,
+}
+
+pub fn exec(env: &Env, cmd: NewCmd) -> Result<NewIdentityMessage, NewIdentityError> {
+    let mnemonic = Mnemonic::new(MnemonicType::for_key_size(256).unwrap(), Language::English);
+    let path = DEFAULT_DERIVATION_PATH.parse().unwrap();
+    let seed = Seed::new(&mnemonic, "");
+    let pk = XPrv::derive_from_path(seed.as_bytes(), &path).context(DerivationSnafu)?;
+    let key = SecretKey::from(pk.private_key());
```
ghost: Mnemonic, default derivation path, Seed, XPrv, SecretKey - all this stuff belongs in the icp-identity crate, not here in the commands.

In this case, maybe something like IdentityKey::generate_secp256k1()

lib/icp-dirs/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+    Standard(ProjectDirs),
+    Overridden(PathBuf),
+}
+
+impl IcpCliDirs {
+    pub fn new() -> Self {
+        if let Some(override_var) = std::env::var_os("ICP_CLI_HOME") {
+            Self::Overridden(override_var.into())
+        } else {
+            Self::Standard(ProjectDirs::from("org.dfinity", "", "icp-cli").unwrap())
```
ghost: Can we return an error here rather than panicking? I think we can get away without `impl Default`.

lib/icp-dirs/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+
+#[derive(Debug, Clone)]
+pub enum IcpCliDirs {
+    Standard(ProjectDirs),
+    Overridden(PathBuf),
+}
+
+impl IcpCliDirs {
+    pub fn new() -> Self {
+        if let Some(override_var) = std::env::var_os("ICP_CLI_HOME") {
```
ghost: The name of the binary is `icp`, not `icp-cli`
```suggestion
        if let Some(override_var) = std::env::var_os("ICP_HOME") {
```

lib/icp-identity/src/lib.rs
```diff
@@ -1 +1,334 @@
...
+    }
+}
+
+pub fn write_identity_defaults(
+    dirs: &IcpCliDirs,
+    defaults: &IdentityDefaults,
+) -> Result<(), WriteIdentityError> {
+    let defaults_path = identity_defaults_path(dirs);
+    let parent = defaults_path.parent().unwrap();
+    fs::create_dir_all(parent).context(CreateDirectorySnafu { path: parent })?;
```
ghost: This is probably going to come up again and again. One possible solution that avoids unwraps (even though these will never panic), since `identity_defaults_path` already knows the parent:
let defaults_path = ensure_identity_defaults_path(dirs)?; // create_dir_all the parent and return the dir

lib/icp-identity/src/lib.rs
```diff
@@ -1 +1,334 @@
...
+    DecodePrivateKey, EncodePrivateKey, EncryptedPrivateKeyInfo, PrivateKeyInfo,
+    der::pem::PemLabel, pkcs5::pbes2::Parameters,
+};
+use rand::RngCore;
+use scrypt::Params;
+use sec1::der::Decode;
+use serde::{Deserialize, Serialize};
+use snafu::{OptionExt, ResultExt, Snafu, ensure};
+use zeroize::Zeroizing;
+
```
ghost: There's too much in this file. You can probably come up with something better, but here are some suggestions:
model.rs
storage.rs or persistence.rs or config.rs (just for managing the json files; paths can go at the top)
pem.rs or secp256k1.rs
create.rs
load.rs

lib/icp-identity/src/lib.rs
```diff
@@ -1 +1,334 @@
...
+    #[snafu(display("failed to read password: {message}"))]
+    GetPasswordError { message: String },
+}
+
+#[derive(Debug, Snafu)]
+pub enum WriteIdentityError {
+    #[snafu(display("failed to write to file `{}`", path.display()))]
+    WriteFileError { path: PathBuf, source: io::Error },
+    #[snafu(display("failed to create directory `{}`", path.display()))]
+    CreateDirectoryError { path: PathBuf, source: io::Error },
```
ghost: These are examples of where I think the icp-fs DRY methods help

lib/icp-identity/src/lib.rs
```diff
@@ -1 +1,334 @@
...
+                    Params::new(17, 8, 1, 32).expect("valid scrypt params"),
+                    &salt,
+                    &iv,
+                )
+                .expect("valid pbes2 params"),
+                password,
+            )
+            .expect("infallible PKI encryption")
+            .to_pem(EncryptedPrivateKeyInfo::PEM_LABEL, Default::default())
+            .expect("infallible EPKI encoding")
```
ghost: Extract method

lib/icp-identity/src/lib.rs
```diff
@@ -1 +1,334 @@
...
+    let spec = match format {
+        CreateFormat::Plaintext => IdentitySpec::Pem {
+            format: PemFormat::Plaintext,
+            algorithm,
+        },
+        CreateFormat::Pbes2 { .. } => IdentitySpec::Pem {
+            format: PemFormat::Pbes2,
+            algorithm,
+        },
+    };
```
ghost: a little DRY
```suggestion
    let pem_format = match format {
        CreateFormat::Plaintext => PemFormat::Plaintext,
        CreateFormat::Pbes2 { .. } => PemFormat::Pbes2,
    };
    let spec = IdentitySpec::Pem {
        format: pem_format,
        algorithm,
    };
```

bin/icp-cli/src/commands/identity/import.rs
adamspofford: Are you sure? That sounds like the flow for creating a password, not entering it, and seed phrases are checksummed so you can't miss a word. We don't do that for principals.

lib/icp-identity/src/lib.rs
```diff
@@ -1 +1,334 @@
...
+        .identities
+        .get(name)
+        .context(NoSuchIdentitySnafu { name })?;
+    match identity {
+        IdentitySpec::Pem { format, algorithm } => {
+            let pem_path = key_pem_path(dirs, name);
+            let pem = fs::read_to_string(&pem_path).context(ReadFileSnafu { path: &pem_path })?;
+            let doc = pem
+                .parse::<Pem>()
+                .context(ParsePemSnafu { path: &pem_path })?;
```
ghost: extract method

lib/icp-identity/src/lib.rs
```diff
@@ -1 +1,334 @@
...
+                    match algorithm {
+                        IdentityKeyAlgorithm::Secp256k1 => {
+                            let key = k256::SecretKey::from_pkcs8_encrypted_der(
+                                doc.contents(),
+                                &password,
+                            )
+                            .context(ParseKeySnafu { path: &pem_path })?;
+                            Ok(Arc::new(Secp256k1Identity::from_private_key(key)))
+                        }
+                    }
```
ghost: extract method

lib/icp-identity/src/lib.rs
```diff
@@ -1 +1,334 @@
...
+                        doc.tag() == PrivateKeyInfo::PEM_LABEL,
+                        "internal error: wrong identity format found"
+                    );
+                    match algorithm {
+                        IdentityKeyAlgorithm::Secp256k1 => {
+                            let key = k256::SecretKey::from_pkcs8_der(doc.contents())
+                                .context(ParseKeySnafu { path: &pem_path })?;
+                            Ok(Arc::new(Secp256k1Identity::from_private_key(key)))
+                        }
+                    }
```
ghost: extract method

lib/icp-identity/src/lib.rs
```diff
@@ -1 +1,334 @@
...
+    match fs::read_to_string(&id_defaults_path) {
+        Ok(id_defaults) => Ok(serde_json::from_str(&id_defaults).context(ParseJsonSnafu {
+            path: &id_defaults_path,
+        })?),
+        Err(e) if e.kind() == ErrorKind::NotFound => {
+            let empty_defaults = IdentityDefaults {
+                v: 1,
+                default: "anonymous".to_string(),
+            };
+            write_identity_defaults(dirs, &empty_defaults).context(WriteDefaultsSnafu)?;
```
ghost: Recommendation: if loading configuration and there is no configuration file, don't create one, and also don't create its parent directories.
```suggestion
```

bin/icp-cli/src/commands/identity/import.rs
ghost: No, I'm not sure. But as far as I can tell, `icp identity import --read-seed-phrase` is the only way to create an identity from a previously-known seed phrase, so in that way, it is like creating a password.

And while seed phrases are checksummed, and you can't miss one word, you can miss 3 or 6 or 9 or 12. For example, both of these seed phrases have valid checksums:
```
rather exhibit number that skirt cheap layer rude unaware twist more kitchen dry hill laundry solid universe tell march snap useless moral cry tuition
rather exhibit number that skirt cheap layer rude unaware twist more kitchen dry hill laundry solid universe tell march snap useless
```
(Obviously, the search space is small if you made the above mistake)

To me, this is closer to creating or changing a password. While someone can just copy once and paste twice, they can also choose to enter their very important seed phrase twice to be sure.

bin/icp-cli/src/commands/identity/import.rs
```diff
@@ -0,0 +1,292 @@
...
+                    ],
+                    found: sections.iter().map(|s| s.tag().to_string()).collect_vec(),
+                }
+                .fail()?
+            } else {
+                TooManyKeyBlocksSnafu { count, path }.fail()?
+            }
+        }
+    };
+    let decrypted_doc: SecretDocument;
```
rikonor: This variable is declared up-front but only used in one code branch. Should it be defined at the point of initialization instead?

bin/icp-cli/src/commands/identity/list.rs
```diff
@@ -0,0 +1,45 @@
...
+use std::fmt::{self, Display, Formatter};
+
+#[derive(Debug, Parser)]
+pub struct ListCmd;
+
+pub fn exec(env: &Env, _cmd: ListCmd) -> Result<ListKeysMessage, ListKeysError> {
+    let list = icp_identity::load_identity_list(env.dirs())?;
+    let defaults = icp_identity::load_identity_defaults(env.dirs())?;
+    Ok(ListKeysMessage {
+        identities: list.to_valid_identity_names(),
```
ghost: Suggestion: a single method that returns all of the identities, including anonymous

bin/icp-cli/src/commands/identity/default.rs
```diff
@@ -0,0 +1,45 @@
...
+use snafu::{Snafu, ensure};
+
+#[derive(Debug, Parser)]
+pub struct DefaultCmd {
+    name: String,
+}
+
+pub fn exec(env: &Env, cmd: DefaultCmd) -> Result<DefaultIdentityMessage, DefaultIdentityError> {
+    let list = icp_identity::load_identity_list(env.dirs())?;
+    ensure!(
```
ghost:
```
$ target/debug/icp identity list
  abc
  abcd
* anonymous
$ target/debug/icp identity default anonymous
Error: no identity found with name `anonymous`
```

bin/icp-cli/src/commands/identity/default.rs
```diff
@@ -0,0 +1,45 @@
...
+        list.identities.contains_key(&cmd.name),
+        NoSuchIdentitySnafu { name: &cmd.name }
+    );
+    let mut defaults = icp_identity::load_identity_defaults(env.dirs())?;
+    defaults.default = cmd.name.clone();
+    icp_identity::write_identity_defaults(env.dirs(), &defaults)?;
```
ghost: Suggestion: extract to method in the icp-identity crate

bin/icp-cli/src/commands/identity/new.rs
```diff
@@ -0,0 +1,68 @@
...
+use parse_display::Display;
+use serde::Serialize;
+use snafu::{ResultExt, Snafu};
+use std::{fs, io};
+
+#[derive(Debug, Parser)]
+pub struct NewCmd {
+    name: String,
+    #[arg(long)]
+    output_seed: Option<Utf8PathBuf>,
```
ghost: The help for this outputs like this, from which it's unclear this is a file
```
      --output-seed <OUTPUT_SEED>      
```

bin/icp-cli/src/commands/identity/default.rs
adamspofford: Hmm. Maybe special-casing this is a mistake.

bin/icp-cli/src/commands/identity/import.rs
adamspofford: der structures borrow from their buffer, so the buffer has to be in scope wherever they are; the code branch that writes to it is the one code branch where the buffer is created inside the match scope. If it were declared where it was written to, you would get an error about dangling references.

Cargo.toml
adamspofford: Yes, this was format-on-save.

lib/icp-identity/src/lib.rs
adamspofford: This implies the inverse of your other comment, though, of not implicitly creating the directories.

ghost: I don't think that has to be the case:
- going to write to the file? call ensure_identity_defaults_path(dirs)
- going to read the file? call identity_defaults_path(dirs) and check if exists

adamspofford: :shrug:

lib/icp-identity/src/manifest.rs
```diff
@@ -0,0 +1,162 @@
...
+
+pub fn write_identity_defaults(
+    dirs: &IcpCliDirs,
+    defaults: &IdentityDefaults,
+) -> Result<(), WriteIdentityError> {
+    let defaults_path = identity_defaults_path(dirs);
+    let parent = defaults_path.parent().unwrap();
+    fs::create_dir_all(parent)?;
+    let json = serde_json::to_string(defaults).unwrap();
+    fs::write(&defaults_path, json.as_bytes())?;
```
ghost:
```suggestion
    save_json_file(&defaults_path, defaults)?;
```

bin/icp-cli/src/commands/identity/import.rs
```diff
@@ -0,0 +1,315 @@
...
+    );
+    SecretKey::from_slice(epk.private_key).context(BadPemKeySnafu { path })
+}
+
+fn import_from_seed_phrase(env: &Env, name: &str, phrase: &str) -> Result<(), DeriveKeyError> {
+    let mnemonic = Mnemonic::from_phrase(phrase, Language::English).context(ParseMnemonicSnafu)?;
+    let path = DEFAULT_DERIVATION_PATH.parse().unwrap();
+    let seed = Seed::new(&mnemonic, "");
+    let pk = XPrv::derive_from_path(seed.as_bytes(), &path).context(DerivationSnafu)?;
+    let key = SecretKey::from(pk.private_key());
```
ghost: These four lines are duplicated verbatim in new.rs. I think it would be reasonable for the icp-identity crate to have methods that create a key usable as an identity both random and from a seed, and that the derivation path could live there rather than here with the commands.

bin/icp-cli/src/commands/identity.rs
```diff
@@ -0,0 +1,53 @@
...
+        IdentitySubcmd::Principal(subcmd) => env.print_result(principal::exec(env, subcmd))?,
+    }
+    Ok(())
+}
+
+const DEFAULT_DERIVATION_PATH: &str = "m/44'/223'/0'/0/0";
+
+#[derive(Debug, Snafu)]
+pub enum IdentityCommandError {
+    #[snafu(transparent)]
```
ghost, line 39: Whitespace in between these enum variants will enhance readability.

bin/icp-cli/src/commands/identity/default.rs
```diff
@@ -0,0 +1,61 @@
...
+use snafu::Snafu;
+
+#[derive(Debug, Parser)]
+pub struct DefaultCmd {
+    name: Option<String>,
+}
+
+pub fn exec(env: &Env, cmd: DefaultCmd) -> Result<DefaultCmdMessage, DefaultIdentityError> {
+    if let Some(name) = cmd.name {
+        let list = icp_identity::manifest::load_identity_list(env.dirs())?;
```
ghost: Can DRY the first line of both code paths

adamspofford: No, you cannot.

bin/icp-cli/src/commands/identity/import.rs
```diff
@@ -0,0 +1,315 @@
...
+    },
+
+    #[snafu(display("malformed key in PEM file `{path}`"))]
+    BadPemContent {
+        path: Utf8PathBuf,
+        source: pkcs8::der::Error,
+    },
+
+    #[snafu(display("incomplete key in PEM file `{path}`: {info}"))]
+    BadPemKeyStructure { path: Utf8PathBuf, info: String },
```
ghost: All of the "bad pem key structure" errors seem to be "missing field X". Is there code yet to some that has some other kind of bad pem key structure?

bin/icp-cli/src/commands/identity/default.rs
```diff
@@ -0,0 +1,61 @@
...
+        icp_identity::manifest::change_default_identity(env.dirs(), &list, &name)?;
+        Ok(DefaultCmdMessage::Change(ChangeDefaultIdentityMessage {
+            default: name,
+        }))
+    } else {
+        let defaults = icp_identity::manifest::load_identity_defaults(env.dirs())?;
+        Ok(DefaultCmdMessage::Display(DisplayDefaultIdentityMessage {
+            default: defaults.default,
+        }))
+    }
```
ghost: The signal-to-noise ratio is low here. It might help to adjust imports and DRY where possible.
```suggestion
    let dirs = env.dirs();
    if let Some(name) = cmd.name {
        let list = load_identity_list(dirs)?l;
        change_default_identity(dirs, &list, &name)?;
        Ok(DefaultCmdMessage::Change(ChangeDefaultIdentityMessage {
            default: name,
        }))
    } else {
        let defaults = load_identity_defaults(dirs)?;
        Ok(DefaultCmdMessage::Display(DisplayDefaultIdentityMessage {
            default: defaults.default,
        }))
    }
```

bin/icp-cli/src/error.rs
```diff
@@ -0,0 +1,62 @@
+use std::{
+    error::Error,
+    fmt::{self, Display, Formatter},
+};
+
+use snafu::FromString;
+
+#[derive(Debug)]
+pub struct AnyError {
```
ghost: Since nothing references this, I'd suggest removing it for now.

lib/icp-fs/src/fs.rs
```diff
@@ -25,6 +25,11 @@ pub fn read<P: AsRef<Utf8Path>>(path: P) -> Result<Vec<u8>, ReadFileError> {
     std::fs::read(path).context(ReadFileSnafu { path })
 }
 
+pub fn read_to_string<P: AsRef<Utf8Path>>(path: P) -> Result<String, ReadFileError> {
```
ghost: There are different things that can go wrong when reading to a string (invalid UTF-8 for example), so let's use a different error type and message.

```suggestion
pub fn read_to_string<P: AsRef<Utf8Path>>(path: P) -> Result<String, ReadToStringError > {
```

lib/icp-identity/src/key.rs
```diff
@@ -0,0 +1,198 @@
...
+            let key =
+                k256::SecretKey::from_pkcs8_der(doc.contents()).context(ParseKeySnafu { path })?;
+            Ok(Arc::new(Secp256k1Identity::from_private_key(key)))
+        }
+    }
+}
+
+pub fn key_pem_path(dirs: &IcpCliDirs, name: &str) -> Utf8PathBuf {
+    dirs.identity_dir().join(format!("keys/{name}.pem"))
+}
```
ghost: Please consider consolidating these types of methods into one place. I've done this with ProjectDirectoryStructure and NetworkDirectoryStructure, but maybe you have a better idea.

Those structs exist to provide a single focused place to look and see how the files are laid out. I think it's less effective to have one standalone method in the middle of this file, and a couple more methods in the middle of manifest.rs.

lib/icp-identity/src/manifest.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+#[derive(Debug, Snafu)]
+pub enum ChangeDefaultsError {
+    #[snafu(transparent)]
+    Load { source: LoadIdentityError },
+    #[snafu(transparent)]
+    Write { source: WriteIdentityError },
+    #[snafu(display("no identity found with name `{name}`"))]
+    NoSuchIdentity { name: String },
```
ghost: wall of text

lib/icp-identity/src/key.rs
```diff
@@ -0,0 +1,198 @@
...
+        IdentityKey::Secp256k1(key) => key.to_pkcs8_der().expect("infallible PKI encoding"),
+    };
+    let pem = match format {
+        CreateFormat::Plaintext => doc
+            .to_pem(PrivateKeyInfo::PEM_LABEL, Default::default())
+            .expect("infallible PKI encoding"),
+        CreateFormat::Pbes2 { password } => make_pkcs5_encrypted_pem(&doc, &password),
+    };
+    let pem_path = ensure_key_pem_path(dirs, name).map_err(WriteIdentityError::from)?;
+    fs::write(&pem_path, pem.as_bytes()).map_err(WriteIdentityError::from)?;
```
ghost: Is this a place where Snafu goes out of its way not to allow `?` to "just work"?  How is writing `.map_err(WriteIdentityError::from)?` better?

adamspofford: There's two levels of conversion here - WriteFileError to WriteIdentityError to CreateIdentityError. `?` only goes through one.

PR #7, opened 2025-05-29
feat: initial skeleton for build command

rikonor:


lib/icp-canister/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+pub struct Adapter {
+    #[serde(rename = "type")]
+    pub typ: AdapterType,
+}
+
+/// Known adapter types that can be used to build a canister.
+/// These correspond to the values found in `build.adapter.type` in the YAML.
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum AdapterType {
```
ghost: I think it's possible to do away with the above `Adapter`, and make the enum itself the adapter
```suggestion
#[serde(tag = "type", rename_all = "lowercase")]
pub enum Adapter {
```

lib/icp-canister/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+    /// A canister written in Rust.
+    Rust,
```
ghost:
```suggestion
    Rust(RustAdapter),
```

lib/icp-project/src/lib.rs
```diff
@@ -38,3 +39,16 @@ impl ProjectDirectoryStructure {
         NetworkDirectoryStructure::new(&network_root)
     }
 }
+
+/// Represents the manifest for an ICP project, typically loaded from `icp.yaml`.
+/// A project is a repository or directory grouping related canisters and network definitions.
+#[derive(Debug, Deserialize)]
+pub struct ProjectManifest {
```
ghost: I'd suggest to put this in `manifest.rs` and leave `structure.rs` as-is. ProjectDirectoryStructure is meant to be a concise place to look to see the directory structure of a project.

lib/icp-canister/src/lib.rs
rikonor: I understand what you're proposing, but I am not sure how I would implement that. Do you have an idea? We need to ensure it still supports the expected yaml format:
```
name: <name>
build:
  adapter:
    type: <type>
```

bin/icp-cli/src/commands/build.rs
```diff
@@ -0,0 +1,68 @@
...
+    Ok(())
+}
+
+#[derive(Debug, Snafu)]
+pub enum BuildCommandError {
+    #[snafu(display("no project (icp.yaml) found in current directory or its parents"))]
+    ProjectNotFound,
+
+    #[snafu(display("failed to load project manifest: {source}"))]
+    ProjectLoad { source: ReadFileError },
```
adamspofford: Errors do not need `: {source}`, the error formatter removes this.

lib/icp-project/src/lib.rs
rikonor: Sure, I can put structure.rs back.

bin/icp-cli/src/commands/build.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    #[snafu(display("failed to parse project manifest: {source}"))]
+    ProjectParse { source: ProjectManifestError },
```
adamspofford: This ends up looking like 'failed to parse project manifest: failed to parse project manifest: <real error>'. Ideally this should be `transparent`.

lib/icp-canister/src/lib.rs
rikonor: Same as above, I understand the general proposal, but I'm not 100% clear on how you want this implemented, and not sure how to make it work with our yaml schema.

lib/icp-project/src/lib.rs
rikonor: Should be back now.

lib/icp-canister/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+    pub fn from_bytes<B: AsRef<[u8]>>(bytes: B) -> Result<Self, CanisterManifestError> {
+        let cm: CanisterManifest = serde_yaml::from_slice(bytes.as_ref()).context(ParseSnafu)?;
+
+        Ok(cm)
+    }
+}
+
+#[derive(Debug, Snafu)]
+pub enum CanisterManifestError {
+    #[snafu(display("failed to parse canister manifest: {}", source))]
```
adamspofford: Errors about a file are easier to dissect when they contain the path to the file, especially when we have multiple kinds of yaml file. This probably necessitates the parse function loading the file vs just converting it from string content.

Cargo.toml
```diff
@@ -1,16 +1,19 @@
 [workspace]
 members = [
     "bin/icp-cli",
+    "lib/icp-canister",
```
adamspofford, line 4: Is icp-canister a meaningfully different set of functionality from icp-project? Do you think they need to be separate?

lib/icp-canister/src/lib.rs
ghost: I think if you make the above change, and delete the other `AdapterType`, then adapters will deserialize from the form above.

lib/icp-canister/src/lib.rs
ghost: By adding a .0 member as above, in combination with `#[serde(tag="type")]`
```
pub struct RustAdapter {
  pub package: String
};
```

then the following would deserialize:
```
build:
  adapter:
    type: rust
    package: <package>
```

lib/icp-canister/src/lib.rs
rikonor: Ok, should be good now, I think.

lib/icp-canister/src/lib.rs
rikonor: Ok, should be good now, I think.

bin/icp-cli/src/commands/build.rs
rikonor: Yep, should be fixed now using `transparent`.

Cargo.toml
rikonor: I think so, yes.

lib/icp-canister/src/lib.rs
rikonor: Sounds good - I switched `from_bytes` to `from_file` and give it a path that is then reported on. Hope it looks better now.

bin/icp-cli/src/commands/build.rs
rikonor: TIL. Changed.

bin/icp-cli/src/commands/build.rs
```diff
@@ -0,0 +1,57 @@
...
+
+#[derive(Debug, Snafu)]
+pub enum BuildCommandError {
+    #[snafu(display("no project (icp.yaml) found in current directory or its parents"))]
+    ProjectNotFound,
+
+    #[snafu(display("failed to load project manifest: {path}"))]
+    ProjectLoad {
+        source: ProjectManifestError,
+        path: String,
```
ghost:
```suggestion
        path: PathBuf,
```

rikonor: Ok, updated all `path: String` to `path: PathBuf`.

lib/icp-project/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        pm.canisters = cs;
+
+        Ok(pm)
+    }
+}
+
+#[derive(Debug, Snafu)]
+pub enum ProjectManifestError {
```
ghost:
```suggestion
pub enum LoadProjectManifestError {
```

lib/icp-project/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+    Parse { source: serde_yaml::Error },
+
+    #[snafu(display("invalid UTF-8 in canister path"))]
+    InvalidPathUtf8,
+
+    #[snafu(transparent)]
+    GlobPattern { source: glob::PatternError },
+
+    #[snafu(transparent)]
+    GlobWalk { source: glob::GlobError },
```
adamspofford: These should not be transparent. The glob error does not tell you what file the glob was encountered in, what the tool was trying to do with the glob, what directory it was trying to walk, etc. Transparent is for when we already have a snafu contextful error.

lib/icp-canister/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+        // Parse
+        let cm: CanisterManifest = serde_yaml::from_slice(bytes.as_ref())?;
+
+        Ok(cm)
+    }
+}
+
+#[derive(Debug, Snafu)]
+pub enum CanisterManifestError {
+    #[snafu(display("canister manifest not found: {path:?}"))]
```
ghost:
```suggestion
    #[snafu(display("canister manifest not found: {}", path.display()))]
```

lib/icp-canister/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// Path to a script or executable used to build the canister.
+    pub script: String,
+}
+
+/// Configuration for an asset canister used to serve static content.
+#[derive(Debug, Deserialize)]
+pub struct AssetsAdapter {
+    /// Directory containing the static assets to include in the canister.
+    pub source: String,
+}
```
ghost: Let's model assets in a different way.

There will probably have to be an "adapter" for asset canisters, but that's really a way to know that this canister should use the asset canister wasm.

I think there should be a separate configuration section, maybe `sync` or `assets`. The idea is that any canister, not just those that use the asset canister.wasm, can support the asset upload and http access interfaces.

lib/icp-canister/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+
+#[derive(Debug, Snafu)]
+pub enum CanisterManifestError {
+    #[snafu(display("canister manifest not found: {path:?}"))]
+    NotFound { path: PathBuf },
+
+    #[snafu(transparent)]
+    Parse { source: serde_yaml::Error },
```
ghost: This Parse error isn't going to have `path` in the context (it can't, since it's coming from an operation on a slice of bytes). So at mininum, this Parse error should have a `path` member. That said, it's for this reason that the icp-fs crate has `load_json_file` and `save_json_file`.

I would suggest adding a `load_yaml_file` method in the `icp-fs` crate, and the entire body of this from_file can be `load_yaml_file(path)`. The method is literally gluing together the idea that we're loading a canister manifest, and a canister manifest is a yaml file.
There's enough context for the reader to see that the tool could load or there was a parse error in `canisters/frontend/canister.yaml`. We don't need the error message to also say that it's a "canister manifest."

lib/icp-project/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+#[derive(Debug, Deserialize)]
+pub struct ProjectManifest {
+    /// List of canister manifests belonging to this project.
+    /// Supports glob patterns to specify multiple canister YAML files.
+    #[serde(default)]
+    pub canisters: Vec<PathBuf>,
+
+    /// List of network definition files relevant to the project.
+    /// Supports glob patterns to reference multiple network config files.
+    #[serde(default)]
```
ghost: What would you think about setting this up so the default is `networks/*.yaml` ?

lib/icp-project/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+
+use icp_fs::fs::{ReadFileError, read};
+
+/// Represents the manifest for an ICP project, typically loaded from `icp.yaml`.
+/// A project is a repository or directory grouping related canisters and network definitions.
+#[derive(Debug, Deserialize)]
+pub struct ProjectManifest {
+    /// List of canister manifests belonging to this project.
+    /// Supports glob patterns to specify multiple canister YAML files.
+    #[serde(default)]
```
ghost: What would you think about setting this up so the default is `canisters/*` ?

lib/icp-project/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub networks: Vec<PathBuf>,
+}
+
+impl ProjectManifest {
+    pub fn from_file<P: AsRef<Path>>(path: P) -> Result<Self, ProjectManifestError> {
+        let path = path.as_ref();
+
+        // Check existence
+        ensure!(path.exists(), NotFoundSnafu { path });
```
ghost: Unless we're treating this case differently at a higher level (and maybe not even then), I would let read() handle this.

lib/icp-project/src/lib.rs
rikonor: Changed to `LoadProjectManifestError` and `LoadCanisterManifestError`.

lib/icp-project/src/lib.rs
rikonor: Fixed the errors that should not be transparent under `icp-project` and `icp-canister`. Note that `glob::GlobError` actually does contain a `path` in it, so I've left that one as transparent. Let me know if you still want the wrapping `GlobWalk` to also contain the path, but it seems unneeded?

lib/icp-canister/src/lib.rs
rikonor: Thanks, TIL - fixed throughout.

lib/icp-canister/src/lib.rs
rikonor: This sounds reasonable to me. I'll remove the `Assets` variant for now until we can discuss it a bit more.

lib/icp-project/src/lib.rs
rikonor: Sure, changed.

lib/icp-project/src/lib.rs
rikonor: Sure, changed.

lib/icp-project/src/lib.rs
rikonor: No problem, removed.

lib/icp-canister/src/lib.rs
rikonor: Followed your suggestion and added a `load_yaml_file` function under `icp-fs`. Switched to using it in `icp-canister` and `icp-project`.

lib/icp-project/src/lib.rs
adamspofford: If the glob error tells you the path it failed to glob into then it doesn't necessarily need to be repeated, as long as you can find where the glob is and what the problem with it is. But the error still needs to tell you which file the glob is found in.

bin/icp-cli/src/commands/build.rs
```diff
@@ -0,0 +1,56 @@
...
+
+#[derive(Parser, Debug)]
+pub struct Cmd;
+
+pub async fn dispatch(_cmd: Cmd) -> Result<(), BuildCommandError> {
+    let path = ProjectDirectory::find()
+        .ok_or(BuildCommandError::ProjectNotFound)?
+        .structure()
+        .root()
+        .join("icp.yaml");
```
ghost:
```suggestion
        .project_yaml_path();
```

bin/icp-cli/src/commands/build.rs
```diff
@@ -0,0 +1,56 @@
...
+    ProjectLoad {
+        source: LoadProjectManifestError,
+        path: PathBuf,
+    },
+
+    #[snafu(display("failed to load canister manifest: {}", path.display()))]
+    CanisterLoad {
+        source: LoadCanisterManifestError,
+        path: PathBuf,
+    },
```
ghost: LoadCanisterManifestError should have context, so I'd suggest making this transparent. If you do prefer to leave this as a separate message, it shouldn't have a path, since the source will.

```suggestion
    #[snafu(transparent)]
    CanisterLoad {
        source: LoadCanisterManifestError,
    },
```

bin/icp-cli/src/commands/build.rs
```diff
@@ -0,0 +1,56 @@
...
+#[derive(Debug, Snafu)]
+pub enum BuildCommandError {
+    #[snafu(display("no project (icp.yaml) found in current directory or its parents"))]
+    ProjectNotFound,
+
+    #[snafu(display("failed to load project manifest: {}", path.display()))]
+    ProjectLoad {
+        source: LoadProjectManifestError,
+        path: PathBuf,
+    },
```
ghost:
```suggestion
    #[snafu(transparent)]
    ProjectLoad {
        source: LoadProjectManifestError,
    },
```

lib/icp-canister/src/lib.rs
```diff
@@ -0,0 +1,87 @@
+use std::path::Path;
+
+use serde::Deserialize;
+use snafu::Snafu;
+
+use icp_fs::yaml::{LoadYamlFileError, load_yaml_file};
+
+/// Configuration for a Rust-based canister build adapter.
+#[derive(Debug, Deserialize)]
+pub struct RustAdapter {
```
ghost: Suggestion: put all these in model.rs

bin/icp-cli/src/commands/build.rs
rikonor: Changed

bin/icp-cli/src/commands/build.rs
rikonor: Changed

bin/icp-cli/src/commands/build.rs
rikonor: Changed

lib/icp-canister/src/lib.rs
rikonor: Done

lib/icp-project/src/lib.rs
rikonor: It seems to mention the `path`:

```rs
pub struct GlobError {
    path: PathBuf,
    error: Error,
}
```

adamspofford: It mentions the path that the glob failed to discover. It does not mention the configuration file that the glob came from.

rikonor: I think this is leading to the following code structure:
```rs
fn do_something(path: ...) -> Result<...> {
  my_fn_1(...).context(<include_path>)?;
  ...
  my_fn_n(...).context(<include_path>)?;
}

let _ = do_something(path)?;
```
which seems to require a lot of duplicated effort to mention the input path in each individual error. This seems like it'd be less bothersome:
```rs
fn do_something(path: ...) -> Result<...> {
  my_fn_1(...).context(<no_path_mentioned>)?;
  ...
  my_fn_n(...).context(<no_path_mentioned>)?;
}

let _ = do_something(path).context(<include_path>)?;
```
Either way, I made the change you requested.

bin/icp-cli/src/commands/build.rs
```diff
@@ -0,0 +1,50 @@
...
+    // List canisters in project
+    let mut cs = Vec::new();
+
+    for c in pm.canisters {
+        let path = c.join("canister.yaml");
+
+        let cm = CanisterManifest::from_file(&path)?;
+
+        cs.push(cm);
+    }
```
ghost: The glob patterns need to be evaluated relative to the project root.  Right now if you run the `build` command in a subdirectory, it doesn't find the canister definitions:

```
$ icp build 
[CanisterManifest { name: "backend", build: Build { adapter: Rust(RustAdapter { package: "backend" }) } }, CanisterManifest { name: "frontend", build: Build { adapter: Rust(RustAdapter { package: "frontend" }) } }]
$ (cd canisters ; icp build)
[]
```

I'd suggest something like this, so that both methods have easy access to the project root directory, and also to put knowledge of the filename `canister.yaml` in `ProjectDirectoryStructure`:

```suggestion
    let pd = ProjectDirectory::find()?
        .ok_or(BuildCommandError::ProjectNotFound)?;

    let pm = pd.load_project_manifest()?;

    // List canisters in project
    let mut cs = Vec::new();

    for c in pm.canisters {
        let cm = pd.load_canister_manifest(c)?;

        cs.push(cm);
    }
```

lib/icp-project/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+        for pattern in pm.canisters {
+            let matches = glob::glob(pattern.as_str()).context(GlobPatternSnafu { pattern })?;
+
+            for cpath in matches {
+                let cpath = cpath.context(GlobWalkSnafu { path: mpath })?;
+
+                let path: Utf8PathBuf = cpath.try_into()?;
+
+                // Skip non-canister directories
+                if !path.join("canister.yaml").exists() {
```
ghost: Recommendation: Put the notion of `project root + <canister path from icp.yaml> + "canister.yaml"` into `ProjectDirectoryStructure`

bin/icp-cli/src/commands/build.rs
rikonor: > put knowledge of the filename canister.yaml in ProjectDirectoryStructure

Can you elaborate a bit on `pd.load_project_manifest()?` and `pd.load_canister_manifest(c)?`? You mention doing this on `ProjectDirectoryStructure` but then in your example you invoke methods on `ProjectDirectory`

lib/icp-project/src/lib.rs
rikonor: Would this not mean that `icp-project` needs to import something from `bin/icp-cli`? I would have thought we'd want `bin/icp-cli` to import things from `lib` but not vice-versa?

bin/icp-cli/src/commands/build.rs
rikonor: In the meantime, I fixed it by using the `<project_dir> + <glob_pattern>` instead of just the `<glob_pattern>`. Let me know if that satisfies you in the meantime, otherwise if you can provide some more details about what you were proposing I can try and take a swing at that instead.

lib/icp-project/src/lib.rs
ghost: I also think it's reasonable to consider something like the second form

bin/icp-cli/src/commands/build.rs
ghost: I was proposing something like this. I'm satisfied with what you've got, except that I think `ProjectDirectoryStructure` should be the only place that knows about the constant `"canister.yaml"`:
```
impl ProjectDirectory {
  fn load_project_manifest(&self) -> Result<ProjectManifest, ...> {
      let path = self.structure.project_yaml_path();
       // I would also consider moving/inlining ProjectManifest::from_path here:
      ProjectManifest::from_path(path)
  }
  fn load_canister_manifest(&self, canister_dir: &Path) -> Result<CanisterManifest, ...> {
      let path = self.structure.canister_yaml_path(canister_dir);
      let manifest = load_yaml_file(path)?;
      Ok(manifest)
  }
}

impl ProjectDirectoryStructure {
  fn canister_yaml_path(&self, canister_dir) -> PathBuf {
    self.root.join(canister_dir).join("canister.yaml")
  }
}

File: lib/icp-project/src/lib.rs (hunk 22 above)
Comment (ghost) [reply lib/icp-project/src/lib.rs]: No, `ProjectDirectoryStructure` and `ProjectDirectory` would move into `icp-project`

File: bin/icp-cli/src/commands/build.rs (hunk 21 above)
Comment (rikonor) [reply bin/icp-cli/src/commands/build.rs]: Got it - I added the proposed method to `ProjectDirectoryStructure`.

File: lib/icp-project/src/lib.rs (hunk 22 above)
Comment (rikonor) [reply lib/icp-project/src/lib.rs]: I see. Ok, I brought `ProjectDirectory` and `ProjectDirectoryStructure` into `icp-project` and am using the `canister_yaml_path` instead of hardcoding the `canister.yaml` string as before. (See https://github.com/dfinity/icp-cli/pull/7#discussion_r2121709706).

File: bin/icp-cli/src/commands/network/run.rs [hunk 23]
```diff
@@ -1,12 +1,10 @@
-use crate::project::directory::ProjectDirectory;
-use crate::{
-    commands::network::run::RunNetworkCommandError::ProjectNotFound,
-    project::directory::FindProjectError,
-};
 use clap::Parser;
-use icp_network::{ManagedNetworkModel, RunNetworkError, run_network};
 use snafu::Snafu;
```
ghost: We should have one "use" section per file. That way the formatter will prevent spurious changes like these
```suggestion
```

lib/icp-canister/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+/// Represents the manifest describing a single canister,
+/// including its name and how it should be built.
+#[derive(Debug, Deserialize)]
+pub struct CanisterManifest {
+    /// Name of the canister described by this manifest.
+    pub name: String,
+
+    /// Build configuration for producing the canister's WebAssembly.
+    pub build: Build,
+}
```
ghost: This is all the model too. I'd suggest putting it in model.rs

lib/icp-project/src/directory.rs
```diff
@@ -1,8 +1,10 @@
-use crate::project::structure::ProjectDirectoryStructure;
+use std::io;
+
```
ghost:
```suggestion
```

lib/icp-project/src/directory.rs
```diff
@@ -1,8 +1,10 @@
...
 use camino::Utf8PathBuf;
-use icp_network::NetworkDirectory;
 use snafu::{ResultExt, Snafu};
-use std::io;
+
```
ghost:
```suggestion
```

lib/icp-project/src/lib.rs
```diff
@@ -0,0 +1,101 @@
+use camino::{Utf8Path, Utf8PathBuf};
+use serde::Deserialize;
+use snafu::{OptionExt, ResultExt, Snafu};
+
```
ghost: The problem with splitting up `use` sections like this is that the IDEs don't get the memo, add imports to the section you didn't want, which then requires manual cleanup.
```suggestion
```

bin/icp-cli/src/commands/build.rs
```diff
@@ -0,0 +1,50 @@
...
+        .structure()
+        .project_yaml_path();
+
+    let pm = ProjectManifest::from_file(&path)?;
+
+    // List canisters in project
+    let mut cs = Vec::new();
+
+    for c in pm.canisters {
+        let path = c.join("canister.yaml");
```
ghost: Can you think of a way to make this use ProjectDirectoryStructure::canister_yaml_path() ?

lib/icp-project/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+            for cpath in matches {
+                let cpath = cpath.context(GlobWalkSnafu { path: mpath })?;
+
+                let path: Utf8PathBuf = cpath.try_into()?;
+
+                // Skip non-canister directories
+                if !pds.canister_yaml_path(&path).exists() {
+                    continue;
+                }
```
ghost: Can you add a test for this? It seems like this would quietly skip a canister directly specified in the workspace, if canister.yaml didn't exist in the directory.

Example:
```
$ cat icp.yaml
canisters:
  - canisters/frontend
$ ls canisters 
ls: canisters: No such file or directory
$ icp build
[]
```

bin/icp-cli/src/commands/build.rs
```diff
@@ -0,0 +1,50 @@
...
+
+use icp_canister::{CanisterManifest, LoadCanisterManifestError};
+use icp_project::{LoadProjectManifestError, ProjectManifest};
+
+use icp_project::directory::{FindProjectError, ProjectDirectory};
+
+#[derive(Parser, Debug)]
+pub struct Cmd;
+
+pub async fn dispatch(_cmd: Cmd) -> Result<(), BuildCommandError> {
```
ghost: Suggested convention (currently followed elsewhere in this repo): `dispatch` for commands that "dispatch" to a subcommand, "exec" for leaf commands like this one.
```suggestion
pub async fn exec(_cmd: Cmd) -> Result<(), BuildCommandError> {
```

lib/icp-project/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+    #[serde(default = "default_networks")]
+    pub networks: Vec<Utf8PathBuf>,
+}
+
+impl ProjectManifest {
+    pub fn from_file<P: AsRef<Utf8Path>>(path: P) -> Result<Self, LoadProjectManifestError> {
+        let mpath = path.as_ref();
+
+        // Project
+        let pds = ProjectDirectoryStructure::new(mpath);
```
ghost: Currently in this PR, we have the `icp build` command doing this:

```
    let path = ProjectDirectory::find()?
        .ok_or(BuildCommandError::ProjectNotFound)?
        .structure()
        .project_yaml_path();

    let pm = ProjectManifest::from_file(&path)?;
```

but then here, `ProjectManifest::from_file` turns around and calls `ProjectDirectoryStructure::new(path to icp.yaml)`.

That's incorrect, partly because ProjectDirectoryStructure::new() accepts a path to the directory containing icp.yaml, not a path to icp.yaml. But more than that, this method shouldn't be constructing a ProjectDirectoryStructure at all, because one has already been constructed.

Here is a suggestion for how to break up the concerns:
- ProjectDirectoryStructure provides paths (already exists)
- ProjectDirectory loads/deserializes models from paths (example: `ProjectDirectory::load_project_manifest(&self) ... { Ok(load_yaml_file(self.structure.canister_yaml_path())?) }` and one for loading a canister manifest) (already exists, though not those methods)
- ProjectManifest and CanisterManifest are only the model matching the yaml representation
- Project and Canister hold higher-level logic. Example:
```
impl Project {
  pub fn load(pd: &ProjectDirectory) ... {
    let manifest = pd.load_canister_manifest()?;
    // Canister::load() may later look up adapters, that sort of thing
    let canisters = glob over manifest.canisters: Canister::load(pd, canister_name)

    Ok(Self { manifest, canisters } )
  }
}
```

Alternatively, move this method into ProjectDirectory, and have it return what it does now (ProjectManifest, with all globs expanded)

bin/icp-cli/src/commands/network/run.rs
rikonor: I'd automatically settled on a pattern of:
```rs
<std imports>
<empty line>
<third party imports>
<empty line>
<internal imports>
```
You're saying to just group all imports together? Sure, I can do that.

bin/icp-cli/src/commands/build.rs
rikonor: Done

bin/icp-cli/src/commands/build.rs
rikonor: Done. I recommend documenting this convention.

lib/icp-project/src/lib.rs
rikonor: I had assumed that's the desired behavior, isn't it?
My assumption was that indicating a directory was for a canister required placing a valid `canister.yaml` file in it.
If one doesn't exist, it's akin to the developer indicating that the directory is not a canister.
This allows specifying a glob where not every directory in the result set is a canister.

lib/icp-project/src/lib.rs
rikonor: I wonder if it'd be simpler to just make public constants for `canister.yaml` and `icp.yaml`, and that way we wouldn't have to pass around ProjectDirectory objects? If that is not satisfactory, I'd appreciate if you could provide a bit more information about your intended approach, because I was unable to fully understand it from the above.

lib/icp-project/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+        let mut pm: ProjectManifest = load_yaml_file(mpath)?;
+
+        // Project canisters
+        let mut cs = Vec::new();
+
+        for pattern in pm.canisters {
+            let mdir = mpath
+                .parent()
+                .context(ProjectDirectorySnafu { path: mpath })?;
+
```
ghost: One possible approach:

```suggestion
    pub fn load(pds: &ProjectDirectoryStructure) -> Result<Self, LoadProjectManifestError> {
        let mpath = pds.project_yaml_path();

        // Load
        let mut pm: ProjectManifest = load_yaml_file(mpath)?;

        // Project canisters
        let mut cs = Vec::new();

        for pattern in pm.canisters {
            let mdir = pds.root();

```

lib/icp-project/src/lib.rs
ghost: I think that's fine if there is an actual glob specified (`canisters/*`, for example). However, if a specific directory is listed (`canisters/frontend`, as in the example above), then we shouldn't ignore it.

lib/icp-project/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// List of canister manifests belonging to this project.
+    /// Supports glob patterns to specify multiple canister YAML files.
+    #[serde(default = "default_canisters")]
+    pub canisters: Vec<Utf8PathBuf>,
+
+    /// List of network definition files relevant to the project.
+    /// Supports glob patterns to reference multiple network config files.
+    #[serde(default = "default_networks")]
+    pub networks: Vec<Utf8PathBuf>,
+}
```
ghost: I think ProjectManifest, default_networks, and default_canisters belong in model.rs

lib/icp-project/src/lib.rs
rikonor: I think that's fair. If I'm understanding correctly, this is the behavior you're expecting:
1. If an explicit path to a directory is specified, raise an exception if that directory does not actually contain a canister.yaml (i.e _is not a canister_).
2. If a glob path is provided, ignore directories within it that are not canisters (i.e don't raise an exception for those).

lib/icp-project/src/lib.rs
rikonor: Perhaps you could clarify in the codebase docs or elsewhere where and when you expect this pattern to be followed.

Using file names like `models`, `utils`, `common`, is not necessarily always an appropriate pattern, and can often just lead to a dumping ground of definitions, especially if abused in the early stages of a project.

ghost: Right now I'm trying to gently nudge toward not having a dumping ground, because the dumping ground is currently lib.rs. If I were doing this I would put these in config/model/canister.rs and config/model/project.rs (see similar structure in icp-network crate), but maybe you have a better idea.

rikonor: Done

lib/icp-project/src/lib.rs
rikonor: I've implemented `TryFrom<&ProjectDirectoryStructure> for ProjectManifest` which should achieve what you're asking for - let me know if that works for you.

lib/icp-project/src/model.rs
```diff
@@ -0,0 +1,97 @@
...
+        // Load
+        let mut pm: ProjectManifest = load_yaml_file(mpath)?;
+
+        // Canisters
+        let mut cs = Vec::new();
+
+        for pattern in pm.canisters {
+            let mdir = mpath
+                .parent()
+                .context(ProjectDirectorySnafu { path: mpath })?;
```
ghost:
```suggestion
            let mdir = pds.structure.root();
```

lib/icp-project/src/model.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// List of network definition files relevant to the project.
+    /// Supports glob patterns to reference multiple network config files.
+    #[serde(default = "default_networks")]
+    pub networks: Vec<Utf8PathBuf>,
+}
+
+impl TryFrom<&ProjectDirectoryStructure> for ProjectManifest {
+    type Error = LoadProjectManifestError;
+
+    fn try_from(pds: &ProjectDirectoryStructure) -> Result<Self, Self::Error> {
```
ghost: If I understand TryFrom, it's meant more for things like type conversion, not disk i/o operations. Since we're loading the project manifest from the filesystem, not trying to convert an instance of a ProjectDirectoryStructure to an instance of a ProjectManifest, I think this would be more clear:

```suggestion
impl ProjectManifest {
    fn load(pds: &ProjectDirectoryStructure) -> Result<Self, LoadProjectManifestError> {
```

lib/icp-project/src/model.rs

lib/icp-project/src/model.rs
rikonor: Sure, that's fair. Changed.

lib/icp-project/src/model.rs
```diff
@@ -0,0 +1,91 @@
...
+                .context(GlobPatternSnafu { pattern })?;
+
+            for cpath in matches {
+                let cpath = cpath.context(GlobWalkSnafu { path: mpath })?;
+
+                let path: Utf8PathBuf = cpath.try_into()?;
+
+                // Skip non-canister directories
+                if !pds.canister_yaml_path(&path).exists() {
+                    continue;
```
ghost, line 80:
```suggestion
                    // todo: only skip if this was actually a glob, not a specific named directory
                    continue;
```

lib/icp-project/src/lib.rs
rikonor: Per offline conversation, this will be a follow up PR (tracked via [SDK-2152](https://dfinity.atlassian.net/browse/SDK-2152))

lib/icp-project/src/model.rs
```diff
@@ -0,0 +1,91 @@
...
+
+        Ok(pm)
+    }
+}
+
+#[derive(Debug, Snafu)]
+pub enum LoadProjectManifestError {
+    #[snafu(display("failed to find project directory for project manifest {path}"))]
+    ProjectDirectory { path: Utf8PathBuf },
+
```
ghost: Looks like this variant is no longer used
```suggestion
```

lib/icp-project/src/model.rs
rikonor: I created a [subtask](https://dfinity.atlassian.net/browse/SDK-2152) for this, so there shouldn't be a need to leave a TODO in the code.

lib/icp-project/src/model.rs
rikonor: Removed

lib/icp-project/src/model.rs
```diff
@@ -0,0 +1,88 @@
...
+                .context(GlobPatternSnafu { pattern })?;
+
+            for cpath in matches {
+                let cpath = cpath.context(GlobWalkSnafu { path: mpath })?;
+
+                let path: Utf8PathBuf = cpath.try_into()?;
+
+                // Skip non-canister directories
+                if !pds.canister_yaml_path(&path).exists() {
+                    continue;
```
ghost, line 80: This comment is showing up as resolved. I think it would be good to leave a TODO comment here about the intended behavior that this isn't quite following.

rikonor: Sure, added some better documentation for the function mentioning this aspect of its behavior.

lib/icp-project/src/model.rs
```diff
@@ -0,0 +1,112 @@
...
+/// when the `canisters` field is not explicitly specified in the YAML.
+fn default_canisters() -> Vec<Utf8PathBuf> {
+    ["canisters/*"].iter().map(Utf8PathBuf::from).collect()
+}
+
+/// Provides the default glob pattern for locating network definition files
+/// when the `networks` field is not explicitly specified in the YAML.
+fn default_networks() -> Vec<Utf8PathBuf> {
+    ["networks/*"].iter().map(Utf8PathBuf::from).collect()
+}
```
adamspofford, line 17: I would recommend these be of type Pattern, not Utf8PathBuf

rikonor: Can you provide a bit more information on what you mean? Are you referring to `glob::Pattern`? If I try and do that, I seem to run into quite a lot of errors (e.g Pattern not be deserializable, I think), could you explain why you recommend this change?

adamspofford: Because `networks/*` isn't actually a path to anything, it's a glob. If you read it, it's an error, and Pattern::new doesn't take Path either. 

If Pattern can't be serialized, String is good enough.

rikonor: @adamspofford-dfinity Please do let me know if you have more feedback about this, I will merge for now since it doesn't seem like something that would be a blocker in this case. Thanks for the comments.

PR #9, opened 2025-05-29
feat: Convert to `camino` and `camino-tempfile`

adamspofford-dfinity:
See slack thread for rationale (tldr: the tool is likely to behaviorally require UTF-8 paths no matter what). Tradeoffs: If a foreign library like `notify` hands us a `PathBuf`, we'll have to add a UTF-8 conversion. In exchange we can get rid of all the `to_string_lossy`, `display`, etc. conversions. The clippy lint is just to make sure we don't forget, it can be `#[allow]`ed anywhere it's more convenient to.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,10 @@
 [workspace]
-members = ["bin/icp-cli", "lib/icp-identity", "lib/icp-network", "lib/icp-fs"]
+members = [
+    "bin/icp-cli",
+    "lib/icp-identity",
+    "lib/icp-network",
+    "lib/icp-fs",
+]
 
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -1,14 +1,10 @@
 [workspace]
-members = [
-    "bin/icp-cli",
-    "lib/icp-identity",
-    "lib/icp-network",
-    "lib/icp-fs"
-]
+members = ["bin/icp-cli", "lib/icp-identity", "lib/icp-network", "lib/icp-fs"]
```
ghost: ...

PR #10, opened 2025-06-04
feat: Implement Canister Build Adapter Framework and Script Adapter

rikonor:
**Summary:**

This PR introduces canister build adapters within the `icp-adapter` crate. It defines a common `Adapter` trait and provides an initial, fully functional implementation for a `ScriptAdapter`. This adapter allows canister builds to be performed by executing arbitrary shell commands. Placeholders for `MotokoAdapter` and `RustAdapter` are also included for future development.

**Detailed Changes:**

- **Build Adapter Framework ([`lib/icp-adapter/src/lib.rs`](lib/icp-adapter/src/lib.rs:1)):**

  - Introduced an `async Adapter` trait with a `compile` method, serving as the core abstraction for different build mechanisms.
  - Established modules for `script`, `motoko`, and `rust` adapters.
  - Implemented a unified `AdapterCompileError` enum using `snafu` for consistent error handling across adapters.

- **Script Adapter ([`lib/icp-adapter/src/script.rs`](lib/icp-adapter/src/script.rs:1)):**

  - The `ScriptAdapter` allows defining canister build processes via one or more shell commands.
  - Commands are parsed using `shellwords`, enabling complex commands with quoted arguments.
  - The adapter executes commands within the canister's specified directory.
  - Standard input, output, and error streams are inherited from the executing process, making build logs directly visible.
  - Comprehensive error handling is in place for command parsing, invocation failures, and non-zero exit codes.
  - The `ScriptAdapter` configuration (e.g., the command(s) to run) is designed to be deserializable from a configuration file (like `canister.yaml`). For example:
    ```yaml
    # canister.yaml (example snippet)
    build:
      adapter: script
      command: "npm run build"
    # or for multiple commands:
    # commands:
    #   - "npm install"
    #   - "npm run build"
    ```

- **Placeholder Adapters:**
  - **Motoko Adapter ([`lib/icp-adapter/src/motoko.rs`](lib/icp-adapter/src/motoko.rs:1)):** Includes a basic `MotokoAdapter` struct and a stubbed `compile` method.
  - **Rust Adapter ([`lib/icp-adapter/src/rust.rs`](lib/icp-adapter/src/rust.rs:1)):** Includes a basic `RustAdapter` struct and a stubbed `compile` method.

**Future Work:**

- Implement the compilation logic for `MotokoAdapter`.
- Implement the compilation logic for `RustAdapter`.
- Integrate these adapters into the broader canister build lifecycle.

---

_This PR description was generated by Roo._

### DIFF
--- a/bin/icp-cli/tests/build_tests.rs
+++ b/bin/icp-cli/tests/build_tests.rs
@@ -1,0 +1,50 @@
+use crate::common::TestEnv;
+use icp_fs::fs::{create_dir_all, write};
+use predicates::{ord::eq, str::PredicateStrExt};
+
+mod common;
+
+#[test]
+fn build_adapter_script_simple() {
+    let env = TestEnv::new();
+
+    // Setup project
+    let project_dir = env.create_project_dir("icp");
+
+    // Project manifest
+    let pm = r#"
+    canisters:
+      - my-canister
+    "#;
+
+    write(
+        project_dir.join("icp.yaml"), // path
+        pm,                           // contents
+    )
+    .expect("failed to write project manifest");
+
+    // Canister manifest
+    let cm = r#"
+    name: my-canister
+    build:
+      adapter:
+        type: script
+        command: echo hi
+    "#;
+
+    create_dir_all(project_dir.join("my-canister")).expect("failed to create canister directory");
+
+    write(
+        project_dir.join("my-canister/canister.yaml"), // path
+        cm,                                            // contents
+    )
+    .expect("failed to write project manifest");
+
+    // Invoke build
+    env.icp()
+        .current_dir(project_dir)
+        .args(["build"])
+        .assert()
+        .success()
+        .stdout(eq("hi").trim());
+}


lib/icp-adapter/src/script.rs
```diff
@@ -0,0 +1,97 @@
...
+    InvalidCommand { command: String, reason: String },
+
+    #[snafu(display("failed to execute command {command}: {source}"))]
+    CommandInvoke {
+        command: String,
+        source: std::io::Error,
+    },
+
+    #[snafu(display("command {command} failed with status code {code}"))]
+    CommandStatus { command: String, code: String },
```
adamspofford: `: {source}` isn't needed in errors

adamspofford: Since commands have multiple words, they should be quoted with "" or \`\` in messages

rikonor: Can you explain what you mean? Which error are you referring to, `CommandStatus` (it doesn't mention `source`)? Why isn't `source` needed, will it be printed anyway? Please explain better.

Regardless, should be fixed now.

rikonor: Fixed 👍

adamspofford: Referring to the whole error enum. The reason it isn't needed is because the full series of source errors is written in a list, so adding it to the error message repeats information. The formatter will strip it out, but it will then add a warning that it stopped it out which we want to avoid printing.

rikonor: Got it - I did change that after your initial comment.

lib/icp-adapter/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+use script::ScriptAdapterCompileError;
+use snafu::Snafu;
+
+pub mod motoko;
+pub mod rust;
+pub mod script;
+
+#[async_trait]
+pub trait Adapter {
+    async fn compile(&self, path: Utf8PathBuf) -> Result<(), AdapterCompileError>;
```
ghost: The path being passed is to the canister directory, yeah?
Also as I understand it, it's preferable for parameters to be &Utf8Path rather than Utf8PathBuf 
```suggestion
    async fn compile(&self, canister_path: &Utf8Path) -> Result<(), AdapterCompileError>;
```

lib/icp-adapter/src/script.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// Set of commands used to build a canister
+    Commands(Vec<String>),
+}
+
+/// Configuration for a custom canister build adapter.
+#[derive(Debug, Deserialize)]
+pub struct ScriptAdapter {
+    /// Command used to build a canister
+    #[serde(flatten)]
+    pub command: CommandField,
```
ghost, line 32: See https://github.com/dfinity/sdk/blob/master/src/dfx-core/src/json/structure.rs#L12 for an alternative (note: not suggesting adding a dependency on the crate, rather, copying that type here)
```suggestion
    pub command: SerdeVec,
```

lib/icp-adapter/src/script.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+#[async_trait]
+impl Adapter for ScriptAdapter {
+    async fn compile(&self, path: Utf8PathBuf) -> Result<(), AdapterCompileError> {
+        let cmds = match &self.command {
+            CommandField::Command(cmd) => std::slice::from_ref(cmd),
+            CommandField::Commands(cmds) => cmds,
+        };
```
ghost: If you use SerdeVec, this can be
```suggestion
        let cmds = self.command.into_vec();
```

lib/icp-adapter/src/script.rs
```diff
@@ -0,0 +1,97 @@
...
+                shellwords::split(input_cmd).context(CommandParseSnafu { command: input_cmd })?;
+
+            // Separate command and args
+            let (cmd, args) = input.split_first().context(InvalidCommandSnafu {
+                command: input_cmd,
+                reason: "command must include at least one element".to_string(),
+            })?;
+
+            // Command
+            let mut cmd = Command::new(cmd);
```
ghost, line 60: Suppose the command is "build.sh", and there's also a `build.sh` somewhere on the path. This will call the build.sh that's on the path, rather than the one in the canister directory.

I'd suggest at minimum to try to canonicalize `cmd` in the canister dir; if found, pass the canonicalized path as the command. dfx does this [here](https://github.com/dfinity/sdk/blob/master/src/dfx/src/util/command.rs#L8).

lib/icp-adapter/src/lib.rs
rikonor: Sounds good, changed.

lib/icp-adapter/src/script.rs
rikonor: Can you explain what you mean by `canonicalize` here? I'm confused, because if I have:
```yaml
name: my-canister
build:
  adapter:
    type: script
    command: <executable>
```
but I want to execute the one from the path, wouldn't canonicalizing make the tool try to run `<path_to_canister>/<executable>`?

raymondk: @ericswanson-dfinity would we actually want to pass stuff through `sh -c` ?

ghost: @raymondk yes, I think we should do everything [dfx does here](https://github.com/dfinity/sdk/blob/master/src/dfx/src/util/command.rs#L8) except maybe the `words.len() == 1` part , including going through `sh -c`.

rikonor: @adamspofford-dfinity Didn't you say previously that we should not pass things through a shell? This was when I was passing things through a shell.

ghost: @rikonor 

If `command` references an executable that's not found in the canister dir, then we'd want to look it up on the path.

If `command` references an executable (and scripts are executables) that is found in the canister dir, then we want to run that, even if there is another executable on the path with the same name.

This is where canonicalize comes in. If `canonicalize` returns Ok(path), it means the file was found, and it
>  Returns the canonical, absolute form of a path with all intermediate components normalized and symbolic links resolved.

```
let canister_relative_cmd = project_root.join(canister_dir).join(cmd);
let cmd = if let Ok(cmd_in_canister_dir) = dunce::canonicalize(canister_relative_cmd) {
    cmd_in_canister_dir
} else {
    cmd
};
let mut cmd = Command::new(cmd);
```

The dunce crate is recommended to avoid UNC paths on Windows where possible.

rikonor: Thank you for the explanation. I'll let @adamspofford-dfinity chime in before I make the change.

adamspofford: Agreed on the canonicalization but I don't think we should be going through `sh`. The pipe thing was added in [this commit](https://github.com/dfinity/sdk/pull/3670/files), before that the command was run directly, and it seems mostly used for `cut` tricks in `rustc --version` which (a) cannot be reproduced in a cross-platform way, and (b) we should not design the tool to make necessary.

ghost: I think it's reasonable to say if the user wants sh -c, they can put that in their command

lib/icp-adapter/src/script.rs
rikonor: Could you clarify what the benefit of this change would be?

From my perspective, it seems to serve a similar purpose, but introduces some ambiguity—`SerdeVec` isn’t immediately obvious and would require readers to look it up. In contrast, `CommandField` is defined just above, which helps with readability.

Is there a specific advantage I might be missing?

lib/icp-adapter/src/script.rs
rikonor: This should be implemented now.

lib/icp-adapter/src/script.rs
rikonor: I applied a similar change in https://github.com/dfinity/icp-cli/pull/10/commits/7b0e7705fe3bbf93b8f19d9481a5d9f857cbf7e3

lib/icp-adapter/src/script.rs
rikonor: I applied a similar change in https://github.com/dfinity/icp-cli/pull/10/commits/7b0e7705fe3bbf93b8f19d9481a5d9f857cbf7e3.

bin/icp-cli/tests/adapter_script_tests.rs
```diff
@@ -0,0 +1,116 @@
+use std::io::Read;
+
+use camino_tempfile::NamedUtf8TempFile;
+use icp_adapter::{
+    Adapter, AdapterCompileError,
+    script::{CommandField, ScriptAdapter, ScriptAdapterCompileError},
+};
+
+#[tokio::test]
+async fn single_command() {
```
ghost: Could we have these tests show these configurations in canister.yaml, then call `icp build`, then check that the scripts did what's expected?

This would also serve to show that the configuration looks like what we expect, and that it all works as part of the build command.

The tests in this directory are integration tests meant to test the built binary at a high level. These tests as-written are unit tests, but I think it would be better to rework them as integration tests here calling `icp build`.  The tests in `identity_tests.rs` show a place to start; you'd create an `icp.yaml`  and a `canister.yaml`.

If you choose to leave these tests as-is, please move them into the icp-adapter project as normal unit tests, and update the test.yml github workflow to run unit tests. It currently only runs only integration tests, since at present there are no unit tests.

lib/icp-adapter/src/script.rs
ghost: Two reasons:
1) Overall, it's a reusable piece. There will be other script inputs, and my guess is they won't all be called `command`.
2) The `#[serde(rename_all = "lowercase")]` on CommandField makes me wonder if the intent here is that the field is called either `command` or `commands` depending on whether it's an array or not. I think it should be `command` always.

Also, given that the configuration format is yaml rather than json, I wonder if it even makes sense to have this split at all, since it will be possible to express a multi-line command like this:
```
command: |
   something
   something else
```

The existing code won't work with the above, so we either need to support it or disallow it. If we support it, we will have to decide what that means. 

What do you think?

bin/icp-cli/tests/adapter_script_tests.rs
```diff
@@ -0,0 +1,116 @@
...
+async fn multiple_commands() {
+    // Create temporary file
+    let mut f = NamedUtf8TempFile::new().expect("failed to create temporary file");
+
+    // Define adapter
+    let v = ScriptAdapter {
+        command: CommandField::Commands(vec![
+            format!("sh -c 'echo cmd-1 >> {}'", f.path()),
+            format!("sh -c 'echo cmd-2 >> {}'", f.path()),
+            format!("sh -c 'echo cmd-3 >> {}'", f.path()),
```
ghost: This is great. I think it will be helpful to show what this looks like in a canister yaml. I guess it would be something like this:

```
build:
  adapter:
    kind: script
    command:
      - sh -c 'echo cmd-1 >> {f}'
      - sh -c 'echo cmd-2 >> {f}'
      - sh -c 'echo cmd-2 >> {f}'
 ```

And I'd wonder what would happen if the user instead wrote this:
```
build:
  adapter:
    kind: script
    command: |
      sh -c 'echo cmd-1 >> {f}'
      sh -c 'echo cmd-2 >> {f}'
      sh -c 'echo cmd-2 >> {f}'
```

File: bin/icp-cli/tests/adapter_script_tests.rs (hunk 6 above)
Comment (rikonor) [reply bin/icp-cli/tests/adapter_script_tests.rs]: Hi, @ericswanson-dfinity. I moved the unit tests to be under `icp-adapter`.

File: lib/icp-adapter/src/script.rs (hunk 3 above)
Comment (rikonor) [reply lib/icp-adapter/src/script.rs:32]: Hi @ericswanson-dfinity - I tend to disagree with (2), I think if you are specifying a list of commands, it makes sense for the field to be called `commands` and not `command`. Regarding (1), I don't disagree, but I would hold off on adding this abstraction until it's needed, especially since it does add some mental overhead in being familiar with `SerdeVec` and it's use-case (IMO).

File: bin/icp-cli/tests/adapter_script_tests.rs (hunk 7 above)
Comment (rikonor) [reply bin/icp-cli/tests/adapter_script_tests.rs]: Hi @ericswanson-dfinity - per https://github.com/dfinity/icp-cli/pull/10#discussion_r2145311359, I put the unit tests under `icp-adapter`. Regarding the multi-line command - I agree that it doesn't seem like that would be properly processed at the moment - perhaps for each _command_ provided, we could split it by lines, although I can already see that it would break commands like this:
```sh
multi-line \
  command \
  --with args
```

.github/workflows/test.yml
```diff
@@ -54,6 +54,9 @@ jobs:
           echo "ICPTEST_DFX_PATH=$(dfx cache show)/dfx" >> "$GITHUB_ENV"
           echo "ICP_POCKET_IC_PATH=$(dfx cache show)/pocket-ic" >> "$GITHUB_ENV"
 
+      - name: Run unit tests
+        run: cargo test --workspace --lib --bins
```
ghost: This run unit tests on every matrix run. For example, with the current set, it runs unit tests three times:
- [with network_tests on ubuntu](https://github.com/dfinity/icp-cli/actions/runs/15638211238/job/44058806638?pr=10#step:7:1)
- [with network_tests on macos](https://github.com/dfinity/icp-cli/actions/runs/15638211238/job/44058806650?pr=10#step:7:1)
- [with identity_tests on ubuntu](https://github.com/dfinity/icp-cli/actions/runs/15638211238/job/44058806654?pr=10#step:7:1)

It should instead run as a separate job (running the unit tests on ubuntu and macus) , and then have the `aggregate` job check for it as well.

rikonor: Thanks, updated to be its own job.

lib/icp-adapter/src/script.rs
```diff
@@ -0,0 +1,227 @@
...
+
+#[derive(Debug, Snafu)]
+pub enum ScriptAdapterCompileError {
+    #[snafu(display("failed to parse command '{command}'"))]
+    CommandParse {
+        command: String,
+        source: shellwords::MismatchedQuotes,
+    },
+
+    #[snafu(display("invalid command '{command}'"))]
```
ghost:
```suggestion
    #[snafu(display("invalid command '{command}': {reason}"))]
```

bin/icp-cli/src/commands/build.rs
```diff
@@ -21,12 +24,27 @@ pub async fn exec(_cmd: Cmd) -> Result<(), BuildCommandError> {
...
+        let cm = CanisterManifest::from_file(&mpath)?;
+        cs.push((c, cm));
     }
 
     // Build canisters
-    println!("{cs:?}");
+    for (path, c) in cs {
+        match c.build.adapter {
+            Adapter::Script(adapter) => {
+                adapter.compile(&path).await?;
```
ghost, line 36: I think we're at the point where the `icp build` command should have its first integration test.

Right now, this PR adds to this end-to-end behavior of parsing icp.yaml, reading canister manifests, executing script-based build adapters, but it only has unit coverage for one internal part.

Could you add an integration test under bin/icp-cli/tests/build_tests.rs? Something that sets up a minimal project (an icp.yaml and one or more canister.yaml files with a script adapter), runs icp build, and verifies that the expected script(s) were executed?

This would validate the command from the user’s perspective and ensure we’re wiring everything together correctly.

Let me know if you'd like help sketching out what the test might look like or using the TestEnv fixture.

lib/icp-adapter/src/script.rs
rikonor: Done

bin/icp-cli/src/commands/build.rs
rikonor: @ericswanson-dfinity I attempted to add an integration test (https://github.com/dfinity/icp-cli/pull/10/commits/4b2e48352edb61a4a45517571bddc5b19b3c745b) that would setup a simple no-op canister manifest that just does `echo hi`, but I'm having trouble collecting its output. I was wondering if you might have an idea - this is the test output:
```sh
Test environment home directory: /var/folders/cv/clkmf0n148ncfc3jh_9cwyk00000gn/T/.tmpgpw8Gv

thread 'build_adapter_script_simple' panicked at /Users/orricon/.rustup/toolchains/1.86.0-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5:
Unexpected stdout, failed diff original var
├── original: hi
├── diff: 
│   ---         orig
│   +++         var
│   @@ -1 +0,0 @@
│   -hi
└── var as str: 

command=`cd "/var/folders/cv/clkmf0n148ncfc3jh_9cwyk00000gn/T/.tmpgpw8Gv/icp" && HOME="/var/folders/cv/clkmf0n148ncfc3jh_9cwyk00000gn/T/.tmpgpw8Gv" "/Users/orricon/workspace/chirbun/icp-cli/target/debug/icp" "build"`
code=0
stdout=""
stderr=""

stack backtrace:
   0: rust_begin_unwind
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panicking.rs:695:5
...
```

rikonor: Should be fixed now. I tried using a single-canister project manifest, which isn't implemented yet.

PR #11, opened 2025-06-06
feat: Support single-canister project definitions

rikonor:
**Summary:**

This PR enhances the project manifest (`icp.yaml`) to support a simplified configuration for single-canister projects. Previously, only a `canisters` list was supported. Now, a project with a single canister can be defined directly using a top-level `canister` key, improving developer experience for simpler projects. The existing `canisters` key for multi-canister projects remains fully supported.

**Detailed Changes:**

- **Flexible Project Manifest (`lib/icp-project/src/model.rs`):**

  - The `RawProjectManifest` struct was modified to use `#[serde(flatten)]` on an `Option<RawCanistersField>`. This allows `serde` to deserialize from either a `canister` key (for a single canister) or a `canisters` key (for multiple canisters) at the top level of `icp.yaml`.
  - The `ProjectManifest::load` function was updated to handle this `Option`. If neither `canister` nor `canisters` is present in the manifest, it now correctly falls back to the default behavior of searching for canisters with the `canisters/*` glob pattern.

- **Unified Build Logic (`bin/icp-cli/src/commands/build.rs`):**
  - The `build` command's execution logic was updated to normalize the canister configuration. It now uses a `match` statement to handle both the single-canister (`CanistersField::Canister`) and multi-canister (`CanistersField::Canisters`) variants, creating a unified list of canisters to be built.
  - This ensures that the downstream build process can iterate over a consistent data structure, regardless of how the canisters are defined in the manifest.

**Configuration Examples:**

This change makes both of the following `icp.yaml` formats valid:

**Single-Canister Project (New):**

```yaml
canister:
  name: my-canister
  build:
    adapter:
      type: script
      command: echo "building my canister"
```

**Multi-Canister Project (Existing):**

```yaml
canisters:
  - canisters/*
```

---

_This PR description was generated by Roo._

### DIFF
--- a/bin/icp-cli/tests/build_tests.rs
+++ b/bin/icp-cli/tests/build_tests.rs
@@ -1,0 +1,50 @@
+use crate::common::TestEnv;
+use icp_fs::fs::{create_dir_all, write};
+use predicates::{ord::eq, str::PredicateStrExt};
+
+mod common;
+
+#[test]
+fn build_adapter_script_simple() {
+    let env = TestEnv::new();
+
+    // Setup project
+    let project_dir = env.create_project_dir("icp");
+
+    // Project manifest
+    let pm = r#"
+    canisters:
+      - my-canister
+    "#;
+
+    write(
+        project_dir.join("icp.yaml"), // path
+        pm,                           // contents
+    )
+    .expect("failed to write project manifest");
+
+    // Canister manifest
+    let cm = r#"
+    name: my-canister
+    build:
+      adapter:
+        type: script
+        command: echo hi
+    "#;
+
+    create_dir_all(project_dir.join("my-canister")).expect("failed to create canister directory");
+
+    write(
+        project_dir.join("my-canister/canister.yaml"), // path
+        cm,                                            // contents
+    )
+    .expect("failed to write project manifest");
+
+    // Invoke build
+    env.icp()
+        .current_dir(project_dir)
+        .args(["build"])
+        .assert()
+        .success()
+        .stdout(eq("hi").trim());
+}
--- a/bin/icp-cli/tests/network_tests.rs
+++ b/bin/icp-cli/tests/network_tests.rs
@@ -39,6 +39,8 @@ fn hello() {
         .arg("motoko")
         .arg("--frontend")
         .arg("simple-assets")
+        .arg("--agent-version") // don't contact npm to look up the agent-js version
+        .arg("99.99")
         .assert()
         .success();
 
--- a/lib/icp-adapter/src/script.rs
+++ b/lib/icp-adapter/src/script.rs
@@ -97,7 +97,7 @@ pub enum ScriptAdapterCompileError {
         source: shellwords::MismatchedQuotes,
     },
 
-    #[snafu(display("invalid command '{command}'"))]
+    #[snafu(display("invalid command '{command}': {reason}"))]
     InvalidCommand { command: String, reason: String },
 
     #[snafu(display("failed to execute command '{command}'"))]


bin/icp-cli/src/commands/build.rs
```diff
@@ -1,45 +1,65 @@
...
+/// 3. Normalizes the canister definitions into a unified list.
+/// 4. Iterates through each defined canister and invokes its respective build adapter
+///    (Rust, Motoko, or custom script) to compile it into WebAssembly.
+///
+/// # Arguments
+/// * `_` - An unused `Cmd` struct, typically for command-line argument parsing.
+///
+/// # Returns
+/// A `Result` indicating success (`Ok(())`) or an error (`Err(BuildCommandError)`)
+/// if the project cannot be found, the manifest fails to load, or a canister build fails.
```
ghost: These boilerplate comments don't add useful information. To the extent that they describe the kinds of errors that the method may return, they are mostly an opportunity to get out of sync.
```suggestion
```

lib/icp-canister/src/model.rs
```diff
@@ -34,22 +37,37 @@ pub struct Build {
...
+    /// file into a `CanisterManifest` struct, providing the necessary
+    /// configuration for a single canister.
+    ///
+    /// # Arguments
+    /// * `path` - A reference to the path of the `canister.yaml` file.
+    ///
+    /// # Returns
+    /// A `Result` which is:
+    /// - `Ok(Self)` if the manifest is successfully loaded and parsed.
+    /// - `Err(LoadCanisterManifestError)` if an error occurs during file loading or parsing.
```
ghost: This comment also seems like unnecessary boilerplate. I'm not sure the comments above add useful information either. This three-line function loads a yaml file from the path specified. I don't think it needs a multi-line comment, but if it needs anything, the top line alone should suffice.

```suggestion
    /// Loads a `CanisterManifest` from the specified YAML file path.
```

lib/icp-project/src/model.rs
```diff
@@ -16,34 +17,65 @@ fn default_networks() -> Vec<Utf8PathBuf> {
     ["networks/*"].iter().map(Utf8PathBuf::from).collect()
 }
 
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum RawCanistersField {
+    Canister(CanisterManifest),
+    Canisters(Vec<Utf8PathBuf>),
```
ghost, line 24: As mentioned by the reviewer for the previous PR, the use of `Utf8PathBuf` here isn't appropriate, because a value like `canisters/*` is a glob pattern, not a path.

Since `glob::Pattern` isn't deserializable, even `String` would be better at this level:
```suggestion
    Canisters(Vec<String>),
```

lib/icp-project/src/model.rs
```diff
@@ -16,34 +17,65 @@ fn default_networks() -> Vec<Utf8PathBuf> {
...
+    /// (for multiple canisters, supporting glob patterns).
+    /// If neither key is present, it defaults to `None`, which is then handled
+    /// by the `ProjectManifest::load` function to apply a default glob pattern.
+    #[serde(flatten)]
+    pub canisters: Option<RawCanistersField>,
+
+    /// List of network definition files relevant to the project.
+    /// Supports glob patterns to reference multiple network config files.
+    #[serde(default = "default_networks")]
+    pub networks: Vec<Utf8PathBuf>,
```
ghost, line 43: This is another place where the the value in the file may be a glob pattern, not a path. Therefore `Utf8PathBuf` isn't appropriate since it may not be holding a path, and even `String` would be a better representation for both possibilities.

```suggestion
    pub networks: Vec<String>,
```

lib/icp-project/src/model.rs
```diff
@@ -16,34 +17,65 @@ fn default_networks() -> Vec<Utf8PathBuf> {
...
+
 /// Represents the manifest for an ICP project, typically loaded from `icp.yaml`.
 /// A project is a repository or directory grouping related canisters and network definitions.
 #[derive(Debug, Deserialize)]
 pub struct ProjectManifest {
     /// List of canister manifests belonging to this project.
-    /// Supports glob patterns to specify multiple canister YAML files.
-    #[serde(default = "default_canisters")]
-    pub canisters: Vec<Utf8PathBuf>,
+    pub canisters: CanistersField,
```
ghost: What would you think about unifying this at this level to always be a vector of canisters?

```suggestion
    pub canisters: Vec<ProjectCanister>,
```

Then elsewhere:
```
struct ProjectCanister {
    root: Utf8PathBuf,
    manifest: CanisterManifest
}
```

This will avoid places that access these (like the build command currently, but there will be more) having to each go through this single-or-multiple logic.

lib/icp-project/src/model.rs
```diff
@@ -16,34 +17,65 @@ fn default_networks() -> Vec<Utf8PathBuf> {
...
+
+    /// List of network definition files relevant to the project.
+    /// Supports glob patterns to reference multiple network config files.
+    #[serde(default = "default_networks")]
+    pub networks: Vec<Utf8PathBuf>,
+}
+
 /// Represents the manifest for an ICP project, typically loaded from `icp.yaml`.
 /// A project is a repository or directory grouping related canisters and network definitions.
 #[derive(Debug, Deserialize)]
```
ghost: This is no longer the deserialized type.  Suggestion: group the deserialize types (which match the .yaml format) together in their own module, and put types that aren't deserialized (that don't necessarily match the yaml format) in a separate module.
```suggestion
#[derive(Debug)]
```

lib/icp-project/src/model.rs
rikonor: I created a small [PR](https://github.com/dfinity/icp-cli/pull/13) with this isolated change. I wanted to give it a try with Pattern to see if it'd work. Let me know if you think it's overly complex and would rather it just be `String` throughout.

lib/icp-project/src/model.rs
rikonor: See https://github.com/dfinity/icp-cli/pull/11#discussion_r2136524863

lib/icp-canister/src/model.rs
rikonor: Done

bin/icp-cli/src/commands/build.rs
rikonor: Done

lib/icp-project/src/model.rs
rikonor: The reason I did it like that, meaning distinguish between single-canister and multi-canister projects, is because we discussed it potentially having useful uses (like for metrics collection, etc).

https://dfinity.slack.com/archives/C08TTAZGYJK/p1749250492979019

ghost: If it turns out to be important to be able to distinguish them, we could do that in another way, like a flag on the ProjectCanister struct. Regardless, if there's something that is going to unify these two possibilities into a single vec that a command should work with, it isn't the build command.

rikonor: Should be good now.

PR #12, opened 2025-06-09
fix(icp-project): Improve canister path resolution and error handling

rikonor:
This PR enhances the project manifest loading logic by introducing strict validation for explicit canister paths defined in `icp.yaml`.

#### Key Change

Previously, if a non-glob path in the `canisters` list pointed to a directory that was not a valid canister (e.g., it was missing a `canister.yaml` file), it was silently ignored. This could lead to confusion and hard-to-debug configuration issues.

With this change, the loading process now performs the following checks for each explicit canister path:
1.  The path must exist and be a directory.
2.  The directory must contain a `canister.yaml` manifest file.

If any of these conditions are not met, the process will fail with a descriptive error, providing immediate and clear feedback on the misconfiguration.

#### Other Improvements

-   **Fix Explicit Path Resolution**: Ensured that explicit canister paths are always resolved relative to the project root for correctness, regardless of where the CLI is run.
-   **Improve Error Messages**: Error messages now report the original relative path from the manifest, making it easier for users to locate the source of the problem.
-   **Update Documentation**: The doc comment for `ProjectManifest::load` has been updated to reflect the new validation behavior and remove obsolete notes.

### DIFF
--- a/bin/icp-cli/tests/common/test_env.rs
+++ b/bin/icp-cli/tests/common/test_env.rs
@@ -1,3 +1,4 @@
+use crate::common::guard::ChildGuard;
 use crate::common::os::PATH_SEPARATOR;
 use assert_cmd::Command;
 use camino::{Utf8Path, Utf8PathBuf};
@@ -122,6 +123,20 @@ impl TestEnv {
         std::fs::create_dir_all(&project_dir).expect("Failed to create icp project directory");
         std::fs::write(project_dir.join("icp.yaml"), "").expect("Failed to write project file");
         project_dir
+    }
+
+    pub fn start_network_in(&self, project_dir: &Utf8Path) -> ChildGuard {
+        let icp_path = env!("CARGO_BIN_EXE_icp");
+        let mut cmd = std::process::Command::new(icp_path);
+        cmd.current_dir(project_dir)
+            .env("HOME", self.home_path())
+            .env_remove("ICP_HOME")
+            .arg("network")
+            .arg("run");
+
+        eprintln!("Running network in {}", project_dir);
+
+        ChildGuard::spawn(&mut cmd).expect("failed to spawn icp network ")
     }
 
     // wait up to 30 seconds for descriptor path to contain valid json
--- a/bin/icp-cli/tests/network_tests.rs
+++ b/bin/icp-cli/tests/network_tests.rs
@@ -1,10 +1,8 @@
 mod common;
 
 use crate::common::TestEnv;
-use crate::common::guard::ChildGuard;
 use predicates::str::contains;
 use serial_test::file_serial;
-use std::process::Command;
 
 #[test]
 #[file_serial]
@@ -13,14 +11,7 @@ fn hello() {
 
     let icp_project_dir = testenv.create_project_dir("icp");
 
-    let icp_path = env!("CARGO_BIN_EXE_icp");
-    let mut cmd = Command::new(icp_path);
-    cmd.env("HOME", testenv.home_path())
-        .current_dir(icp_project_dir)
-        .arg("network")
-        .arg("run");
-
-    let _child_guard = ChildGuard::spawn(&mut cmd).expect("failed to spawn icp network ");
+    let _child_guard = testenv.start_network_in(&icp_project_dir);
 
     testenv.configure_dfx_local_network();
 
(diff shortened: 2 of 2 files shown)


lib/icp-project/src/model.rs
```diff
@@ -112,25 +118,64 @@ impl ProjectManifest {
...
+                                .collect()
+                        }
+
+                        // Explicit path
+                        false => {
+                            // Resolve the explicit path against the project root.
+                            let canister_path = pds.root().join(&pattern);
+
+                            // Check if path exists and that it's a directory.
+                            if !canister_path.exists() || !canister_path.is_dir() {
```
ghost: is_dir() already checks that the path exists
```suggestion
                            if !canister_path.is_dir() {
```

lib/icp-project/src/model.rs
```diff
@@ -112,25 +118,64 @@ impl ProjectManifest {
...
-                    //   If it's not, we should raise an error when the specified path
-                    //   does not represent a canister directory (e.g doesnt contain a canister.yaml file)
-                    let matches = glob::glob(pds.root().join(&pattern).as_str())
-                        .context(GlobPatternSnafu { pattern })?;
-
-                    for cpath in matches {
-                        // Directory path of the found canister.
-                        let cpath = cpath.context(GlobWalkSnafu { path: mpath })?;
-                        let path: Utf8PathBuf = cpath.try_into()?;
+                    let dirs = match is_glob(&pattern) {
```
ghost, line 113: Maybe this is an intentional style choice here, but `let dirs = match (bool expression) { true => {...} false => { ... } };` is another way of writing:
```
let dirs = if is_glob(&pattern) {
    // Glob
} else {
    // Explicit path
};
```

Suggestion: extract the two arms of this match to methods, and then see if you still like it as a `match`.

lib/icp-project/src/model.rs
```diff
@@ -112,25 +118,64 @@ impl ProjectManifest {
...
+                            if !canister_path.exists() || !canister_path.is_dir() {
+                                return Err(LoadProjectManifestError::CanisterPath {
+                                    path: pattern,
+                                });
+                            }
+
+                            // Check for a canister manifest.
+                            if !pds.canister_yaml_path(&canister_path).exists() {
+                                return Err(LoadProjectManifestError::NoManifest { path: pattern });
+                            }
```
ghost: Suggestion: put the path to the expected canister manifest (path to return value of `canister_yaml_path()` in the error.

lib/icp-project/src/model.rs
```diff
@@ -157,10 +202,16 @@ pub enum LoadProjectManifestError {
     #[snafu(transparent)]
     InvalidPathUtf8 { source: camino::FromPathBufError },
 
+    #[snafu(display("canister path must exist and be a directory '{path}'"))]
+    CanisterPath { path: Utf8PathBuf },
+
+    #[snafu(display("canister path does not contain a manifest '{path}'"))]
+    NoManifest { path: Utf8PathBuf },
```
ghost:
```suggestion
    #[snafu(display("no canister manifest found at '{path}'"))]
    NoManifest { path: Utf8PathBuf },
```

This is the difference between these two error messages:
```
canister path does not contain a manifest 'src/backend'
no canister manifest found at 'src/backend/canister.yaml'
```

lib/icp-project/src/model.rs
rikonor: Yes it’s a matter of taste, in other languages you can’t assign the result of an if/else so I prefer to assign the result of a match.

lib/icp-project/src/model.rs

lib/icp-project/src/model.rs

lib/icp-project/src/model.rs

lib/icp-project/src/model.rs
rikonor: Resolving, but feel free to re-open if you want to discuss.

bin/icp-cli/tests/project_tests.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    // Load Project
+    let pds = ProjectDirectoryStructure::new(project_dir);
+    let pm = ProjectManifest::load(&pds);
+
+    // Assert failure
+    assert!(matches!(
+        pm,
+        Err(LoadProjectManifestError::NoManifest { .. })
+    ));
```
ghost: These are unit tests, but the `icp-cli/tests/` directory is made to hold integration tests.  I think most, if not all, of these new tests would be great as integration tests. 

```suggestion
    env.icp()))
        .arg("build")
        .assert()
        .failure()
        .stderr(contains(format!("no canister manifest found at {}",
            project_dir.join("canister-1").join("canister.yaml")));

```

rikonor: Sorry, moved to be under the `icp-project` crate. Will add integration tests as well.

ghost: I think it would be fine to have a `project_tests.rs` for project-related integration tests, and since we don't have a `project` command, it could use other commands (like `build`) to exercise the code. What do you think?

PR #13, opened 2025-06-09
feat: Introduce explicit glob pattern handling for project manifest

rikonor:
This change makes it so that we don't treat potential glob patterns in canister paths as always explicit paths. Instead we simply treat them as strings and make a determination later if they're glob patterns or explicit paths.


lib/icp-project/src/model.rs
```diff
@@ -112,25 +130,70 @@ impl ProjectManifest {
...
-                    for cpath in matches {
-                        // Directory path of the found canister.
-                        let cpath = cpath.context(GlobWalkSnafu { path: mpath })?;
-                        let path: Utf8PathBuf = cpath.try_into()?;
+                    let dirs = match is_glob(&pattern) {
+                        // Glob
+                        true => {
+                            // Pattern
+                            let pattern = pattern.value();
+                            let pattern = pattern.as_str();
```
ghost: The `Pattern` struct introduces an interesting possibility here. Suppose Pattern had a method that returned a variant depending on whether the contained string was a glob pattern or a specific directory:
```
pun enum GlobOrPath {
    Glob(String),
    Path(Utf8PathBuf)
}
impl Pattern {
    pub fn globOrPath(&self) -> GlobOrPath {
        if self.is_glob() {
            GlobOrPath::Glob(self.value.as_str())
        } else {
            GlobOrPath::Path(Utf8PathBuf::from_str(pattern.value().as_str()).unwrap())
        }
    }
    fn is_glob(&self) -> bool {
        let pattern = self.value();
        let s = pattern.as_str();
        s.contains('*') || s.contains('?') || s.contains('[') || s.contains('{')
    }
}
```

Then this match could unpack these here:

```
                   let dirs = pattern.globOrPath() {
                        Glob(pattern) => {
                             .. get matches ..
                        }
                        Path(path) => {
                             .. look for specific canister.yaml
                        }
                    }
```

lib/icp-project/src/model.rs
```diff
@@ -112,25 +130,70 @@ impl ProjectManifest {
...
+                            let canister_path = pds.root().join(&path);
+
+                            // Check if path exists and that it's a directory.
+                            if !canister_path.exists() || !canister_path.is_dir() {
+                                return Err(LoadProjectManifestError::CanisterPath { path });
+                            }
+
+                            // Check for a canister manifest.
+                            if !pds.canister_yaml_path(&canister_path).exists() {
+                                return Err(LoadProjectManifestError::NoManifest { path });
```
ghost: Issue carried over from other PR: it would be better for this error message to contain the path to canister.yaml

rikonor: Seems to have slipped between the cracks. Should be fixed now.

lib/icp-project/src/model.rs
rikonor: Doesn't this already do the same thing?
```rs
let dirs = match is_glob(&pattern) {
    // Glob
    true => { ... },
    
    // Explicit path
    false => { ... },
}
```

PR #24, opened 2025-06-18
feat: Implement `canister create` and `canister install` commands

rikonor:
This PR introduces two new subcommands to the `icp-cli` tool: `canister create` and `canister install`. These commands provide essential functionality for managing canister lifecycles on the Internet Computer.

### `canister create`

The `canister create` command allows users to create new canisters on the IC. Key features include:

- **Project-Based Canister Creation**: The command operates within the context of an `icp.yaml` project file, creating canisters defined in the manifest.
- **Flexible Canister Selection**: Users can create all canisters defined in the project or specify a single canister by name.
- **Resource Configuration**: The command provides options to configure compute allocation, memory allocation, freezing threshold, and other canister settings.
- **Controller Management**: Users can specify one or more controllers for the new canister.
- **ID Management**: The command supports specifying an effective canister ID for the creation call and can assign a specific ID to the new canister.
- **Quiet Mode**: A quiet mode is available to suppress human-readable output and print only the canister ID, which is useful for scripting.

### `canister install`

The `canister install` command enables the installation of Wasm modules to canisters. Its main features are:

- **Project-Aware Installation**: Like the create command, it uses the `icp.yaml` file to identify which canisters to install.
- **Targeted Installs**: Users can install code to all canisters in the project or to a specific canister by name.
- **Artifact Lookup**: The command automatically looks up the build artifact (Wasm module) and the canister ID from the local project environment.
- **Clear Feedback**: The command provides clear output to the user, confirming which canisters have had code installed.

### Implementation Details

- The new commands are implemented in the `bin/icp-cli/src/commands/canister/` directory, with `create.rs` and `install.rs` containing the logic for each subcommand.
- The `canister.rs` file has been updated to dispatch to the new subcommands.
- The `store_id.rs` and `store_artifact.rs` files were added to handle the local storage and lookup of canister IDs and build artifacts.
- The changes are integrated with the existing `Env` and project structures, ensuring a consistent and predictable user experience.

### DIFF
--- a/bin/icp-cli/src/main.rs
+++ b/bin/icp-cli/src/main.rs
@@ -17,10 +17,10 @@ struct Cli {
     #[arg(long, global = true)]
     identity: Option<String>,
 
-    #[arg(long, default_value = "./icp/ids.json")]
+    #[arg(long, default_value = ".icp/ids.json")]
     id_store: Utf8PathBuf,
 
-    #[arg(long, default_value = "./icp/artifacts")]
+    #[arg(long, default_value = ".icp/artifacts")]
     artifact_store: Utf8PathBuf,
 
     #[command(flatten)]


bin/icp-cli/src/commands/canister/create.rs
```diff
@@ -1,16 +1,226 @@
...
+    model::{LoadProjectManifestError, ProjectManifest},
+};
 use snafu::Snafu;
 
 #[derive(Debug, Parser)]
-pub struct CanisterCreateCmd {}
+struct CanisterIDs {
+    /// The effective canister ID to use when calling the management canister.
+    #[clap(long)]
+    effective_id: Principal,
```
adamspofford, line 19: This should be guessed rather than required.

bin/icp-cli/src/commands/canister/create.rs
```diff
@@ -1,16 +1,226 @@
...
+
+    /// Canister ID configuration, including the effective and optionally specific ID.
+    #[clap(flatten)]
+    ids: CanisterIDs,
+
+    /// One or more controllers for the canister. Repeat `--controller` to specify multiple.
+    #[clap(long)]
+    controller: Vec<Principal>,
+
+    /// Resource-related options and thresholds for the new canister.
```
adamspofford: Flattened sections don't need a doc comment.

bin/icp-cli/src/commands/canister/create.rs
```diff
@@ -1,16 +1,226 @@
...
+
+    // Create an agent pointing to the desired network endpoint
+    let agent = Agent::builder()
+        .with_url(&cmd.network_url)
+        .with_arc_identity(identity)
+        .build()?;
+
+    if cmd.network_url.contains("127.0.0.1") || cmd.network_url.contains("localhost") {
+        agent.fetch_root_key().await?;
+    }
```
adamspofford, line 101: Agent creation and the associated root key check should be factored into a function, it will be used in many places.

bin/icp-cli/src/commands/canister/create.rs
```diff
@@ -1,16 +1,226 @@
...
+}
+
+#[derive(Debug, Parser)]
+pub struct CanisterCreateCmd {
+    /// The name of the canister within the current project
+    name: Option<String>,
+
+    /// The URL of the IC network endpoint
+    #[clap(long, default_value = "http://localhost:8000")]
+    network_url: String,
```
adamspofford, line 60: This should probably be extracted to the icp-canister command, if not the root icp command, with global = true; it applies to many things, similar to --identity.

bin/icp-cli/src/store_id.rs
```diff
@@ -0,0 +1,100 @@
...
+    LookupLoadStore {
+        source: lockedjson::LoadJsonWithLockError,
+    },
+
+    #[snafu(display("could not find ID for canister '{name}'"))]
+    IdNotFound { name: String },
+}
+
+pub trait Lookup {
+    fn lookup(&self, name: &str) -> Result<Principal, LookupError>;
```
adamspofford: Why define these traits? I don't think they represent something that any function would be generic over.

bin/icp-cli/tests/build_tests.rs
```diff
@@ -11,15 +11,21 @@ fn build_adapter_script() {
     // Setup project
     let project_dir = env.create_project_dir("icp");
 
+    // Create temporary file
+    let f = NamedUtf8TempFile::new().expect("failed to create temporary file");
```
adamspofford, line 15: What's this about?

bin/icp-cli/src/commands/canister/create.rs
rikonor: Yes, but it doesn't apply to _everything_, hence I'd argue it should still be explicitly defined by commands that need it, or if we want it to be a standard utility, we can describe it as a struct somewhere and then commands that need it pull it in with `flatten`.

bin/icp-cli/src/commands/canister/create.rs
rikonor: Yes, agreed, but, per discussion with Eric I did not do it so that he could have a say about it when he's back (See [here](https://dfinity.atlassian.net/browse/SDK-2173?focusedCommentId=190800))

bin/icp-cli/src/commands/canister/create.rs
rikonor: Why not?

bin/icp-cli/src/commands/canister/create.rs
rikonor: Could you please explain? I'm not sure what "guessing" means in this context.

bin/icp-cli/src/store_id.rs
rikonor: The goal was to define behavior for "looking up an ID" and for "registering an ID with a canister's name". Then, if a piece of code requires that functionality, it defines it as a dependency on the behavior, rather than a concrete struct.

bin/icp-cli/tests/build_tests.rs
rikonor: While you were gone, and at Raymond's request to expedite canister installation, I modified the build adapter so that it must return a blob for the wasm it built. To do that, the script adapter expects the final command it runs to print the path to the wasm blob it built to stdout.

bin/icp-cli/src/commands/canister/create.rs
adamspofford: Because it doesn't do anything. If there's something unclear about what it does, you can leave it as a regular comment, but a doc comment implies that it'll show up in the command help, which it won't.

bin/icp-cli/src/commands/canister/create.rs
adamspofford: Please explicitly mark it todo then, so we don't lose it.

bin/icp-cli/src/commands/canister/create.rs
adamspofford: As in, use the default effective canister ID, or (non-MS1 behavior) if the user specifies a subnet, use the first canister ID in the subnet's range.

bin/icp-cli/src/store_id.rs
adamspofford: The interface accessed from Env is a concrete object representing access to the store. If we decide that there should be two backends, that can be handled *within* the object, possibly via trait object and possibly via enum, but there likely won't. Adding traits 'just in case' in Rust is rarely a good idea.

bin/icp-cli/tests/canister_tests.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    // TODO(or.ricon): Query canister
+    // env.dfx()
+    //     .current_dir(&project_dir)
+    //     .args([
+    //         "canister",
+    //         "call",
+    //         "--network",
+    //         "http://localhost:8000",
+    //         &cid,
```
adamspofford: This is just missing a `trim()`, then it works.

bin/icp-cli/src/commands/canister/create.rs
rikonor: I did not understand that you were differentiating between a regular comment and a doc comment. I'll make them normal comments.

bin/icp-cli/src/commands/canister/create.rs
rikonor: No problem, done.

bin/icp-cli/tests/canister_tests.rs
rikonor: Thanks for the tip - fixed.

bin/icp-cli/src/store_id.rs
rikonor: It wasn't _'just in case'_, but due to the way Env is currently structured, there was no need for trait objects, hence the concrete type is used instead. I've removed the traits, as requested.

bin/icp-cli/src/commands/canister/create.rs
rikonor: Should be set as a default now per offline conversation. I'll create tasks for a more elaborate `effective ID` determination approach for the local/remote cases.

rikonor: Created https://dfinity.atlassian.net/browse/SDK-2187.

PR #25, opened 2025-06-19
feat: Introduce `deploy` command

rikonor:
This PR introduces a new `deploy` command to the `icp-cli` tool. This command simplifies the deployment process by orchestrating the `build`, `canister create`, and `canister install` commands into a single, streamlined workflow.

### `deploy`

The `deploy` command automates the process of deploying canisters to the Internet Computer. Its key features include:

- **Simplified Workflow**: The `deploy` command combines the `build`, `canister create`, and `canister install` steps into a single command, reducing the complexity of deploying canisters.
- **Project-Based Deployment**: The command operates within the context of an `icp.yaml` project file, deploying canisters defined in the manifest.
- **Flexible Canister Selection**: Users can deploy all canisters defined in the project or specify a single canister by name.
- **Network Configuration**: The command allows users to specify the IC network endpoint to deploy to.
- **Controller Management**: Users can specify one or more controllers for the canisters being deployed.

### Implementation Details

- The new command is implemented in the `bin/icp-cli/src/commands/deploy.rs` file.
- The `commands.rs` file has been updated to include the `deploy` command and dispatch to its implementation.
- The `deploy` command reuses the existing `build`, `canister create`, and `canister install` commands to perform the deployment steps.

### DIFF
--- a/bin/icp-cli/tests/deploy_tests.rs
+++ b/bin/icp-cli/tests/deploy_tests.rs
@@ -117,20 +117,13 @@ fn deploy() {
         .assert()
         .success();
 
-    // TODO(or.ricon): Query canister
-    // env.dfx()
-    //     .current_dir(&project_dir)
-    //     .args([
-    //         "canister",
-    //         "call",
-    //         "--network",
-    //         "http://localhost:8000",
-    //         &cid,
-    //         "greet",
-    //         "(\"test\")",
-    //     ])
-    //     .assert()
-    //     .success();
+    // Query canister
+    env.icp()
+        .current_dir(&project_dir)
+        .args(["canister", "call", "my-canister", "greet", "(\"test\")"])
+        .assert()
+        .success()
+        .stdout(eq("(\"Hello, test!\")").trim());
 }
 
 #[test]
@@ -190,18 +183,11 @@ fn deploy_twice_should_fail() {
         .failure()
         .stderr(contains("cannot be installed because the canister is not empty").trim());
 
-    // TODO(or.ricon): Query canister
-    // env.dfx()
-    //     .current_dir(&project_dir)
-    //     .args([
-    //         "canister",
-    //         "call",
-    //         "--network",
-    //         "http://localhost:8000",
-    //         &cid,
-    //         "greet",
-    //         "(\"test\")",
-    //     ])
-    //     .assert()
-    //     .success();
+    // Query canister
+    env.icp()
+        .current_dir(&project_dir)
+        .args(["canister", "call", "my-canister", "greet", "(\"test\")"])
+        .assert()
+        .success()
+        .stdout(eq("(\"Hello, test!\")").trim());
 }


bin/icp-cli/tests/deploy_tests.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    // Deploy project (second time)
+    env.icp()
+        .current_dir(&project_dir)
+        .args(["deploy", "--effective-id", "ghsi2-tqaaa-aaaan-aaaca-cai"])
+        .assert()
+        .failure()
+        .stderr(contains("cannot be installed because the canister is not empty").trim());
+
+    // TODO(or.ricon): Query canister
```
rikonor: @adamspofford-dfinity I could figure out how to extract the CID from the output of the `deploy` command, but it would probably be even better to use your new `canister call` contribution instead and provide the canister name.

bin/icp-cli/src/commands/deploy.rs
```diff
@@ -0,0 +1,181 @@
...
+    let pd = ProjectDirectory::find()?.ok_or(CommandError::ProjectNotFound)?;
+
+    // Get the project directory structure for path resolution.
+    let pds = pd.structure();
+
+    // Load the project manifest, which defines the canisters to be built.
+    let pm = ProjectManifest::load(pds)?;
+
+    // Choose canisters to create
+    let cs = pm
```
ghost: What does "cs" mean?

bin/icp-cli/src/commands/deploy.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    // Create the selected canisters
+    eprintln!("\nCreating canisters:");
+    for (_, c) in &cs {
+        eprintln!("- {}", c.name);
+
+        // TODO(or.ricon): Temporary approach that can be revisited.
+        //                 Currently we simply invoke the adjacent `canister::create` command.
+        //                 We should consider refactoring `canister::create` to use library code instead.
```
ghost, line 92: I'd suggest the time is now to use library code rather than calling other commands' `exec` methods.

bin/icp-cli/src/commands/deploy.rs
rikonor: It's supposed to be short for `canisters`, but I renamed all the occurrences of it to `canisters` just now for clarity.

bin/icp-cli/src/commands/deploy.rs
rikonor: This approach was taken due to the upcoming demo, which required this functionality. Since that has passed, I'm happy to discuss how to refine this if the preference is to do it as part of this PR.

rikonor: Per [offline conversation](https://dfinity.slack.com/archives/C08TTAZGYJK/p1750710495103809?thread_ts=1750697742.428009&cid=C08TTAZGYJK), this will be addressed in a future change.

bin/icp-cli/tests/deploy_tests.rs
rikonor: Updated the tests to use @adamspofford-dfinity `icp canister call` implementation.

PR #30, opened 2025-06-24
feat: support taking canister deployment options from canister manifest

rikonor:


lib/icp-canister/src/model.rs
```diff
@@ -48,6 +76,10 @@ pub struct CanisterManifest {
     /// The build configuration specifying how to compile the canister's source
     /// code into a WebAssembly module, including the adapter to use.
     pub build: Build,
+
+    /// The deploy configuration specifying the various options to deploy
+    /// the canister with.
+    pub deploy: Deploy,
```
ghost: This happens during create, and only sometimes during deploy (which doesn't create canisters that already exist). Putting it under a `deploy` key could imply that these options are updated during upgrade or even during deploy when there is no other change.

```suggestion
    pub create: Create,
```

rikonor: Ah, good point, changed 👍

PR #31, opened 2025-06-24
refactor: memoize project in Env 

ghost:
This PR simplifies commands' loading of the project manifest and consolidates the error message for a missing project.

It makes Env memoize the ProjectManifest. Upcoming work will build on this to look up the network configuration (which needs to load the project) and ultimately build an Agent.


bin/icp-cli/src/commands/build.rs
```diff
@@ -51,16 +42,16 @@ pub async fn exec(env: &Env, cmd: Cmd) -> Result<(), CommandError> {
     }
 
     // Iterate through each resolved canister and trigger its build process.
-    for (path, c) in pm.canisters {
-        let wasm = match c.build.adapter {
+    for (path, c) in &pm.canisters {
```
rikonor, line 45: I previously introduced a bug here. Note that we create `canisters` above out of `pm.canisters`, but then here `pm.canisters` again. It should be fixed [here](https://github.com/dfinity/icp-cli/pull/35/files#diff-adb925452501c4dbb5e4d63147e400bfe08f1ce7633b899bfca289c9cd05ef12R58).

lib/icp-canister/src/model.rs
```diff
@@ -82,21 +79,3 @@ pub struct CanisterManifest {
     #[serde(default)]
     pub create: Create,
 }
-
-impl CanisterManifest {
```
rikonor, line 86: Would it make sense to keep the loading logic for a canister manifest with `icp-canister`, and then use that in `icp-project`? This way `icp-canister` retains the responsibility of how to load a canister manifest, and also retains the ability to be used in isolation from `icp-project`.

ghost: The way I was looking at is that CanisterManifest is responsible for implementing Deserialize, while ProjectDirectory ties the pathname and format together to load the structure, by constructing the yaml path and then loading the structure as yaml from it. 

The method below (that this PR deletes), in contrast, as a wrapper around load_yaml_file, and doesn't say anything about where the path is. I don't think it will be necessary to load a canister manifest outside of the location in the project where it's found, though if that turns out to be needed, say in tests, the test could achieve that by calling load_yaml_file directly.

PR #34, opened 2025-06-27
feat: builders should reference an environment variable for their output

rikonor:
This updates build adapters to use an environment variable for their wasm output path, instead of returning it directly as a bytes blob.


bin/icp-cli/tests/build_tests.rs
```diff
@@ -22,7 +22,7 @@ fn build_adapter_script() {
           build:
             adapter:
               type: script
-              command: echo {}
+              command: sh -c 'cp {} $ICP_WASM_OUTPUT_PATH'
```
ghost: This without quotes will/should work since this PR uses a temp directory, generally found under /tmp and without spaces. However, tests also often serve as examples to follow, and I can imagine the possibility that in the future we might use paths under `<project dir>/.icp` for output paths. In that case, there might be spaces in the directory name (typically a parent), so surrounding the output path in quotes would be good practice.
```suggestion
              command: sh -c 'cp {} "$ICP_WASM_OUTPUT_PATH"'
```

rikonor: Good point, updated in the various uses of it.

PR #35, opened 2025-06-27
feat: Support multiple build adapters

rikonor:
This update enables the definition of multiple build adapters for a single canister. As long as at least one of the build adapters successfully builds and configures the output WASM file, the overall build will be considered successful. This change enables the build step to assume additional responsibilities beyond simply generating the WASM file.

### DIFF
--- a/bin/icp-cli/tests/build_tests.rs
+++ b/bin/icp-cli/tests/build_tests.rs
@@ -22,7 +22,7 @@ fn build_adapter_script() {
           build:
             adapter:
               type: script
-              command: sh -c 'cp {} $ICP_WASM_OUTPUT_PATH'
+              command: sh -c 'cp {} "$ICP_WASM_OUTPUT_PATH"'
         "#,
         f.path()
     );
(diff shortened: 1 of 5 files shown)


bin/icp-cli/tests/build_tests.rs
```diff
@@ -40,3 +40,46 @@ fn build_adapter_script() {
...
+
+    // Create temporary file
+    let f = NamedUtf8TempFile::new().expect("failed to create temporary file");
+
+    // Project manifest
+    let pm = format!(
+        r#"
+        canister:
+          name: my-canister
+          build:
```
ghost: This doesn't seem consistent with the latest from the design doc:
```
name: my-canister-that-needs-wasm-building
# This is the build step
# At the end of it, you have:
# - a wasm
# - possibly some "other things"
build:
  # This is how I get my wasm
  wasm:
    type: script
    # The final output is a wasm
    # The toolchain or string of commands is responsible for
    # not building unnecessarily (keeping of track of files changed that will require
    # that you rebuild)
    # and any optimizations, adding metadata bla bla bla that you might want
    # icp-cli doesn't care what you have done to the wasm
    commands:
      - cargo build ....
      - wasm optimpize
      - avengers assemble
      - sprinkle magic dust
      - echo WASM_PATH

  # Some steps that need to run before I actually do anything
  # with the canister so that I can fail early
  other-things:
```

PR #39, opened 2025-07-01
feat: configurable networks

ghost:
This PR makes icp-cli load the network configurations listed in the `networks` key of icp.yaml

- removes the `--network-url <url>` parameter
- adds `--network <name>` parameter which can also be specified with the `ICP_NETWORK` environment variable
- adds `--ic` as shorthand for `--network ic`
- consolidates agent creation logic
- removes all root key fetching.
  - For the local network, it now reads the root key from the network descriptor (which it needs anyway in order to determine the gateway port)
  - For other networks, gets the root key from the network configuration file.
- tests can run the network on a random port, in which case they don't need `file_serial`.

### DIFF
--- a/examples/icp-rust/networks/staging.yaml
+++ b/examples/icp-rust/networks/staging.yaml
@@ -1,0 +1,5 @@
+# This "staging" network is an alias for mainnet
+# "icp deploy --network staging" would deploy a second set of canisters to mainnet
+
+mode: connected
+url: https://icp0.io


bin/icp-cli/src/commands/canister/delete.rs
```diff
@@ -1,21 +1,22 @@
...
 
-    /// The URL of the IC network endpoint
-    #[clap(long, default_value = "http://127.0.0.1:8000")]
-    pub network_url: String,
+    #[clap(flatten)]
+    network: NetworkOpt,
 }
 
 pub async fn exec(env: &Env, cmd: CanisterDeleteCmd) -> Result<(), CanisterDeleteError> {
+    env.set_network_opt(cmd.network);
```
rikonor: Do you think it would make sense to make the environment unaware of `clap`? Meaning that instead of saying something like `env.set_network_opt` you could say `env.set_network` and provide an argument that is defined by or near `Env`? That way `Env` doesn't have to assume knowledge of `clap` or cli options.

rikonor: I assume this can't be done ahead of time (before we dispatch to a given command) because some commands might not require a network argument, but it would have been nice if `exec` could have just immediately referred to the network from the env. Oh well.

bin/icp-cli/src/commands/network/run.rs
```diff
@@ -11,7 +11,16 @@ pub async fn exec(env: &Env, _cmd: Cmd) -> Result<(), RunNetworkCommandError> {
     let project = env.project()?;
     let pd = &project.directory;
     let network_name = "local";
-    let config = ManagedNetworkModel::default();
+    let config = project.find_network_config(network_name).ok_or_else(|| {
```
rikonor: Nitpick, but would there be a more appropriate name than `find_network_config`? To me, `find` implies you'd get back some location, not the actual configuration.

bin/icp-cli/src/commands/network/run.rs
```diff
@@ -11,7 +11,16 @@ pub async fn exec(env: &Env, _cmd: Cmd) -> Result<(), RunNetworkCommandError> {
...
+        RunNetworkCommandError::NetworkConfigNotFound {
+            network_name: network_name.to_string(),
+        }
+    })?;
+    let NetworkConfig::Managed(config) = config else {
```
rikonor, line 16: I've been meaning to bring this up - what exactly is a `managed` network? It's not entirely clear from the word `managed`, is it essentially a network that's defined via a local configuration file? Could that be either a local or remote network?

bin/icp-cli/src/env.rs
```diff
@@ -1,39 +1,109 @@
...
 use std::sync::{Arc, OnceLock};
 
 pub struct Env {
     dirs: IcpCliDirs,
-    identity: Option<String>,
+    identity_name: Option<String>,
     pub id_store: IdStore,
     pub artifact_store: ArtifactStore,
     project: TryOnceLock<Project>,
+    identity: TryOnceLock<Arc<dyn Identity>>,
```
rikonor: Maybe consolidate `identity_name` and `identity`? Or at least place them near each other. Also, it'd be pretty helpful to have some doc comments on these.
```
identity: (Option<String>, TryOnceLock<Arc<dyn Identity>>)
```

bin/icp-cli/src/commands/canister/delete.rs
ghost: Yeah, that's the reason. I didn't want to add the network opts to the top-level command parser because not every command has them.

bin/icp-cli/tests/common/network.rs
```diff
@@ -0,0 +1,9 @@
+pub struct TestNetwork {
+    pub gateway_port: u16,
+    pub root_key: String,
+}
+
+pub struct TestNetworkForDfx {
```
rikonor, line 9: Although I'm sure it'd be clear from the rest of the code, it'd be helpful to have some comments explaining what these two structs are.

bin/icp-cli/src/commands/canister/delete.rs
ghost: NetworkOpt is just a struct, and env.get_network_opt(opts) uses the struct, not the clap machinery derived on it. So I wouldn't say Env is aware of clap.

That said, we coul turn this around, something like
```
cmd.network.apply(env);
```

Then NetworkOpt::apply() would call env.set_network(network_name). What do you think?

bin/icp-cli/src/commands/network/run.rs
ghost: A `managed` network is a network that this project manages (you run `icp network run/start` here). A `connected` network is a network that this project connects to (something else manages it). Naming is hard. Happy to bikeshed this in some medium better than CR comments.

bin/icp-cli/src/commands/canister/delete.rs
rikonor: > NetworkOpt is just a struct, and env.get_network_opt(opts) uses the struct

This is true, but I imagine NetworkOpt came into existence so that clap arguments could be loaded. Ideally, it'd have one purpose and not also act as an argument type for `Env`. My preference would be for a small struct (probably identical, or at least for now) that's defined adjacent to `Env`, so that each of the structs has a specific use. That said, it's very much my personal preference, so feel free to disregard.

I think `cmd.network.apply` would then imply that `NetworkOpt` now needs to be aware of `Env`, so it'd be the same issue, but in the other direction.

bin/icp-cli/src/commands/network/run.rs
rikonor: Got it, thanks for explaining.

bin/icp-cli/src/commands/network/run.rs
ghost: Thanks! Changed to get_network_config, returns a Result rather than an Option.

bin/icp-cli/tests/common/network.rs
ghost: Done

bin/icp-cli/src/env.rs
ghost: Moved them next to each other and added doc comments.

bin/icp-cli/src/env.rs
```diff
@@ -61,7 +61,7 @@ impl Env {
             .get_or_try_init(|| self.find_and_load_project())
     }
 
-    pub fn set_network_opt(&self, opt: NetworkOpt) {
+    pub fn require_network(&self, opt: NetworkOpt) {
```
rikonor, line 77: Do we need `opt` if you're immediately converting it to a `name`? Maybe it'd be better to accept the name directly?

PR #42, opened 2025-07-02
feat: add pre-built build adapter and tests for it

rikonor:
This change introduces a pre-built build adapter that allows specifying a local path to a WASM file or a remote URL for it. For the URL case, it also supports providing an optional expected sha256 checksum of the downloaded WASM file.

Additionally, it adds the _beginning_ of a minimal example of a static assets canister.

### DIFF
--- a/lib/icp-adapter/src/pre_built.rs
+++ b/lib/icp-adapter/src/pre_built.rs
@@ -13,18 +13,12 @@ use snafu::Snafu;
 pub struct LocalSource {
     /// Local path on-disk to read a WASM file from
     path: Utf8PathBuf,
-
-    /// Optional sha256 checksum of the local WASM file
-    sha256: Option<String>,
 }
 
 #[derive(Clone, Debug, Deserialize, PartialEq)]
 pub struct RemoteSource {
     /// Url to fetch the remote WASM file from
     url: String,
-
-    /// Optional sha256 checksum of the remote WASM file
-    sha256: Option<String>,
 }
 
 #[derive(Clone, Debug, Deserialize, PartialEq)]
@@ -42,6 +36,9 @@ pub enum SourceField {
 pub struct PrebuiltAdapter {
     #[serde(flatten)]
     source: SourceField,
+
+    /// Optional sha256 checksum of the WASM
+    sha256: Option<String>,
 }
 
 #[async_trait]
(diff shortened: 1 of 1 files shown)


lib/icp-adapter/src/pre_built.rs
```diff
@@ -0,0 +1,153 @@
...
+                let u = Url::from_str(&s.url)
+                    .map_err(|err| PrebuiltAdapterCompileError::Url { source: err })?;
+
+                // Construct request
+                let req = Request::new(
+                    Method::GET,  // method
+                    u.to_owned(), // url
+                );
+
+                // Execute request
```
ghost, line 71: Suggest using the https://crates.io/crates/backoff crate for retries

lib/icp-adapter/src/pre_built.rs
```diff
@@ -0,0 +1,153 @@
...
+                // Read response body
+                let bs = resp
+                    .bytes()
+                    .await
+                    .map_err(|err| PrebuiltAdapterCompileError::Request { source: err })?
+                    .to_vec();
+
+                if let Some(sha256) = &s.sha256 {
+                    // Calculate checksum
+                    let cksm = hex::encode({
```
ghost: a `sha256` is a checksum and `cksum` is also a checksum
```suggestion
                if let Some(expected) = &s.sha256 {
                    // Calculate checksum
                    let actual = hex::encode({
```

lib/icp-adapter/src/pre_built.rs
```diff
@@ -0,0 +1,153 @@
...
+use sha2::{Digest, Sha256};
+use snafu::Snafu;
+
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+pub struct RemoteSource {
+    /// Url to fetch the remote WASM file from
+    url: String,
+
+    /// Optional sha256 checksum of the remote WASM file
+    sha256: Option<String>,
```
raymondk: Do we need a hash for a local file as well?

lib/icp-adapter/src/pre_built.rs
raymondk: Maybe we can add a todo for this?
Also do we need a todo to handle  large files or are we expecting in practice that they're small enough that having it all in memory is not an issue?

rikonor: I second this being a follow-up.

lib/icp-adapter/src/pre_built.rs
rikonor: I wouldn't have thought so, but it seems like an easy thing to add - want me to?

lib/icp-adapter/src/pre_built.rs
rikonor: Changed

lib/icp-adapter/src/pre_built.rs
raymondk: it think it makes sense to have it - it can flow through the same code.

rikonor: Ok, I added it as well.

lib/icp-adapter/src/pre_built.rs
rikonor: To add to this, I suggest having the HTTP client be passed to the adapter or a new adapter handler, so that a client isn't defined within the adapter itself. I did not make this change in this case because introducing dependency injection to the codebase seems like a significantly different change.

PR #45, opened 2025-07-06
feat: json output for identity list

raymondk:


bin/icp-cli/src/commands/identity/list.rs
```diff
@@ -1,23 +1,48 @@
...
+        Format::Json => handle_json_output(list, defaults),
+        Format::Human => handle_human_output(list, defaults),
+    }
+
+    Ok(())
+}
+
+// Prints out the JSON output
+// TODO: This needs to flag the default identity
+fn handle_json_output(list: IdentityList, _defaults: IdentityDefaults) {
```
ghost, line 31: I think we should define specific types for commands' json output, rather than using whatever the internal types happen to be. This will both make it obvious when a change would touch some command's json output format and more clearly define what a command's output is.

For inspiration: https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#list-users

raymondk: ... I don't understand at what level Eric meant the types should be.

rikonor: I think he meant that instead of encoding `icp_identity:: IdentityList`, you should define a new type (probably inside of `handle_json_output` and set its fields according to `list`, and then encode that.

PR #46, opened 2025-07-07
feat: Add an assets canister sync adapter

rikonor:


bin/icp-cli/src/commands/sync.rs
```diff
@@ -8,12 +12,18 @@ use snafu::Snafu;
...
+    pub network: NetworkOpt,
 }
 
 pub async fn exec(env: &Env, cmd: Cmd) -> Result<(), CommandError> {
     // Load the project manifest, which defines the canisters to be synced.
     let pm = env.project()?;
 
+    // Initialize network handler
+    env.require_network(cmd.network.name());
+
```
ghost: This doesn't actually initialize a network handler. It saves the network name parameter for when it's needed later. Every other command handler follows a convention of calling this at the top of the method, before calling any of the fallible methods on Env (like project() or agent()).

```suggestion
pub async fn exec(env: &Env, cmd: Cmd) -> Result<(), CommandError> {
    env.require_network(cmd.network.name());

    // Load the project manifest, which defines the canisters to be synced.
    let pm = env.project()?;

```

rikonor: Ok

bin/icp-cli/tests/sync_tests.rs
```diff
@@ -83,6 +96,19 @@ fn sync_adapter_script_multiple() {
...
+    // Wait for network
+    env.ping_until_healthy(&project_dir);
+
+    // Deploy project
+    env.icp()
+        .current_dir(&project_dir)
+        .args(["deploy", "--effective-id", "ghsi2-tqaaa-aaaan-aaaca-cai"])
+        .assert()
+        .success();
+
```
ghost, line 116: Could we add a curl call after this to show that the asset was deployed? (look for "hello there")

Example: https://github.com/dfinity/icp-cli/blob/main/bin/icp-cli/tests/network_tests.rs#L72

bin/icp-cli/tests/sync_tests.rs
```diff
@@ -38,6 +38,19 @@ fn sync_adapter_script_single() {
...
+    // Wait for network
+    env.ping_until_healthy(&project_dir);
+
+    // Deploy project
+    env.icp()
+        .current_dir(&project_dir)
+        .args(["deploy", "--effective-id", "ghsi2-tqaaa-aaaan-aaaca-cai"])
+        .assert()
+        .success();
```
ghost: Since `deploy` will also sync, it might be good to show that here:
```suggestion
        .success()
        .stdout(eq("syncing").trim());	        .stdout(eq("syncing").trim());
```

bin/icp-cli/tests/sync_tests.rs
rikonor: Yep, done 👍 Thanks for the reference.

bin/icp-cli/tests/sync_tests.rs
rikonor: Sounds good - updated.

PR #48, opened 2025-07-08
refactor: move identity option to commands that need it

ghost:
Makes it so only the commands that use an identity have the `--identity` parameter.

This will avoid the command-line help showing options that aren't actually used.
It will also simplify the means by which the upcoming `network ping` command uses the `anonymous` identity.

### DIFF
--- a/bin/icp-cli/src/commands/canister/create.rs
+++ b/bin/icp-cli/src/commands/canister/create.rs
@@ -1,5 +1,5 @@
 use crate::env::{Env, EnvGetAgentError, GetProjectError};
-use crate::options::NetworkOpt;
+use crate::options::{IdentityOpt, NetworkOpt};
 use crate::store_id::{LookupError, RegisterError};
 use clap::Parser;
 use ic_agent::{AgentError, export::Principal};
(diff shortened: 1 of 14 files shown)


bin/icp-cli/src/commands/canister/create.rs
```diff
@@ -72,6 +75,7 @@ pub struct CanisterCreateCmd {
 }
 
 pub async fn exec(env: &Env, cmd: CanisterCreateCmd) -> Result<(), CanisterCreateError> {
+    env.require_identity(cmd.identity.name());
     env.require_network(cmd.network.name());
```
rikonor, line 79: I wonder if this pattern could be simplified so that the calling function takes care of calling this on behalf of the command? Although, that might then be a little confusing because you have a command flag that you don't actually use it in the command handler itself.

ghost: One way to do that would be to use a builder for Env, with a separate pass drilling into the command and calling builder methods with the parameters from the command in use. I think we might explore that at some point.

PR #49, opened 2025-07-08
feat: network ping command

ghost:
Adds the `icp network ping` command, which accepts the following forms:

```
# ping local/project network
icp network ping
icp network ping local
icp network ping --network local

# ping mainnet
icp network ping ic
icp network ping --ic
icp network ping --network ic

# ping some other network
icp network ping staging
icp network ping --network staging
```

Fixes https://dfinity.atlassian.net/browse/SDK-2173


bin/icp-cli/tests/network_ping_tests.rs
```diff
@@ -0,0 +1,323 @@
...
+        .args(["network", "ping", "--network", "project-b"])
+        .assert()
+        .success();
+    testenv
+        .icp()
+        .current_dir(&no_network_project_dir)
+        .env("ICP_NETWORK", "project-b")
+        .args(["network", "ping"])
+        .assert()
+        .success();
```
rikonor: Nit, but would it make sense to try to make some helper function (or even a closure within the test body) to reduce repetitiveness? E.g:
```rs
enum Name {
  Environment(String),
  Flag(String),
  Positional(String),
}

let call_ping = |name: Name| match network {
  ...
};

call_ping(Name::Positional("project-b")).assert().success();
```

ghost: Thank you for this suggestion. I'll try it and see how it turns out.

PR #50, opened 2025-07-08
test: json field predicate

ghost:
This PR adds a `JsonFieldPredicate` the intended purpose of which is to make it easier to verify the output of commands that output a json document. 

It might be easier to start the review at [one of the tests](https://github.com/dfinity/icp-cli/pull/50/files#diff-2c12f5ffcc8afcd46da7a842f8773bf881865308de5996216994f7922fb5fcaeR34).

### DIFF
--- a/bin/icp-cli/tests/network_ping_tests.rs
+++ b/bin/icp-cli/tests/network_ping_tests.rs
@@ -1,6 +1,7 @@
 mod common;
 
 use crate::common::TestEnv;
+use crate::common::predicates::json::field::json_field;
 use assert_cmd::assert::Assert;
 use camino::Utf8Path;
 use predicates::str::contains;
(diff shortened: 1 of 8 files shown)


bin/icp-cli/tests/network_ping_tests.rs
```diff
@@ -198,36 +181,22 @@ fn attempt_ping_other_project() {
...
-        .expect("missing 'replica_health_status' field")
-        .as_str()
-        .expect("'replica_health_status' was not a string");
-
-    assert_eq!(status, "healthy", "unexpected replica_health_status");
+        .stdout(
+            json_field("root_key")
+                .array()
+                .equals(expected_root_key.clone()),
+        )
```
rikonor, line 194: This seems a bit overkill when we could parse the stdout into a struct and verify that way, no?

PR #51, opened 2025-07-09
chore: add more examples and document their purpose and usage

rikonor:


examples/icp-sync/icp.yaml
```diff
@@ -0,0 +1,15 @@
+canister:
+  name: my-canister
+
+  build:
+    - adapter:
+        type: script
+        commands:
+          - sh -c 'command -v moc >/dev/null 2>&1 || { echo >&2 "moc not found. To install moc, see https://internetcomputer.org/docs/current/developer-docs/setup/install/.\n"; exit 1; }'
```
raymondk: The new urls don't have `/current` in them.
```suggestion
          - sh -c 'command -v moc >/dev/null 2>&1 || { echo >&2 "moc not found. To install moc, see https://internetcomputer.org/docs/building-apps/getting-started/install \n"; exit 1; }'
```

examples/icp-motoko/icp.yaml
```diff
@@ -1,10 +1,9 @@
 canister:
-  name: icp-template-motoko
+  name: my-canister
   build:
     adapter:
       type: script
       commands:
        - sh -c 'command -v moc >/dev/null 2>&1 || { echo >&2 "moc not found. To install moc, see https://internetcomputer.org/docs/current/developer-docs/setup/install/.\n"; exit 1; }'
```
raymondk:
```suggestion
       - sh -c 'command -v moc >/dev/null 2>&1 || { echo >&2 "moc not found. To install moc, see https://internetcomputer.org/docs/building-apps/getting-started/install \n"; exit 1; }'
```

examples/icp-static-react-site/icp.yaml
```diff
@@ -0,0 +1,22 @@
+canister:
+  name: my-canister
+
+  build:
+    - adapter:
```
raymondk, line 5: Do we add something in here that checks that npm is installed and in the path? Do you need to run `npm install` first?

examples/icp-sync/icp.yaml
rikonor: Ah, thanks - fixed.

examples/icp-motoko/icp.yaml
rikonor: Ah, thanks - fixed.

examples/icp-static-react-site/icp.yaml
rikonor: Good idea - I added this:
```yaml
- adapter:
    type: script
    commands:
      - sh -c 'command -v npm >/dev/null 2>&1 || { echo >&2 "npm not found. To install npm, see https://docs.npmjs.com/downloading-and-installing-node-js-and-npm \n"; exit 1; }'
      - npm install
      - npm run build
```

PR #54, opened 2025-07-16
chore: refactor canister manifest from build.[-adapter] build.steps

rikonor:


lib/icp-project/src/project.rs
```diff
@@ -522,13 +527,13 @@ mod tests {
             project_dir.path().join("canister-1"),
             CanisterManifest {
                 name: "canister-1".into(),
-                build: BuildSteps::Single(Build {
-                    adapter: AdapterBuild::Script(ScriptAdapter {
+                build: BuildSteps {
+                    steps: vec![AdapterBuild::Script(ScriptAdapter {
```
raymondk: Should we rename AdapterBuild and AdapterSync to be `BuildStep` and `SyncStep` also - to keep the code consistent with the names we're using in the file?

rikonor: Seems reasonable, changed.

PR #57, opened 2025-07-23
feat: Support for deployment environments

rikonor:


lib/icp-project/src/project.rs
```diff
@@ -196,6 +199,58 @@ impl Project {
...
+        // Default environments network
+        let environments = environments
+            .into_iter()
+            .map(|mut v| match v.network {
+                // Explicitly-specified network
+                Some(_) => v,
+
+                // Default network for an environment is `local`
+                None => {
+                    v.network = Some("local".into());
```
raymondk: Shouldn't "local" be a constant since it's used in other places and has a  special meaning?

lib/icp-project/src/project.rs
```diff
@@ -196,6 +199,58 @@ impl Project {
...
+                    v
+                }
+            })
+            .collect::<Vec<EnvironmentManifest>>();
+
+        // Complain about environments that point to non-existent networks
+        for e in &environments {
+            if let Some(network) = &e.network {
```
raymondk, line 333: Some(network_name) or Some(nname)?

lib/icp-project/src/project.rs
```diff
@@ -180,7 +183,7 @@ impl Project {
                 // Duplicate
                 Entry::Occupied(e) => {
                     return Err(LoadProjectManifestError::DuplicateNetwork {
-                        name: e.key().to_owned(),
+                        network: e.key().to_owned(),
```
raymondk, line 273: Am I mistaken or do we not check for duplicate canister names?
Maybe we can add a TODO to make the canisters a hashmap and make sure the names are unique?

lib/icp-project/src/project.rs
```diff
@@ -358,8 +416,19 @@ pub enum LoadProjectManifestError {
         source: LoadNetworkConfigurationsError,
     },
 
-    #[snafu(display("project defines two similarly named networks: '{name}'"))]
-    DuplicateNetwork { name: String },
+    #[snafu(display("project contains two similarly named networks: '{network}'"))]
+    DuplicateNetwork { network: String },
```
raymondk, line 532: `name` was a better name IMO.
I assume it would make the other errors weird though since both fields are also names....

lib/icp-network/src/structure.rs
```diff
@@ -7,22 +7,14 @@ pub struct NetworkDirectoryStructure {
...
         Self {
-            network_root,
-            port_descriptor_dir,
+            network_root: network_root.to_path_buf(),
+            port_descriptor_dir: port_descriptor_dir.to_path_buf(),
         }
     }
+}
 
-    pub fn network_root(&self) -> &Utf8Path {
```
raymondk, line 18: isn't it good practice to have getters anymore?
I assume that the fields in struct should not be `pub` though?

lib/icp-network/src/directory.rs
```diff
@@ -32,18 +34,16 @@ impl NetworkDirectory {
             structure,
         }
     }
+}
 
-    pub fn network_name(&self) -> &str {
-        &self.network_name
-    }
-
+impl NetworkDirectory {
```
raymondk, line 39: Looks like there are two `impl NetworkDirectory {}` blocks?
Also - you're getting rid of the getters and making private fields public purpose?

lib/icp-network/src/access.rs
```diff
@@ -1,64 +1,65 @@
...
+
+        // Specify url
+        let builder = builder.with_url(&self.url);
+
+        // Create agent
         let agent = builder
-            .with_arc_identity(identity) // would come from command-line or config
-            .with_ingress_expiry(timeout) // would come from network descriptor
+            .with_arc_identity(identity)
+            .with_ingress_expiry(Duration::from_secs(4 * MINUTE))
```
raymondk, line 54: I think those TODOs are still valid.
```suggestion
            .with_arc_identity(identity) // TODO would come from the command line or config
            .with_ingress_expiry(Duration::from_secs(4 * MINUTE))  // TODO would come from the network descriptor
```

lib/icp-network/src/access.rs
```diff
@@ -1,64 +1,65 @@
...
     pub default_effective_canister_id: Option<Principal>,
+
+    /// Network's root-key
     root_key: Option<Vec<u8>>,
-    routing: NetworkAccessRouting,
-}
 
-pub enum NetworkAccessRouting {
-    SingleUrl(String),
-    RouteProvider(Arc<dyn ic_agent::agent::route_provider::RouteProvider>),
```
raymondk, line 24: I think we should keep this. For mainnet,- we have to have to resolve api boundary nodes and pick one, we can't have everyone going through ic0.app

examples/icp-environments/icp.yaml
```diff
@@ -0,0 +1,28 @@
+canister:
```
raymondk, line 1: I'm not in love with the config file....

`canister` is weird as the only one that you can split to other files.
I can imagine a ton of duplication in the settings per environment - I suppose we can also make it split to different files later.

bin/icp-cli/src/commands/identity/default.rs
```diff
@@ -8,16 +8,23 @@ pub struct DefaultCmd {
...
-        change_default_identity(dirs, &list, &name)?;
-        println!("Set default identity to {name}");
-    } else {
-        let defaults = load_identity_defaults(dirs)?;
-        println!("{}", defaults.default);
+pub fn exec(ctx: &Context, cmd: DefaultCmd) -> Result<(), DefaultIdentityError> {
+    // Load project directories
+    let dirs = ctx.dirs();
+
+    match cmd.name {
```
raymondk, line 15: you just love matching instead of if/else

lib/icp-project/src/project.rs
rikonor: Per conversation with Raymond. Don't do this:
```yaml
canisters:
  <name-1>:
      ...

networks:
  <name>:
     ...

environments:
  <name: ...
     ...
```
but rather, do this:
```yaml
canisters:
  - name: ...
    ...

networks:
  - name: ...
     ...

environments:
  - name: ...
     ...
```
and then put in an explicit check for duplicates.

lib/icp-network/src/access.rs
raymondk: We agree to remove this for now - and add it later if it's needed.

lib/icp-project/src/project.rs
rikonor: I created `NETWORK_LOCAL` and `ENVIRONMENT_LOCAL` constants.

lib/icp-project/src/project.rs
rikonor: Personally, I like:
```rs
if let Some(network) = &e.network {
    if !networks.contains_key(network) {
        return Err(LoadProjectManifestError::EnvironmentNetworkDoesntExist {
            environment: e.name.to_owned(),
            network: network.to_owned(),
        });
    }
}
```
better than:
```rs
if let Some(name) = &e.network {
    if !networks.contains_key(name) {
        return Err(LoadProjectManifestError::EnvironmentNetworkDoesntExist {
            environment: e.name.to_owned(),
            network: name.to_owned(),
        });
    }
}
```
Do you feel that `name` or `network_name` is clearer?

lib/icp-project/src/project.rs
rikonor: Yeah, I just found myself settling on this pattern:
```
MissingCanister { environment: String, canister: String }
```
as opposed to:
```
MissingCanister { environment_name: String, canister_name: String }
```
which in the case of just one "thing" translates to:
```
MissingCanister { canister: String }
```

This works too:
```
MissingCanister { name: String }
```
but then it's a different pattern for the case where you have one "thing" or multiple "things" (for one thing you'd use `name` and for multiple you'd use `<description>`, so it's a bit inconsistent).

I can change it if you feel strongly about it.

lib/icp-network/src/structure.rs
rikonor: Fine to leave out?

lib/icp-network/src/directory.rs
rikonor: I was separating `impl` blocks by use-case (constructors, utilities, etc) as opposed to putting them all together.

lib/icp-project/src/project.rs
raymondk: Then why not use it?

PR #59, opened 2025-08-05
feat: introduce canister recipes

rikonor:
This change adds `recipe` to the canister manifest. It provides the base for recipes, and adds a few hard-coded recipes (which we can add to or remove later, if needed).

This is then followed by handlebars-based recipes in https://github.com/dfinity/icp-cli/pull/60.


lib/icp-network/src/manifest.rs
raymondk, line 1: I suppose this doesn't need to be here?

rikonor: Thanks, removed 👍

PR #63, opened 2025-08-22
feat: token management, deployment with cycles

viviveevee:
Uses a PocketIC binary from `master` in the monorepo. To fetch: `wget https://download.dfinity.systems/ic/75cc133e4407cfeb94a73dd623dc6f8398ae3c1e/binaries/x86_64-darwin/pocket-ic.gz`

### DIFF
--- a/bin/icp-cli/src/commands/canister/create.rs
+++ b/bin/icp-cli/src/commands/canister/create.rs
@@ -1,6 +1,8 @@
+use candid::{CandidType, Decode, Encode, Nat};
 use clap::Parser;
 use ic_agent::{AgentError, export::Principal};
 use ic_utils::interfaces::management_canister::LogVisibility;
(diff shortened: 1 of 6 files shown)


bin/icp-cli/src/commands/canister/create.rs
```diff
@@ -72,6 +74,10 @@ pub struct Cmd {
     #[clap(flatten)]
     pub settings: CanisterSettings,
 
+    /// How many cycles the canister should be created with. Also needs to pay for canister creation cost.
```
rikonor, line 77: Could we rework this docstring? Perhaps:
```
Amount of canister cycles to provide the created canister
```
What is meant by `Also needs to pay for canister creation cost.`? Is this referring to the fact that out of the attached cycles amount, a certain number of cycles will be used up immediately for the canister creation itself? Do we know the exact amount?

PR #68, opened 2025-09-02
chore: infer the effective canister id from the subnet id.

vincent-dfinity:
Ticket [SDK-2212](https://dfinity.atlassian.net/browse/SDK-2212)
- Infer the effective canister id from the subnet id. 
- The new `read_state_subnet_canister_ranges` on the Rust agent is used to get the canister id ranges from the subnet id.
- Use `--subnet-id` instead of `--effective-id` for `icp deploy`.
- Updated the tests.
- Bumped `ic-agent` and `ic-utils` to `v0.44.1`.

[SDK-2212]: https://dfinity.atlassian.net/browse/SDK-2212

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,9 +36,9 @@ glob = "0.3.2"
 handlebars = "6.3.2"
 hex = "0.4.3"
 httptest = "0.16.3"
-ic-agent = { version = "0.44.0" }
+ic-agent = { version = "0.44.1" }
 ic-asset = { version = "0.25.0" }
-ic-utils = { version = "0.44.0" }
+ic-utils = { version = "0.44.1" }
 icp-adapter = { path = "lib/icp-adapter" }
 icp-canister = { path = "lib/icp-canister" }
 icp-dirs = { path = "lib/icp-dirs" }
@@ -92,6 +92,3 @@ enum_variant_names = "allow"
 
 [profile.dev.package.scrypt]
 opt-level = 3
-
-[patch.crates-io]
-ic-agent = { version = "0.44.0", git = "https://github.com/dfinity/agent-rs", rev = "0884eac1930aeb47893928081ce000df4b53d3e0" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1988,8 +1988,9 @@ dependencies = [
 
 [[package]]
 name = "ic-agent"
-version = "0.44.0"
-source = "git+https://github.com/dfinity/agent-rs?rev=0884eac1930aeb47893928081ce000df4b53d3e0#0884eac1930aeb47893928081ce000df4b53d3e0"
+version = "0.44.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8faeb37240aa11d18f7fa3f42b26f64269212fed781bd10c5d280f5edec9529f"
 dependencies = [
  "arc-swap",
  "async-channel",
@@ -2011,7 +2012,7 @@ dependencies = [
  "http-body-util",
  "ic-certification",
  "ic-ed25519",
- "ic-transport-types 0.44.0",
+ "ic-transport-types 0.44.1",
  "ic-verify-bls-signature",
  "k256 0.13.4",
  "leb128",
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -83,3 +83,6 @@ enum_variant_names = "allow"
 
 [profile.dev.package.scrypt]
 opt-level = 3
+
+[patch.crates-io]
+ic-agent = { version = "0.44.0", git = "https://github.com/dfinity/agent-rs", rev = "0884eac1930aeb47893928081ce000df4b53d3e0" }
```
rikonor: Are we planning to include this, or is this PR blocked on a future `ic-agent` release?

bin/icp-cli/src/commands/deploy.rs
```diff
@@ -54,12 +54,50 @@ pub async fn exec(ctx: &Context, cmd: Cmd) -> Result<(), CommandError> {
...
+    // Check if no canister was selected.
+    if canisters.is_empty() {
+        if let Some(name) = &cmd.name {
+            // If a canister name was specified but not present.
             return Err(CommandError::CanisterNotFound {
                 name: name.to_owned(),
             });
+        } else {
+            // Should we at least warn here?
+            return Ok(());
```
rikonor: I suggest issuing a warning.

bin/icp-cli/src/commands/deploy.rs
```diff
@@ -54,12 +54,50 @@ pub async fn exec(ctx: &Context, cmd: Cmd) -> Result<(), CommandError> {
...
+        }
+    }
+
+    // Infer the effective canister id from the subnet id if provided.
+    let mut effective_id = Principal::from_text(DEFAULT_EFFECTIVE_ID).unwrap();
```
rikonor: What's `DEFAULT_EFFECTIVE_ID` at this point? Should we instead maintain a default subnet ID? E.g in the clap configuration you could specify a default subnet-id instead of making it an optional, and then you simply infer the effective canister ID based on the subnet id (either the defeault or user provided).

Cargo.toml
vincent-dfinity: I think we can include this. We can remove this patch once we have the future ic-agent release and ic-asset releases.

bin/icp-cli/src/commands/deploy.rs
vincent-dfinity: TBH, I don't really know what the default subnet id can be. 
After running the pocket-ic with `icp network run`, I noticed the default application subnet id is `gnkm6-o3f2j-s4j4o-tn4cp-ebkfd-46tuv-xaitz-fv54k-u7b2d-ejijp-vqe`. But I cannot find the subnet with the same id on this page https://dashboard.internetcomputer.org/network/subnets...

bin/icp-cli/src/commands/deploy.rs
vincent-dfinity: Sure, what's the right way to issue a warning in icp-cli? I didn't see there's a logging system...

Cargo.toml
rikonor: I don't have any objection to this, but might be good to get another person to sign off on it (@lwshang?)

bin/icp-cli/src/commands/deploy.rs
rikonor: There was a bit of refactoring done recently, so you may need to rebase main to have this, but I'd consider using:
```
let _ = ctx.term.write_line("<your message>");
```

bin/icp-cli/src/commands/deploy.rs
rikonor: Now, I wonder how we came up with the `DEFAULT_EFFECTIVE_ID`. I'm not so sure at this point 🙂

Cargo.toml
lwshang: > The new `read_state_subnet_canister_ranges` on the Rust agent is used to get the canister id ranges from the subnet id.

We can cut the ic-agent v0.44.1 release to include the new function. Then the patch here can be removed.
Since that will be a minor(patch) release, we won't have to release ic-asset.

bin/icp-cli/src/commands/deploy.rs
vincent-dfinity: Thanks. It took me a while to merge the main branch... Anyway this issue doesn't exist as it's handled on the main branch already.

bin/icp-cli/src/commands/deploy.rs
vincent-dfinity: I created a ticket to track it anyway, https://dfinity.atlassian.net/browse/SDK-2321. :)

Cargo.toml
vincent-dfinity: https://github.com/dfinity/agent-rs/pull/662 for v0.44.1.

vincent-dfinity: Fixed https://github.com/dfinity/agent-rs/pull/662.

PR #83, opened 2025-09-11
feat: frontend env variables example

ilbertt:
Blocked by #86

### DIFF
--- a/bin/icp-cli/src/commands/canister/binding_env_vars.rs
+++ b/bin/icp-cli/src/commands/canister/binding_env_vars.rs
@@ -1,9 +1,13 @@
+// This is a temporary placeholder command
+// For now it's only used to set environment variables
+// Eventually we will add support for canister settings operation
+
 use std::collections::HashSet;
 
 use clap::Parser;
 use futures::{StreamExt, stream::FuturesOrdered};
 use ic_agent::AgentError;
-use ic_utils::interfaces::management_canister::builders::CanisterInstallMode;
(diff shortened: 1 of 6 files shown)


examples/icp-frontend-environment-variables/frontend/canister.yaml
```diff
@@ -0,0 +1,16 @@
...
+      sha256: 2c9cfd9c111393ce3ff335307cda6fe38c6a474c926750d235e2a5ad96942795
+
+sync:
+  steps:
+    - type: script
+      command: pnpm build
+
+    - type: assets
+      dirs:
+        - app/dist
```
raymondk: I think the asset building should happen in the build phase.
And I didn't have vite installed, so this should make sure things are installed first.

```suggestion
  steps:
    - type: pre-built
      path: canister/assetstorage.wasm.gz
      sha256: 2c9cfd9c111393ce3ff335307cda6fe38c6a474c926750d235e2a5ad96942795

    - type: script
      command: pnpm install
      
    - type: script
      command: pnpm build

sync:
  steps:
    - type: assets
      dirs:
        - app/dist
```

ilbertt: Agree with moving the `pnpm build` script to the build step. `pnpm install` should be done right after cloning the repo, so it should be the responsibility of the dev. I think a proper README that explains it should be enough

ilbertt: Fixed

PR #86, opened 2025-09-12
feat: Inject canister ids as environment variables for generated bind…

raymondk:
…ings

### DIFF
--- a/bin/icp-cli/src/commands/canister/binding_env_vars.rs
+++ b/bin/icp-cli/src/commands/canister/binding_env_vars.rs
@@ -202,8 +202,7 @@ pub async fn exec(ctx: &Context, cmd: Cmd) -> Result<(), CommandError> {
                     "Update environment variables with new canister bindings: {:?}",
                     environment_variables
                 );
-                mgmt
-                    .update_settings(&cid)
+                mgmt.update_settings(&cid)
                     .with_environment_variables(environment_variables)
                     .await?;
 


bin/icp-cli/src/commands/canister/install.rs
```diff
@@ -122,6 +125,12 @@ pub async fn exec(ctx: &Context, cmd: Cmd) -> Result<(), CommandError> {
     // Prepare a futures set for concurrent operations
     let mut futs = FuturesOrdered::new();
 
+    // Get the list of name to canister id for this environment
+    // We need this to inject the `ICP_CANISTER_ID:` environment variables
+    // as we're installing the canisters
+    let canister_list = ctx.id_store.lookup_by_environment(&env.name)?;
```
lwshang: The current design of "Canister Linking via Env Vars" requires the canister IDs of all canisters in the project should be known when install_code.
Should we firstly check if all canisters in the project have been created?

bin/icp-cli/src/commands/canister/install.rs
```diff
@@ -174,7 +184,42 @@ pub async fn exec(ctx: &Context, cmd: Cmd) -> Result<(), CommandError> {
...
+                let update_canister_builder =
+                    mgmt.update_settings(&cid).with_environment_variables(
+                        var_map
+                            .iter()
+                            .map(|(n, v)| EnvironmentVariable {
+                                name: n.clone(),
+                                value: v.clone(),
+                            })
+                            .collect::<Vec<EnvironmentVariable>>(),
+                    );
```
lwshang: For the "update env vars" logic, I would suggest to let `icp` always create a new list which contains:
- The ones set in yaml config. (As in https://github.com/dfinity/icp-cli/pull/65)
- The `ICP_CANISTER_ID:...` ones for canister linking.

The `update_settings` operation simply overwrites the existing list with the new list.
This way, the "env var changes" of a canister can be "stateless". I mean the current state doesn't depend on the previous state. The current state can always be constructed directly from the icp yaml config and the canister IDs.

bin/icp-cli/src/commands/canister/install.rs
```diff
@@ -133,6 +142,7 @@ pub async fn exec(ctx: &Context, cmd: Cmd) -> Result<(), CommandError> {
             let cmd = cmd.clone();
             let mgmt = mgmt.clone();
             let pb = pb.clone();
+            let canister_list = canister_list.clone();
```
rikonor: Instead of calculating the necessary environment variables in the loop itself, why don't we calculate it immediately after we collect the canister list. Basically something like this:
```
// earlier in the install code
let cids = ctx.id_store.list(...) ...

// immediately after, calculate vars that will be applied to every canister
let vars = ...

for (_, c) in cs {
  let install_fn = {
    ...
    let cvars = ... + vars;
    // update the canister settings

    ... install
  };
}
```

bin/icp-cli/src/commands/canister/install.rs
```diff
@@ -174,7 +184,42 @@ pub async fn exec(ctx: &Context, cmd: Cmd) -> Result<(), CommandError> {
                     _ => panic!("invalid install mode"),
                 };
 
+                // Fetch the current environment variables
+                // So we can update them with the new mappings
+                debug!("Fetching environment variables");
```
rikonor: I had originally assumed we would do the variable injection as part of the deployment, but I guess it does make sense to do it as part of the install, since that handles both cases. Although, alternativelly, one could argue it should be part of canister creation as well (i.e create canisters, then update their environment variables). Not sure what the more correct approach is.

bin/icp-cli/src/commands/canister/settings.rs
```diff
@@ -0,0 +1,240 @@
...
+                    .unwrap_or_default();
+
+                // inject the ids of the other canisters
+                for (k, v) in binding_vars.iter() {
+                    environment_variables.insert(k.to_string(), v.to_string());
+                }
+
+                // Convert from HashMap<String, String> to Vec<EnvironmentVariable>
+                // as required by the IC management canister interface
+                let environment_variables = environment_variables
```
raymondk: This is not setting environment variables coming from the environment - neither is `create`

PR #91, opened 2025-09-18
feat: mint cycles

viviveevee:
* add `icp cycles mint`
* forward `icp cycles balance` to `icp token cycles balance`
* fix display of `icp token transfer` conclusion to respect token decimals

### DIFF
--- a/bin/icp-cli/tests/sync_tests.rs
+++ b/bin/icp-cli/tests/sync_tests.rs
@@ -86,9 +86,9 @@ fn sync_adapter_script_multiple() {
           sync:
             steps:
               - type: script
-                command: echo "first"
-              - type: script
                 command: echo "second"
+              - type: script
+                command: echo "first"
         "#,
     );
 


bin/icp-cli/src/commands/cycles/balance.rs
```diff
@@ -1,17 +1,23 @@
...
+    context::{Context, GetProjectError},
+};
 
-#[derive(Debug, Parser)]
-pub struct Cmd;
-
-pub async fn exec(_ctx: &Context, _: Cmd) -> Result<(), CommandError> {
-    unimplemented!()
+pub async fn exec(_ctx: &Context, cmd: token::balance::Cmd) -> Result<(), CommandError> {
+    token::balance::exec(_ctx, "cycles", cmd)
```
rikonor: Rename `_ctx` to `ctx`?

bin/icp-cli/src/commands/cycles/mint.rs
```diff
@@ -1,17 +1,274 @@
...
+use bigdecimal::ToPrimitive;
+use candid::{CandidType, Decode, Encode, Nat, Principal};
 use clap::Parser;
+use ic_agent::AgentError;
+use ic_ledger_types::AccountIdentifier;
+use ic_ledger_types::Memo;
+use ic_ledger_types::Subaccount;
+use ic_ledger_types::Tokens;
+use ic_ledger_types::TransferArgs;
+use ic_ledger_types::TransferResult;
```
rikonor: Consolidate the imports?

bin/icp-cli/src/commands/cycles/mint.rs
```diff
@@ -1,17 +1,274 @@
...
+    options::{EnvironmentOpt, IdentityOpt},
+};
+
+pub const MEMO_MINT_CYCLES: u64 = 0x544e494d; // == 'MINT'
 
 #[derive(Debug, Parser)]
-pub struct Cmd;
+pub struct Cmd {
+    #[clap(flatten)]
+    pub environment: EnvironmentOpt,
```
rikonor: What do you think about putting the "common" args at the end of the struct? Also, consider changing `clap(flatten)` to `command(flatten)` (according to the clap docs, the macros in clap 4 should be `arg` and `command`)

bin/icp-cli/src/commands/cycles/mint.rs
```diff
@@ -1,17 +1,274 @@
...
+
+    #[clap(flatten)]
+    pub identity: IdentityOpt,
+
+    #[clap(long = "icp-amount", conflicts_with = "cycles_amount")]
+    pub icp_amount: Option<BigDecimal>,
```
rikonor: What do you think about:
1. `icp cycles mint --icp <icp>` instead of `icp cycles mint --icp-amount <icp-amount>`
2. `icp cycles mint --cycles <cycles>` instead of `icp cycles mint --cycles-amount <cycles-amount>`
```
in which case you don't even have to specify the `long = X`, it can just be `arg(long)`.

File: bin/icp-cli/src/commands/cycles/mint.rs [hunk 5]
```diff
@@ -1,17 +1,274 @@
...
+        amount: Tokens::from_e8s(icp_e8s_to_deposit),
+        fee: Tokens::from_e8s(10_000),
+        from_subaccount: None,
+        to: account_id,
+        created_at_time: None,
+    };
+
+    let transfer_result = agent
+        .update(
+            &Principal::from_text("ryjl3-tyaaa-aaaaa-aaaba-cai").unwrap(),
```
rikonor: What do you think about putting these as constants under (e.g) `main.rs`? See [here](https://github.com/dfinity/icp-cli/pull/90/files#diff-a692c953c43c0e554fe90e55c5f03caf5f43746d2f88aaf8020c18e6cf07f669R27-R35).

bin/icp-cli/src/commands/cycles/mint.rs
```diff
@@ -1,17 +1,274 @@
...
+        .map_err(|e| CommandError::CanisterError {
+            canister: "ICP ledger".to_string(),
+            source: e,
+        })?;
+    let transfer_response = Decode!(&transfer_result, TransferResult).unwrap();
+    let block_index = match transfer_response {
+        Ok(block_index) => block_index,
+        Err(err) => {
+            match err {
+                ic_ledger_types::TransferError::TxDuplicate { duplicate_of } => duplicate_of,
```
rikonor: Replace `ic_ledger_types::X` with just `X`?

bin/icp-cli/src/commands/cycles/mint.rs
```diff
@@ -1,17 +1,274 @@
...
+                ic_ledger_types::TransferError::InsufficientFunds { balance } => {
+                    let required = BigDecimal::new((icp_e8s_to_deposit + 10_000).into(), 8); // transfer fee
```
rikonor: Make `10_000` into a constant? See [here](https://github.com/dfinity/icp-cli/pull/90/files#diff-be9c2b577061eb5b4e73555fe68a91e046145927102be4934964be03f775b356R28-R30).

bin/icp-cli/src/commands/cycles/mint.rs
```diff
@@ -1,17 +1,274 @@
...
+                err => {
+                    return Err(CommandError::TransferError { src: err });
+                }
+            }
+        }
+    };
+
+    let notify_response = agent
+        .update(
+            &Principal::from_text("rkp4c-7iaaa-aaaaa-aaaca-cai").unwrap(),
```
rikonor: Make into a constant?

bin/icp-cli/src/commands/cycles/mint.rs
```diff
@@ -1,17 +1,274 @@
...
+            "notify_mint_cycles",
+        )
+        .with_arg(
+            Encode!(&NotifyMintCyclesArgs {
+                block_index,
+                deposit_memo: None,
+                to_subaccount: None,
+            })
+            .unwrap(),
```
rikonor: Replace with an actual error? [Example](https://github.com/dfinity/icp-cli/pull/90/files#diff-be9c2b577061eb5b4e73555fe68a91e046145927102be4934964be03f775b356R324)

bin/icp-cli/src/commands/cycles/mint.rs
```diff
@@ -1,17 +1,274 @@
...
+    // display
+    let deposited = BigDecimal::new((minted.minted - 100_000_000_u128).into(), 12); // deposit charges 100M fee
+    let new_balance = BigDecimal::new(minted.balance.into(), 12);
+    println!("Minted {deposited} TCYCLES to your account, new balance: {new_balance} TCYCLES.");
+
+    Ok(())
+}
+
+#[derive(Debug, Deserialize, CandidType)]
+struct CmcResponse {
```
rikonor: I assume Cmc means cycles minting canister, but it's not really clear offhand. I would consider improving the name, or just adding appropriate comments. Example, I renamed these to [this](https://github.com/dfinity/icp-cli/pull/90/files#diff-be9c2b577061eb5b4e73555fe68a91e046145927102be4934964be03f775b356R56-R64).

Similarly, for the other structs, I feel like the names are overly long, e.g `NotifyMintCyclesX`, could also just be `NotifyX` (i.e `NotifyOk`, `NotifyError`, `NotifyResponse`, etc).

bin/icp-cli/src/commands/token/transfer.rs
```diff
@@ -122,16 +122,17 @@ pub async fn exec(ctx: &Context, token: &str, cmd: Cmd) -> Result<(), CommandErr
     );
 
     // Calculate units of token to transfer
-    let amount = cmd.amount.clone() * 10u128.pow(decimals);
+    let units_amount = cmd.amount.clone() * 10u128.pow(decimals);
```
rikonor: What are `units`? I can see how `amount` isn't very clear, but could we instead name this more clearly, like `token_amount` or `cycles_amount`, as `units` isn't much clearer.

bin/icp-cli/tests/common/context/cycles_ledger.rs
```diff
@@ -0,0 +1,30 @@
...
+                    Principal::anonymous(),
+                    "icrc1_balance_of",
+                    Encode!(&icrc_ledger_types::icrc1::account::Account { owner, subaccount })
+                        .unwrap(),
+                )
+                .await
+                .unwrap(),
+            Nat
+        )
+        .unwrap()
```
rikonor: What do you think about formatting this snippet slightly differently, so it's less indented?

bin/icp-cli/src/commands/cycles/mint.rs
```diff
@@ -1,17 +1,274 @@
...
+        NotifyMintCyclesResponse::Ok(ok) => ok,
+        NotifyMintCyclesResponse::Err(err) => {
+            return Err(CommandError::NotifyMintCyclesError { src: err });
+        }
+    };
+
+    // display
+    let deposited = BigDecimal::new((minted.minted - 100_000_000_u128).into(), 12); // deposit charges 100M fee
+    let new_balance = BigDecimal::new(minted.balance.into(), 12);
+    println!("Minted {deposited} TCYCLES to your account, new balance: {new_balance} TCYCLES.");
```
rikonor: Could you switch to using the `console` crate, like we do in other commands, for printing user-facing information?

bin/icp-cli/tests/common/context/cycles_ledger.rs
```diff
@@ -0,0 +1,30 @@
...
+use pocket_ic::nonblocking::PocketIc;
+use std::cell::Ref;
+
+const CYCLES_LEDGER_ID: Principal = Principal::from_slice(&[0, 0, 0, 0, 2, 16, 0, 2, 1, 1]); // um5iw-rqaaa-aaaaq-qaaba-cai
+
+pub struct CyclesLedgerPocketIcClient<'a> {
+    pub pic: Ref<'a, PocketIc>,
+}
+
+impl CyclesLedgerPocketIcClient<'_> {
```
rikonor: Is a pocket-ic specific ledger client necessary, or would a simple utility function using an ic-agent sufficient?

bin/icp-cli/tests/common/context/icp_shorthand.rs
```diff
@@ -0,0 +1,57 @@
...
+pub struct IcpShorthand<'a> {
+    ctx: &'a TestContext,
+}
+
+impl<'a> IcpShorthand<'a> {
+    pub fn new(ctx: &'a TestContext) -> Self {
+        Self { ctx }
+    }
+
+    pub fn active_principal(&self) -> Principal {
```
rikonor, line 14: Do you have any thoughts on simplifying these by making them into standalone functions that take whatever `self.ctx.icp()` returns? Then those can be used on demand, rather than requiring instantiating an entire struct that holds a reference to TestContext.

bin/icp-cli/tests/common/context/icp_ledger.rs
```diff
@@ -0,0 +1,56 @@
...
+                "icrc1_transfer",
+                Encode!(&icrc_ledger_types::icrc1::transfer::TransferArg {
+                    from_subaccount: None,
+                    to: icrc_ledger_types::icrc1::account::Account { owner, subaccount },
+                    fee: None,
+                    created_at_time: None,
+                    memo: None,
+                    amount: amount.into(),
+                })
+                .unwrap(),
```
rikonor: Some thoughts on making this more readable:
```
let arg = TransferArg {
  ...
};

let bs = Encode!(...)?;

self.pic.update_call(
  LEDGER_ID,            // sender
  GOVERNANCE_ID, // proxy-canister
  ...,
)?;
```

bin/icp-cli/tests/token_tests.rs
```diff
@@ -0,0 +1,140 @@
...
+        1_000_000_000_u128
+    );
+    assert_eq!(
+        ctx.icp_ledger().balance_of(bob_principal, None).await,
+        0_u128
+    );
+
+    // Simple ICP transfer
+    ctx.icp_().use_identity("alice");
+    ctx.icp()
```
rikonor, line 92: It's a bit confusing that we have `icp_()` and `icp()` now. Could we discuss that? What do you think about something like this (similarly to what I mention in another comment about converting these to utility functions):
```
use_identity(ctx.icp(), "alice")
```

bin/icp-cli/src/commands/cycles/mint.rs
viviveevee: makes sense to move them further down, they're not very interesting

bin/icp-cli/src/commands/cycles/mint.rs
viviveevee: 👍 using `main.rs` for this feels a little weird, but I don't have a better suggestion

viviveevee: main is not accessible from tests, so the constants are in `tests/common/mod.rs` too now

bin/icp-cli/src/commands/cycles/mint.rs
viviveevee: I'm using `NotifyMintX` now. `NotifyX` is a bit too short IMO since the CMC has three different `notify_x` endpoints

bin/icp-cli/src/commands/token/transfer.rs
viviveevee: It's now `ledger_amount` with a comment that explains the difference between the user-supplied amount and what the ledger expects

bin/icp-cli/tests/common/context/cycles_ledger.rs
viviveevee: As-is it's not required for the cycles ledger helper, but it allows us to use caller impersonation in the future. We require this e.g. for minting ICP in the ICP ledger helper, and I bet it will come in handy at other times too.

We could mix styles (some helpers use ic-agent, some use pocket-ic), but I prefer the consistency since the two interfaces have subtle differences.

bin/icp-cli/tests/common/context/icp_shorthand.rs
viviveevee: I have an opinion, but don't hold it strongly.

* Where would you use those where the current version would not fit? TestContext gives isolation from other tests, which we want to have in unit tests anyways. Your "[...] standalone functions that take whatever `self.ctx.icp()` returns" implies that TestContext is instantiated already.
* TestContext is not super cheap, but it is cheap since it only creates a tempdir. Avoiding that for unit tests feels to me like a premature optimization. 
* Discoverability through autocomplete suffers. Now I can type `ctx.icp_().` and get a great list of what is there. That won't work if it's standalone functions. Also writing that extra argument is tedious

bin/icp-cli/tests/token_tests.rs
viviveevee: Same argument as above, I like having it autocompletable. But I am all for renaming if you have a better idea. I lifted the idea for `icp()` vs `icp_()` from Haskell's `mapM` and `mapM_`.

Some ideas (always manual command first, shorthand second):
* `icp()` and `do()`
* `icp()` and `icp_auto()`
* `command()` and `icp()`
* `run()` and `do()`

bin/icp-cli/src/commands/cycles/mint.rs
rikonor: But is the non-test code pulling from `tests/common/mod.rs`? That seems backward... I'd consider just putting them in main.rs but then for the tests either duplicating them, or using hardcoded values instead of constants. More important, I think, that the non-test code can refer to constants in a reasonable place (agreed that main.rs is not ideal, but I don't have a better idea either yet).

bin/icp-cli/src/commands/cycles/mint.rs
rikonor: Ah, that makes perfect sense! 👍

bin/icp-cli/tests/token_tests.rs
rikonor: > always manual command first, shorthand second

I think the issue is with not understanding what `shorthand` means in this context. I agree about the autocomplete, but that seems like you'd have that from a simple module name as well, no? e.g `util::X`. I don't feel that strongly about it, either, only about it being a bit clearer what is the purpose of IcpShorthand, because on its own that name does not convey any information (maybe some comments in the code would be helpful, or perhaps just a clearer name).

bin/icp-cli/tests/common/context/icp_ledger.rs
```diff
@@ -0,0 +1,57 @@
...
+use icrc_ledger_types::icrc1::{
+    account::{Account, Subaccount},
+    transfer::TransferArg,
+};
+use pocket_ic::nonblocking::PocketIc;
+use std::cell::Ref;
+
+use crate::common::{GOVERNANCE_ID, ICP_LEDGER_CID};
+
+pub struct IcpLedgerPocketIcClient<'a> {
```
rikonor, line 11: Perhaps we could put these under a `clients` module (potentially under `tests`, or even under `icp-network/pocket-ic`), then call these `IcpLedger`, CyclesLedger`, etc. That would put them in a more appropriate location, and calling them would look like: `pocket_ic::IcpLedger` or `clients::IcpLedger`. It's a bit of a nitpick, but if we're going to add more, perhaps it's worth considering this in advance.

bin/icp-cli/tests/common/context/cycles_ledger.rs
rikonor: Regarding caller impersonation, is that not possible simply with ic-agent, by creating multiple identities and multiple agents?

bin/icp-cli/tests/common/context/icp_shorthand.rs
```diff
@@ -0,0 +1,57 @@
...
+            .assert()
+            .success();
+        self.ctx
+            .icp()
+            .args(["identity", "default", &random_name])
+            .assert()
+            .success();
+    }
+
+    pub fn create_identity(&self, name: &str) {
```
vincent-dfinity, line 28: It would be good to have `pub fn get_principal(&self, name: &str) -> Principal` to get the principal without using it.

bin/icp-cli/tests/common/context/cycles_ledger.rs
viviveevee: Not if you need it to be a specific principal, e.g. the governance canister. It's not possible to craft a private key for that since it's a canister, and not a self-authenticating principal. Not even if you allow enumerating all possible private keys since those end up with a different suffix in the principal bytes.

If you simply need multiple distinct identities then ic-agent is good enough

bin/icp-cli/src/commands/cycles/mint.rs
viviveevee: I put the constants both in `main.rs` (for non-test code) and `tests/common/mod.rs` (for test code). Hardcoded would work for tests too, but I prefer the readability of the constants

bin/icp-cli/tests/common/context/icp_shorthand.rs
viviveevee: Feel free to add it when you need it

bin/icp-cli/tests/common/context/icp_ledger.rs
viviveevee: I tried moving it to `icp-network/`, but then we have no way to get to the canister id constants anymore 😅  How do you like the separate `clients` module I put the clients in now?

bin/icp-cli/tests/token_tests.rs
viviveevee: How do you feel about `ctx.icp()` and `clients::icp_client(&ctx)`? I think this looks relatively clean

rikonor: That seems OK to me, although I think you're fine with just `clients::icp(&ctx)`

bin/icp-cli/src/commands/cycles/mint.rs
```diff
@@ -1,17 +1,278 @@
...
 
-use crate::context::{Context, GetProjectError};
+use crate::{
+    CYCLES_MINTING_CANISTER_CID, ICP_LEDGER_CID,
+    context::{Context, ContextGetAgentError, GetProjectError},
+    options::{EnvironmentOpt, IdentityOpt},
+};
+
+pub const MEMO_MINT_CYCLES: u64 = 0x544e494d; // == 'MINT'
+const ICP_TRANSFER_FEE: u64 = 10_000;
```
rikonor, line 19: Add a comment here? What is this transfer fee? Is it in cycles or ICP?

bin/icp-cli/src/commands/cycles/mint.rs
```diff
@@ -1,17 +1,278 @@
...
+            .await
+            .map_err(|e| CommandError::CanisterError {
+                canister: "cmc".to_string(),
+                source: e,
+            })?;
+
+        let cmc_response =
+            Decode!(&cmc_response, ConversionRateResponse).expect("CMC response type changed");
+        let cycles_per_e8s = cmc_response.data.xdr_permyriad_per_icp as u128;
+        let cycles_plus_fees = cycles_amount + 100_000_000_u128; // Cycles ledger charges 100M for deposits
```
rikonor, line 92: Make the `100_000_000_u128` into a constant with a descriptive comment?

PR #101, opened 2025-09-25
chore: await output to be printed before exiting

viviveevee:
GPT-5 suggests `sync_tests.rs` is flaky because we do not wait for the build output print task to finish printing all output before considering command execution to be done.

This change returns a handle to await before `icp canister build` or `icp sync` are allowed to claim their work is finished. This way `icp` should not exit before everything is printed

### DIFF
--- a/bin/icp-cli/src/commands/identity/list.rs
+++ b/bin/icp-cli/src/commands/identity/list.rs
@@ -1,6 +1,7 @@
 use crate::context::Context;
 use clap::Parser;
 use icp_identity::manifest::{load_identity_defaults, load_identity_list};
+use itertools::Itertools;
 use snafu::Snafu;
 
 #[derive(Debug, Parser)]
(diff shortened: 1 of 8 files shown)


bin/icp-cli/src/commands/build/mod.rs
```diff
@@ -90,13 +90,19 @@ pub async fn exec(ctx: &Context, cmd: Cmd) -> Result<(), CommandError> {
                     match step {
                         // Compile using the custom script adapter.
                         BuildStep::Script(adapter) => {
-                            // Setup script progress handling
-                            let tx = script_handler.setup_output_handler();
+                            // Setup script progress handling and receiver join handle
+                            let (tx, recv_handle) = script_handler.setup_output_handler();
```
rikonor: nitpick, but consider `rx` (consistent with `tx`)

bin/icp-cli/src/progress.rs
```diff
@@ -4,6 +4,7 @@ use futures::Future;
 use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
 use itertools::Itertools;
 use tokio::sync::mpsc;
+use tokio::task::JoinHandle;
```
rikonor: Consolidate imports?

bin/icp-cli/src/commands/build/mod.rs
rikonor: Oh, that's not a receiver, it's an actual joinhandle, my bad.

PR #103, opened 2025-09-25
chore: add a prelude to alias camino types to stdlib names

rikonor:
This change tries to solve a minor pain point of manually importing and working with `camino` types.

Since `camino::Utf8PathBuf` is meant to replace `PathBuf`, it makes sense to simply alias the two so that we can use them interchangeably. This can be done via `use camino::Utf8PathBuf as PathBuf` in the various files that need it.

A more convenient option is to set up a prelude to do the aliasing, and then we can use it via `use icp::prelude::*` in the various files that need it. This can also be augmented with other patterns we want to follow later on.

The change does two things:
1. Sets up an `icp::prelude` module, with a few simple camino-specific aliases.
2. Update the various references to Utf8PathBuf, etc, to use the aliases (i.e stdlib-named) versions instead.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,8 +11,6 @@ env:
   CARGO_NET_RETRY: 10
   # Use the local .curlrc
   CURL_HOME: .
-  # Disable DFX telemetry
-  DFX_TELEMETRY: "off"
 
 jobs:
   discover:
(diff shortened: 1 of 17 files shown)


crates/icp-project/src/lib.rs
```diff
@@ -174,10 +174,11 @@ impl Project {
                                 .context(GlobWalkSnafu { path: &pattern })?;
 
                             // Convert to Utf8 paths
-                            let paths = paths
-                                .into_iter()
-                                .map(Utf8PathBuf::try_from)
-                                .collect::<Result<Vec<_>, _>>()?;
+                            let paths = paths.into_iter().map(PathBuf::try_from).collect::<Result<
```
raymondk, line 177: I'm surprised cargo fmt is happy with this...

rikonor: 🤷

PR #104, opened 2025-09-26
chore: remove dependency on dfx from tests

raymondk:
Since we can start a local network and there tests for that we don't need to rely on dfx anymore.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,8 +11,6 @@ env:
   CARGO_NET_RETRY: 10
   # Use the local .curlrc
   CURL_HOME: .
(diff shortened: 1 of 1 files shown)


crates/icp-cli/tests/network_tests.rs
```diff
@@ -9,93 +8,10 @@ use serial_test::file_serial;
 mod common;
 use crate::common::{TestContext, TestNetwork};
 
-#[test]
-fn network_random_port() {
```
raymondk, line 13: This test seems unnecessary since we do pretty much the same thing in the deploy_tests without a depenency on dfx.

PR #106, opened 2025-09-26
chore: move fs portion of icp-fs crate into icp::fs module

rikonor:
This change:
1. Brings some utilities from the `icp-fs` into several new modules under `icp`: `icp::fs`, `icp::fs::json`, and `icp::fs::yaml`.
2. Moves the LockFile construct into `icp-network`, as it's only used by that, atm.
3. Removes the `icp-fs` crate.

### DIFF
--- a/crates/icp-cli/src/store_artifact.rs
+++ b/crates/icp-cli/src/store_artifact.rs
@@ -1,3 +1,5 @@
+use std::sync::Mutex;
+
 use icp::{
     fs::{create_dir_all, read, write},
     prelude::*,
@@ -22,35 +24,53 @@ pub enum LookupError {
     LookupArtifactNotFound { name: String },
 }
 
-pub struct ArtifactStore(PathBuf);
+pub struct ArtifactStore {
+    path: PathBuf,
+    lock: Mutex<()>,
+}
 
 impl ArtifactStore {
-    pub fn new(path: &PathBuf) -> Self {
-        Self(path.clone())
+    pub fn new(path: &Path) -> Self {
+        Self {
+            path: path.to_owned(),
+            lock: Mutex::new(()),
+        }
     }
 }
 
 impl ArtifactStore {
     pub fn save(&self, name: &str, wasm: &[u8]) -> Result<(), SaveError> {
+        // Lock Artifact Store
+        let _g = self
+            .lock
+            .lock()
+            .expect("failed to acquire artifact store lock");
+
         // Create artifacts directory
-        create_dir_all(&self.0).context(ArtifactsDirSnafu)?;
+        create_dir_all(&self.path).context(ArtifactsDirSnafu)?;
 
         // Store artifact
-        write(&self.0.join(name), wasm).context(SaveWriteFileSnafu)?;
+        write(&self.path.join(name), wasm).context(SaveWriteFileSnafu)?;
 
         Ok(())
     }
 
     pub fn lookup(&self, name: &str) -> Result<Vec<u8>, LookupError> {
+        // Lock Artifact Store
+        let _g = self
+            .lock
+            .lock()
+            .expect("failed to acquire artifact store lock");
+
         // Not Found
-        if !self.0.join(name).exists() {
+        if !self.path.join(name).exists() {
             return Err(LookupError::LookupArtifactNotFound {
                 name: name.to_owned(),
             });
         }
 
         // Load artifact
-        let wasm = read(&self.0.join(name)).context(LookupReadFileSnafu)?;
+        let wasm = read(&self.path.join(name)).context(LookupReadFileSnafu)?;
 
         Ok(wasm)
     }
--- a/crates/icp-cli/src/store_id.rs
+++ b/crates/icp-cli/src/store_id.rs
@@ -1,4 +1,4 @@
-use std::io::ErrorKind;
+use std::{io::ErrorKind, sync::Mutex};
 
 use ic_agent::export::Principal;
 use icp::{fs::json, prelude::*};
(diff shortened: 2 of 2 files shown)


crates/icp-canister/src/handlebars.rs
```diff
@@ -1,8 +1,9 @@
 use std::{collections::HashMap, str::FromStr, string::FromUtf8Error};
 
 use async_trait::async_trait;
-use handlebars::*;
-use icp_fs::fs::{ReadFileError, read};
+use handlebars::{Context, Helper, HelperDef, HelperResult, Output};
+// use handlebars::*;
```
raymondk: Remove this?

crates/icp/src/fs/json.rs
```diff
@@ -0,0 +1,47 @@
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    fs::{read, write_string},
+    prelude::*,
+};
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
raymondk, line 9: Do those errors have enough detail if the deserialialisation fails? (File name, line with the error, etc...)

crates/icp/src/fs/json.rs
```diff
@@ -0,0 +1,47 @@
...
+
+pub fn load<T: for<'a> Deserialize<'a>>(path: &Path) -> Result<T, Error> {
+    serde_json::from_slice(&read(path)?).map_err(|err| Error::Parse {
+        path: path.into(),
+        err,
+    })
+}
+
+pub fn save<T: Serialize>(path: &Path, value: &T) -> Result<(), Error> {
+    write_string(
```
raymondk, line 38: We don't need to lock the file before writing it?

crates/icp/src/fs/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+pub fn write(path: &Path, contents: &[u8]) -> Result<(), Error> {
+    std::fs::write(path, contents).map_err(|err| Error {
+        path: path.into(),
+        err,
+    })
+}
+
+pub fn write_string(path: &Path, contents: &str) -> Result<(), Error> {
```
raymondk, line 73: Same question with the locking. Isn't it necessary?

crates/icp/src/fs/yaml.rs
```diff
@@ -0,0 +1,24 @@
...
+    #[error("failed to parse json file at {path}")]
+    Parse {
+        path: PathBuf,
+
+        #[source]
+        err: serde_yaml::Error,
+    },
+}
+
+pub fn load<T: for<'a> Deserialize<'a>>(path: &Path) -> Result<T, Error> {
```
raymondk, line 19: Same as with the json, is this going to give enough details about the error if parsing fails?

crates/icp-canister/src/handlebars.rs
rikonor: Thanks, removed.

crates/icp/src/fs/json.rs
rikonor: Regarding file name/path, yes, each error has that in it:
```
#[derive(Debug, thiserror::Error)]
pub enum Error {
    #[error(transparent)]
    Io(#[from] crate::fs::Error),

    #[error("failed to parse json file at {path}")]
    Parse {
        path: PathBuf,
        ...
    },

    #[error("failed to serialize json file to {path}")]
    Serialize {
        path: PathBuf,
        ...
    },
}
```
What do you mean by "line with the error"?

crates/icp/src/fs/yaml.rs
rikonor: Yep, same - each error has a `path` (`icp::fs::Error` does too), so the file path, etc, should be included in the error.

crates/icp/src/fs/json.rs
rikonor: As far as I can tell, we never attempt to perform concurrent read/writes on this file, so this is not an issue.

crates/icp/src/fs/mod.rs
rikonor: Same answer as https://github.com/dfinity/icp-cli/pull/106#discussion_r2383205611.

crates/icp/src/fs/json.rs
raymondk: If the json/yaml are invalid - does the user get some information about the reason why the yaml was not parsed?

raymondk: it looks like yes:
```❯ icp build
Error: failed to parse json file at /Users/raymond/dev/dfinity/icp-cli/examples/icp-frontend-environment-variables/icp.yaml

Caused by:
    could not find expected ':' at line 4 column 1, while scanning a simple key at line 2 column 1```

File: crates/icp/src/fs/yaml.rs [hunk 6]
```diff
@@ -0,0 +1,24 @@
+use serde::Deserialize;
+
+use crate::{fs::read, prelude::*};
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error(transparent)]
+    Io(#[from] crate::fs::Error),
+
+    #[error("failed to parse json file at {path}")]
```
raymondk:
```suggestion
    #[error("failed to parse yaml file at {path}")]
```

crates/icp/src/fs/json.rs
raymondk: isn't it better to be safe for writes in case commands are being run concurrently or different threads end up trying to make changes to the same file?

crates/icp/src/fs/mod.rs
raymondk: Maybe we can reuse the fd_lock crate and hide the lock complexity behind these functions?
The build steps for eg use these functions and they run in parallel so there is a chance they could corrupt each other's data depending on how the build steps are defined.

rikonor: My preference is put a lock on a resource directly, rather than have a mechanism to lock any generic path. See https://github.com/dfinity/icp-cli/pull/106/commits/225e4522f38b3dadcb5b776a06cfc4b749251e1f.

crates/icp/src/fs/json.rs
rikonor: Added a lock to the id store and artifact store (See https://github.com/dfinity/icp-cli/pull/106/commits/225e4522f38b3dadcb5b776a06cfc4b749251e1f)

PR #107, opened 2025-09-26
chore: clean up icp-dirs and move it under icp::Directories

rikonor:
This change refines the utilities in `icp-dirs` and then relocates them to the `icp` crate, making them accessible under `icp::Directories`.


crates/icp/src/directories.rs
```diff
@@ -0,0 +1,77 @@
+use crate::prelude::*;
+use directories::ProjectDirs;
+
```
raymondk, line 9: Could use some comments to explain the intent

crates/icp/src/directories.rs
```diff
@@ -0,0 +1,77 @@
...
+        .ok_or(DirectoriesError::LocateHome)?;
+
+        // Convert to utf8 paths
+        let dirs = DirectoriesInner::from_dirs(dirs)?;
+
+        Ok(Self::Standard(dirs))
+    }
+}
+
+impl Directories {
```
raymondk, line 98: Docs for what each of those is supposed to be ?

PR #108, opened 2025-09-29
feat: create canisters via cycles ledger

viviveevee:
Subnet selection will be implemented in a follow-up PR.

### DIFF
--- a/crates/icp-cli/tests/common/clients.rs
+++ b/crates/icp-cli/tests/common/clients.rs
@@ -1,22 +1,19 @@
-use cycles_ledger::CyclesLedgerPocketIcClient;
 use icp::prelude::*;
-pub use icp_cli::IcpCliClient;
-use icp_ledger::IcpLedgerPocketIcClient;
 
 use crate::common::TestContext;
 
-mod cycles_ledger;
-mod icp_cli;
-mod icp_ledger;
+pub mod cycles_ledger;
+pub mod icp_cli;
+pub mod ledger;
 
-pub fn icp(ctx: &TestContext, current_dir: impl Into<PathBuf>) -> IcpCliClient<'_> {
-    IcpCliClient::new(ctx, current_dir.into())
+pub fn icp(ctx: &TestContext, current_dir: impl Into<PathBuf>) -> icp_cli::Client<'_> {
+    icp_cli::Client::new(ctx, current_dir.into())
 }
 
-pub fn icp_ledger(ctx: &TestContext) -> IcpLedgerPocketIcClient<'_> {
-    IcpLedgerPocketIcClient::new(ctx)
+pub fn ledger(ctx: &TestContext) -> ledger::Client<'_> {
+    ledger::Client::new(ctx)
 }
 
-pub fn cycles_ledger(ctx: &TestContext) -> CyclesLedgerPocketIcClient<'_> {
-    CyclesLedgerPocketIcClient::new(ctx)
+pub fn cycles_ledger(ctx: &TestContext) -> cycles_ledger::Client<'_> {
+    cycles_ledger::Client::new(ctx)
 }
--- a/crates/icp-cli/tests/common/clients/icp_cli.rs
+++ b/crates/icp-cli/tests/common/clients/icp_cli.rs
@@ -3,12 +3,12 @@ use icp::prelude::*;
 
 use crate::common::TestContext;
 
-pub struct IcpCliClient<'a> {
+pub struct Client<'a> {
     ctx: &'a TestContext,
     current_dir: PathBuf,
 }
 
-impl<'a> IcpCliClient<'a> {
+impl<'a> Client<'a> {
     pub fn new(ctx: &'a TestContext, current_dir: PathBuf) -> Self {
         Self { ctx, current_dir }
     }
--- a/crates/icp-cli/tests/canister_settings_tests.rs
+++ b/crates/icp-cli/tests/canister_settings_tests.rs
@@ -1,6 +1,6 @@
 use crate::common::{
     TestContext,
-    clients::{self, IcpCliClient},
+    clients::{self, icp_cli},
 };
 use icp::{fs::write_string, prelude::*};
 use predicates::{
@@ -247,7 +247,7 @@ fn canister_settings_update_controllers() {
         );
 }
 
-fn get_principal(client: &IcpCliClient, identity: &str) -> String {
+fn get_principal(client: &icp_cli::Client<'_>, identity: &str) -> String {
     client.create_identity(identity);
     client.get_principal(identity).to_string()
 }
--- a/crates/icp-cli/tests/common/clients/cycles_ledger.rs
+++ b/crates/icp-cli/tests/common/clients/cycles_ledger.rs
@@ -6,11 +6,11 @@ use std::cell::Ref;
 
 use crate::common::TestContext;
 
-pub struct CyclesLedgerPocketIcClient<'a> {
+pub struct Client<'a> {
     pic: Ref<'a, PocketIc>,
 }
 
-impl<'a> CyclesLedgerPocketIcClient<'a> {
+impl<'a> Client<'a> {
     pub fn new(ctx: &'a TestContext) -> Self {
         Self {
             pic: ctx.pocketic(),
--- a/crates/icp-cli/tests/common/clients/ledger.rs
+++ b/crates/icp-cli/tests/common/clients/ledger.rs
@@ -9,11 +9,11 @@ use std::cell::Ref;
 
 use crate::common::TestContext;
 
-pub struct IcpLedgerPocketIcClient<'a> {
+pub struct Client<'a> {
     pic: Ref<'a, PocketIc>,
 }
 
-impl<'a> IcpLedgerPocketIcClient<'a> {
+impl<'a> Client<'a> {
     pub fn new(ctx: &'a TestContext) -> Self {
         Self {
             pic: ctx.pocketic(),
--- a/crates/icp-cli/tests/cycles_tests.rs
+++ b/crates/icp-cli/tests/cycles_tests.rs
@@ -35,7 +35,7 @@ async fn cycles_balance() {
         .success();
 
     // Mint ICP to cycles, specify ICP amount
-    clients::icp_ledger(&ctx)
+    clients::ledger(&ctx)
         .mint_icp(identity, None, 123456789_u64)
         .await;
     ctx.icp()
@@ -49,7 +49,7 @@ async fn cycles_balance() {
 
     // Mint ICP to cycles, specify cycles amount
     let identity = clients::icp(&ctx, &project_dir).use_new_random_identity();
-    clients::icp_ledger(&ctx)
+    clients::ledger(&ctx)
         .mint_icp(identity, None, 123456789_u64)
         .await;
     ctx.icp()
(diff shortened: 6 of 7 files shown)


crates/icp-cli/src/commands/canister/create.rs
```diff
@@ -48,14 +31,6 @@ pub struct CanisterSettings {
...
     #[arg(long)]
     pub reserved_cycles_limit: Option<u64>,
-
-    /// Optional Wasm memory limit in bytes. Sets an upper bound for Wasm heap growth.
-    #[arg(long)]
-    pub wasm_memory_limit: Option<u64>,
-
-    /// Optional Wasm memory threshold in bytes. Triggers a callback when exceeded.
-    #[arg(long)]
-    pub wasm_memory_threshold: Option<u64>,
```
viviveevee, line 58: Not supported by the cycles ledger (yet)

crates/icp-cli/src/commands/canister/create.rs
```diff
@@ -84,6 +55,117 @@ pub struct Cmd {
...
     #[arg(long, short = 'q')]
     pub quiet: bool,
+
+    /// Cycles to fund canister creation (in raw cycles).
+    #[arg(long, default_value_t = 2_000_000_000_000u128)]
+    pub cycles: u128,
+}
+
+#[derive(Clone, Debug, CandidType, Deserialize)]
+struct CanisterSettingsArg {
```
viviveevee: Shall we factor all these canister interface types into a new module? Maybe `icp::canister_interfaces::cycles_ledger`?

crates/icp-cli/src/commands/canister/create.rs
```diff
@@ -84,6 +55,117 @@ pub struct Cmd {
     /// Suppress human-readable output; print only canister IDs, one per line, to stdout.
     #[arg(long, short = 'q')]
     pub quiet: bool,
+
+    /// Cycles to fund canister creation (in raw cycles).
+    #[arg(long, default_value_t = 2_000_000_000_000u128)]
```
rikonor: What do you think about putting this behind a constant? I think it would be nice to have that constant in a visible place (same as the other prominent canister IDs we have).

rikonor: It honestly might be nice to put something like: `MILLION`, `BILLION`, `TRILLION`, etc in the prelude, so that it's available everywhere, then the constant become `2 * TRILLION`. Just a thought.

crates/icp-cli/src/commands/canister/create.rs
rikonor: I agree, although I think a few options should be discussed:

1. Vendoring relevant canister interfaces in the codebase.
2. Using external crates for this.
3. Doing nothing and keeping it local to the point of use.

I think the sane default would be (3), and perhaps as a second step, you could simply put it under a module at the bottom of the file. To keep things namespaced neatly, I'd even consider this:

```rs
fn exec() {
    ...
    // example usage
    let args = ledger::cycles::InitArgs {};
    ...
}

mod ledger {
    pub mod cycles {
        pub struct InitArgs {}
    }

    // assuming we move this module to its own file, allows us to add additional submodules
    pub mod icp {}
}
```

crates/icp-cli/src/commands/canister/create.rs
viviveevee: I like the idea of a constant, but I'd rather put it in `create.rs` since it's very specific to canister creation

crates/icp-cli/src/commands/canister/create.rs
```diff
@@ -214,97 +292,69 @@ pub async fn exec(ctx: &Context, cmd: Cmd) -> Result<(), CommandError> {
...
+                        .or(c.settings.reserved_cycles_limit)
+                        .map(Nat::from),
+                    memory_allocation: cmd
+                        .settings
+                        .memory_allocation
+                        .or(c.settings.memory_allocation)
+                        .map(Nat::from),
+                    compute_allocation: cmd
                         .settings
-                        .environment_variables
```
rikonor, line 232: What will happen to environment variables? Would you perhaps be willing to give (in the PR description) a brief description of the changes? E.g, I understand the gist of it is that we are going to no longer utilize the management canister for creating canisters, but now I am confused as to which canister we're going to use. I assumed CMC stood for cycles-minting-canister, but is that the same thing as the cycles ledger?

Perhaps it would even make sense to provide some comments or docstrings at the top of the file explaining the procedure in a bit more depth.

crates/icp-cli/src/commands/canister/create.rs
```diff
@@ -84,6 +55,117 @@ pub struct Cmd {
...
+        refund_block: Option<Nat>,
+        fee_block: Option<Nat>,
+    },
+    TooOld,
+    InsufficientFunds {
+        balance: Nat,
+    },
+}
+
+fn format_create_canister_error(err: &CreateCanisterError, requested_cycles: u128) -> String {
```
rikonor: Bit of a random thought, but what do you think about doing:

```rs
impl CreateCanisterError {
    format(requested_cycles: u128) -> String {}
}
```

and then down below we can call it `err.format(cmd.cycles)`? This would be a little more concise.

crates/icp-cli/src/commands/deploy/mod.rs
```diff
@@ -40,6 +40,10 @@ pub struct Cmd {
     /// One or more controllers for the canisters being deployed. Repeat `--controller` to specify multiple.
     #[arg(long)]
     pub controller: Vec<Principal>,
+
+    /// Cycles to fund canister creation (in cycles).
+    #[arg(long, default_value_t = 2_000_000_000_000u128)]
```
rikonor: Same as the other comment - ideally, we'd make this a constant, so it doesn't have to be redefined multiple times in the code (which would make it dangerous to update).

crates/icp-cli/tests/common/clients/icp_cli.rs
```diff
@@ -1,20 +1,23 @@
 use candid::Principal;
+use icp::prelude::PathBuf;
```
rikonor: Nit, but consider replacing with `icp::prelude::*`.

crates/icp-cli/tests/common/clients/icp_cli.rs
```diff
@@ -1,20 +1,23 @@
...
 
 use crate::common::TestContext;
 
 pub struct IcpCliClient<'a> {
```
rikonor: Does it make sense to name this `clients::icp_cli::IcpCliClient`? Potentially just `clients::icp::Client` could work, no?

crates/icp-cli/tests/common/clients.rs
```diff
@@ -8,8 +9,8 @@ mod cycles_ledger;
 mod icp_cli;
 mod icp_ledger;
 
-pub fn icp(ctx: &TestContext) -> IcpCliClient<'_> {
-    IcpCliClient::new(ctx)
+pub fn icp<'a>(ctx: &'a TestContext, current_dir: &'a PathBuf) -> IcpCliClient<'a> {
```
rikonor: How come we have to introduce lifetimes?

rikonor: Couldn't you also just pass a `&Path` and then, if you want to store a PathBuf, just call `to_owned` on it? That would make you not require the lifetime, right?

crates/icp-cli/tests/common/mod.rs
```diff
@@ -19,6 +19,8 @@ pub const CYCLES_LEDGER_CID: &str = "um5iw-rqaaa-aaaaq-qaaba-cai";
 /// Governance on mainnet
 pub const GOVERNANCE_ID: &str = "rrkah-fqaaa-aaaaa-aaaaq-cai";
 
+pub const TRILLION: u128 = 1_000_000_000_000;
```
rikonor: My only concern here, is that this is under `tests`. Same for the canister IDs, etc.

crates/icp-cli/tests/common/clients/icp_cli.rs
viviveevee:
```
use clients::{icp::Client, ledger::Client};
```
becomes impossible with that. `IcpClient` is too generic of a name

crates/icp-cli/tests/common/clients.rs
viviveevee: We still need to at least mention the anonymous lifetime, but it looks better now with `impl Into<PathBuf>`. Passing `&Path` does not work because then there are still two references in the arguments and lifetime elision cannot decide which one of the two lifetimes to attach to the resulting type

crates/icp-cli/tests/common/mod.rs
viviveevee: It's now `icp::TRILLION` so we can access it everywhere. Canister IDs are now all in the `icp-canister-interfaces` crate

crates/icp-cli/src/commands/canister/create.rs
viviveevee: We're using the cycles ledger, which behind the scenes uses the CMC to create canisters using the cycles the user holds in their account. The code no longer mentions the CMC at all besides fetching the ICP/XDR conversion rate.

I explain the change of functionality in the PR description now. IMO it is not necessary to explain much in the code since canister creation is now 'just' a call to the cycles ledger. What happens behind that can stay abstracted away

rikonor: > IMO it is not necessary to explain much in the code since canister creation is now 'just' a call to the cycles ledger. What happens behind that can stay abstracted away

Creating a canister via the cycles ledger is not necessarily an obvious thing. Therefore, I do think it would be valuable to have a bit of an explanation at the top of the file explaining the mechanism by which a canister is now created, even if a minimal one.

crates/icp-cli/src/commands/canister/create.rs
```diff
@@ -1,9 +1,10 @@
 use std::collections::HashSet;
 
+use candid::{Decode, Encode, Nat};
 use clap::Parser;
 use futures::{StreamExt, stream::FuturesOrdered};
 use ic_agent::{AgentError, export::Principal};
-use ic_utils::interfaces::management_canister::{LogVisibility, builders::EnvironmentVariable};
+use icp::TRILLION;
```
rikonor: Perhaps we could put this in the prelude?

crates/icp-cli/tests/common/clients/icp_cli.rs
rikonor: You could always do:

```
use clients::{icp, ledger};

let _ = icp::Client;
let _ = ledger::Client;
```

crates/icp-cli/tests/canister_create_tests.rs
```diff
@@ -1,6 +1,6 @@
-use crate::common::TestContext;
+use crate::common::{TestContext, clients};
 use camino_tempfile::NamedUtf8TempFile as NamedTempFile;
-use icp::fs::write_string;
+use icp::{TRILLION, fs::write_string};
```
rikonor: Consider putting `TRILLLION` in the prelude.

crates/icp-cli/tests/common/mod.rs
rikonor: Sounds good, left a comment about putting `TRILLION` in the prelude.

PR #112, opened 2025-10-01
chore: revamp the various manifests and their parsing logic

rikonor:


docs/cli-reference.md
```diff
@@ -59,6 +59,7 @@ This document contains the help content for the `icp-cli` command-line program.
 
 ###### **Options:**
 
+* `--project-dir <PROJECT_DIR>` — Directory to look for icp.yaml file
```
raymondk: This is more "directory to use as your project base directory. If not specified the directory structure is traversed up until an icp.yaml file is found.

rikonor: Updated 👍

PR #113, opened 2025-10-02
feat: colocate canisters on subnets

viviveevee:
Projects that have multiple canisters usually want these canisters to live on the same subnet because it improves latency of cross-canister calls a lot. With this PR, canisters in the same project will land on the same subnet as far as this is possible.

Subnets get picked with this logic:

- If the user specifies a subnet explicitly, target it.
- If there are canisters already in the project that have been created, see which subnets those are on
  - If they are all on the same subnet, target that subnet
  - If they are on different subnets, ask the user
- If no canister is on a subnet yet, pick a random open subnet and target that

### DIFF
--- a/crates/icp-network/src/managed/run.rs
+++ b/crates/icp-network/src/managed/run.rs
@@ -326,7 +326,7 @@ pub async fn initialize_instance(
 
     let agent_url = format!("http://localhost:{}", gateway_info.port);
     eprintln!("Agent url is {}", agent_url);
-    let status = crate::status::ping_and_wait(&agent_url).await?;
+    let status = status::ping_and_wait(&agent_url).await?;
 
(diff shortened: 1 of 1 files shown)


crates/icp-network/src/managed/pocketic.rs
```diff
@@ -74,13 +74,16 @@ impl PocketIcAdminInterface {
...
+
                 // The rest of the features are disabled by default
                 ..Default::default()
             }),
 
             subnet_config_set: (SubnetConfigSet {
-                // Configure a single application subnet
-                application: 1,
+                // Configure two application subnets
+                application: 2,
```
viviveevee: Not happy about this. Does anyone have a better idea how to test the different subnet case?

PR #115, opened 2025-10-02
chore: remove boilerplate when loading projects and environments

rikonor:


crates/icp/src/manifest/mod.rs
raymondk, line 88: is "ProjectManifestLocator" a more appropriate name?

PR #117, opened 2025-10-07
feat: --version

viviveevee:
https://dfinity.atlassian.net/browse/SDK-2372

Copies the `--version` setup from dfx

### DIFF
--- a/crates/icp-cli/src/version.rs
+++ b/crates/icp-cli/src/version.rs
@@ -1,0 +1,19 @@
+use lazy_static::lazy_static;
+
+lazy_static! {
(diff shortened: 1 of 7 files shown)


crates/icp-cli/src/version.rs
```diff
@@ -0,0 +1,19 @@
...
+/// debug runs are unique.
+pub fn icp_cli_version_str() -> &'static str {
+    &VERSION_STR
+}
+
+/// Returns the git sha of the build.
+#[allow(unused)]
+pub fn git_sha() -> &'static str {
+    &GIT_SHA
+}
```
viviveevee, line 19: Currently unused, but to be displayed once debug logs are enabled

PR #124, opened 2025-10-09
chore: Add license and audit checks

raymondk:


crates/schema-gen/Cargo.toml
```diff
@@ -1,7 +1,8 @@
 [package]
 name = "schema-gen"
 version = "0.1.0"
-edition = "2021"
+edition = "2024"
+license = "Apache-2.0"
```
ilbertt: Why not inherit this parameters from the root?

```suggestion
edition.workspace = true
license.workspace = true
```

Same for all the other packages

LICENSE
ilbertt, line 1: It's technically not correct to remove the appendix, as it is part of the license itself.

.github/workflows/audit.yml
```diff
@@ -0,0 +1,36 @@
...
+jobs:
+  test:
+    name: audit:required
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      issues: write
+
+    steps:
+      - uses: actions/checkout@v4
```
ilbertt, line 35: Latest security recommendations are to pin the version of the third party actions that we use. E.g. https://github.com/dfinity/icp-cli-recipes/pull/2

Can be done in a separate PR for all the workflows together.

.github/workflows/deny.yml
```diff
@@ -0,0 +1,30 @@
...
+  # Disable DFX telemetry
+  DFX_TELEMETRY: 'off'
+
+jobs:
+  cargo-deny:
+    name: license-check:required
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: rm rust-toolchain.toml
```
ilbertt, line 27: Why this command?

LICENSE
raymondk: ... I copied it from the sdk repo ;)

crates/schema-gen/Cargo.toml
```diff
@@ -1,7 +1,8 @@
 [package]
 name = "schema-gen"
 version = "0.1.0"
-edition = "2021"
+edition.workspace = true
+license.workspace = true
```
ilbertt, line 5: Actually, there should be no need for the license to be specified in each internal crate. Or is the checker complaining?

.github/workflows/deny.yml
raymondk: @adamspofford-dfinity do you know why it is that way in the sdk repo?

LICENSE
ilbertt: Yeah, they have been updated across many repos recently... In my opinion, the most correct version is the one with the appendix, but fine with this as well

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.86.0"
+channel = "1.88.0"
```
ilbertt: We're now on [1.90.0](https://releases.rs/docs/1.90.0/). Any reason why this is behind?

crates/schema-gen/Cargo.toml
raymondk: The checker was complaining

rust-toolchain.toml
raymondk: I used the minimum version that `cargo-deny` wanted.
We can use a later version if there is something useful there.

ilbertt: I'd go with the latest version

raymondk: sure

raymondk: I bumped it and there were a bunch of clippy things that came up and my rust-analyzer did not handle all the new macros... so I reverted it.

PR #125, opened 2025-10-10
feat: display output of failed builds

viviveevee:
Currently, when a build fails, the error is not visible to the user. It is only briefly displayed in the rolling 4-line buffer. With this PR, the build step outputs are buffered, and for every build that fails, the output is displayed with a readable report.

Example (remaining) output:
```
Build output for canister my-canister:

Building: script (command: bash -c 'for i in 1 2 3 4 5; do echo "Processing step $i"; sleep 1; done') 1 of 3
Processing step 1
Processing step 2
Processing step 3
Processing step 4
Processing step 5

Building: script (command: echo "Step 2 completed") 2 of 3
Step 2 completed

Building: script (command: bash -c 'for i in 1 2 3 4 5; do echo "Compiling module $i"; done; echo "ERROR Build failed"; exit 1') 3 of 3
Compiling module 1
Compiling module 2
Compiling module 3
Compiling module 4
Compiling module 5
ERROR Build failed
Failed to build canister: command 'bash -c 'for i in 1 2 3 4 5; do echo "Compiling module $i"; done; echo "ERROR Build failed"; exit 1'' failed with status code 1
Error: One or more canisters failed to build
```

https://dfinity.atlassian.net/browse/SDK-2305

### DIFF
--- a/crates/icp-cli/tests/recipe_tests.rs
+++ b/crates/icp-cli/tests/recipe_tests.rs
@@ -1,0 +1,366 @@
+use indoc::{formatdoc, indoc};
+use k256::sha2::{Digest, Sha256};
+use predicates::{prelude::PredicateBooleanExt, str::contains};
+
+use crate::common::{TestContext, spawn_test_server};
+use icp::fs::write_string;
+
+mod common;
+
+const RECIPE_TEMPLATE: &str = indoc! {"
+    build:
+      steps:
+        - type: script
+          command: sh -c 'echo \"test\" > \"$ICP_WASM_OUTPUT_PATH\"'
+"};
+const RECIPE_TEMPLATE_CHECKSUM: &str =
+    "475752eb684a14dcddcb0884c275285b03b53d5519a51925d3e05eb01a494d68";
+
+#[test]
+fn recipe_remote_url_without_checksum() {
+    let ctx = TestContext::new();
+
+    // Setup project
+    let project_dir = ctx.create_project_dir("icp");
+
+    // Spawn HTTP server with the recipe template
+    let server = spawn_test_server("GET", "/recipe.hbs", RECIPE_TEMPLATE.as_bytes());
+    let addr = server.addr();
+
+    // Project manifest
+    let pm = formatdoc! {"
+        canister:
+          name: my-canister
+          recipe:
+            type: http://{addr}/recipe.hbs
+    "};
+
+    write_string(
+        &project_dir.join("icp.yaml"), // path
+        &pm,                           // contents
+    )
+    .expect("failed to write project manifest");
+
+    // Invoke build
+    ctx.icp()
+        .current_dir(project_dir)
+        .args(["build"])
+        .assert()
+        .success();
+}
+
+#[test]
+fn recipe_remote_url_invalid_checksum() {
+    let ctx = TestContext::new();
+
+    // Setup project
+    let project_dir = ctx.create_project_dir("icp");
+
+    // Calculate checksum
+    let actual = hex::encode({
+        let mut h = Sha256::new();
+        h.update(RECIPE_TEMPLATE.as_bytes());
+        h.finalize()
+    });
+
+    // Spawn HTTP server with the recipe template
+    let server = spawn_test_server("GET", "/recipe.hbs", RECIPE_TEMPLATE.as_bytes());
+    let addr = server.addr();
+
+    // Project manifest with invalid checksum
+    let pm = formatdoc! {"
+        canister:
+          name: my-canister
+          recipe:
+            type: http://{addr}/recipe.hbs
+            sha256: invalid
+    "};
+
+    write_string(
+        &project_dir.join("icp.yaml"), // path
+        &pm,                           // contents
+    )
+    .expect("failed to write project manifest");
+
+    // Invoke build
+    ctx.icp()
+        .current_dir(project_dir)
+        .args(["build"])
+        .assert()
+        .failure()
+        .stderr(
+            contains("checksum mismatch")
+                .and(contains("expected invalid"))
+                .and(contains(format!("actual {actual}"))),
+        );
+}
+
+#[test]
+fn recipe_remote_url_wrong_checksum() {
+    let ctx = TestContext::new();
+
+    // Setup project
+    let project_dir = ctx.create_project_dir("icp");
+
+    let actual = hex::encode({
+        let mut h = Sha256::new();
+        h.update(RECIPE_TEMPLATE.as_bytes());
+        h.finalize()
+    });
+
+    // A random wrong but valid checksum
+    let expected = "508f107eba1885bd165f4ef75c1b8cf914cf9eda6365f93214bb0cc39bd07ddd";
+
+    // Spawn HTTP server with the recipe template
+    let server = spawn_test_server("GET", "/recipe.hbs", RECIPE_TEMPLATE.as_bytes());
+    let addr = server.addr();
+
+    // Project manifest with wrong checksum
+    let pm = formatdoc! {"
+        canister:
+          name: my-canister
+          recipe:
+            type: http://{addr}/recipe.hbs
+            sha256: {expected}
+    "};
+
+    write_string(
+        &project_dir.join("icp.yaml"), // path
+        &pm,                           // contents
+    )
+    .expect("failed to write project manifest");
+
+    // Invoke build
+    ctx.icp()
+        .current_dir(project_dir)
+        .args(["build"])
+        .assert()
+        .failure()
+        .stderr(
+            contains("checksum mismatch")
+                .and(contains(format!("expected {expected}")))
+                .and(contains(format!("actual {actual}"))),
+        );
+}
+
+#[test]
+fn recipe_remote_url_valid_checksum() {
+    let ctx = TestContext::new();
+
+    // Setup project
+    let project_dir = ctx.create_project_dir("icp");
+
+    // Spawn HTTP server with the recipe template
+    let server = spawn_test_server("GET", "/recipe.hbs", RECIPE_TEMPLATE.as_bytes());
+    let addr = server.addr();
+
+    // Project manifest with valid checksum
+    let pm = formatdoc! {"
+        canister:
+          name: my-canister
+          recipe:
+            type: http://{addr}/recipe.hbs
+            sha256: {RECIPE_TEMPLATE_CHECKSUM}
+    "};
+
+    write_string(
+        &project_dir.join("icp.yaml"), // path
+        &pm,                           // contents
+    )
+    .expect("failed to write project manifest");
+
+    // Invoke build
+    ctx.icp()
+        .current_dir(project_dir)
+        .args(["build"])
+        .assert()
+        .success();
+}
+
+#[test]
+fn recipe_local_file_without_checksum() {
+    let ctx = TestContext::new();
+
+    // Setup project
+    let project_dir = ctx.create_project_dir("icp");
+
+    write_string(
+        &project_dir.join("recipe.hbs"), // path
+        RECIPE_TEMPLATE,                 // contents
+    )
+    .expect("failed to write recipe template");
+
+    // Project manifest without checksum
+    let pm = formatdoc! {"
+        canister:
+          name: my-canister
+          recipe:
+            type: file://./recipe.hbs
+    "};
+
+    write_string(
+        &project_dir.join("icp.yaml"), // path
+        &pm,                           // contents
+    )
+    .expect("failed to write project manifest");
+
+    // Invoke build
+    ctx.icp()
+        .current_dir(project_dir)
+        .args(["build"])
+        .assert()
+        .success();
+}
+
+#[test]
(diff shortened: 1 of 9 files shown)


crates/icp-cli/src/commands/build/mod.rs
```diff
@@ -158,3 +173,24 @@ pub async fn exec(ctx: &Context, cmd: Cmd) -> Result<(), CommandError> {
...
+    output: RollingLines,
+}
+
+fn dump_build_output(canister_name: &str, step_outputs: Vec<StepOutput>) {
+    let crop_message = if step_outputs.len() == MAX_LINES_PER_STEP {
+        format!(" (last {MAX_LINES_PER_STEP} lines)")
+    } else {
+        String::new()
+    };
+    println!("Build output for canister {canister_name}{crop_message}:");
```
raymondk: I think we might want those to go through ctx.term

crates/icp-cli/tests/build_tests.rs
```diff
@@ -79,3 +80,97 @@ fn build_adapter_script_multiple() {
...
+                - type: script
+                  command: sh -c 'for i in $(seq 1 20); do echo "failing build step $i"; done; exit 1'
+          - name: unimportant-canister
+            build:
+              steps:
+                - type: script
+                  command: echo "hide this" 
+                - type: script
+                  command: sh -c 'touch "$ICP_WASM_OUTPUT_PATH"'
+        "#;
```
raymondk: We're using indoc to nicely format all the files.

```suggestion
    let pm = indoc! {r#"
        canisters:
          - name: my-canister
            build:
              steps:
                - type: script
                  command: echo "success 1"
                - type: script
                  command: echo "success 2"
                - type: script
                  command: sh -c 'for i in $(seq 1 20); do echo "failing build step $i"; done; exit 1'
          - name: unimportant-canister
            build:
              steps:
                - type: script
                  command: echo "hide this" 
                - type: script
                  command: sh -c 'touch "$ICP_WASM_OUTPUT_PATH"'
      "#}
```

crates/icp-cli/tests/build_tests.rs
```diff
@@ -79,3 +80,97 @@ fn build_adapter_script_multiple() {
...
+}
+
+#[test]
+fn build_adapter_display_failing_prebuilt_output() {
+    let ctx = TestContext::new();
+
+    // Setup project
+    let project_dir = ctx.create_project_dir("icp");
+
+    // Project manifest with a prebuilt step that will fail (non-existent file)
```
raymondk, line 153:
```suggestion
    // Project manifest with a prebuilt step that will fail (non-existent file)
    let pm = indoc! {r#"
        canisters:
          - name: my-canister
            build:
              steps:
                - type: script
                  command: echo "initial step succeeded"
                - type: pre-built
                  path: /nonexistent/path/to/wasm.wasm
                  sha256: 0000000000000000000000000000000000000000000000000000000000000000
      "#};
```

crates/icp-cli/src/progress.rs
```diff
@@ -40,14 +43,9 @@ pub struct RollingLines {
 }
 
 impl RollingLines {
-    /// Create a new buffer with a fixed capacity, pre-filled with empty strings.
+    /// Create a new buffer with a fixed capacity.
     pub fn new(capacity: usize) -> Self {
-        let mut buf = VecDeque::with_capacity(capacity);
-
-        for _ in 0..capacity {
```
raymondk, line 47: if we don't fill it with empty lines, won't the output jump around too much as the lines fill up?

viviveevee: We previously filtered out empty lines for displaying, I just thought it was pointless to do it that way. I'll test it manually

crates/icp-cli/src/commands/build/mod.rs
```diff
@@ -93,34 +96,46 @@ pub async fn exec(ctx: &Context, cmd: Cmd) -> Result<(), CommandError> {
...
                 for (i, step) in c.build.steps.iter().enumerate() {
                     // Indicate to user the current step being executed
                     let current_step = i + 1;
-                    let pb_hdr = format!("Building: {step} {current_step} of {step_count}");
+                    let pb_hdr = format!("\nBuilding: {step} {current_step} of {step_count}");
 
                     let script_handler = ScriptProgressHandler::new(pb.clone(), pb_hdr.clone());
 
                     // Setup script progress handling and receiver join handle
                     let (tx, rx) = script_handler.setup_output_handler();
```
rikonor: Why do we have this `script_handler` and `setup_output_handler` if you are now handling output in the command itself? Can the script handler not be made to support outputting failure information as well? The point was to remove the burden of handling these details from the command (I believe this applies to `build` and `sync` both).

crates/icp-cli/src/commands/build/mod.rs
```diff
@@ -93,34 +96,46 @@ pub async fn exec(ctx: &Context, cmd: Cmd) -> Result<(), CommandError> {
...
+                    let build_result = ctx
+                        .builder
                         .build(
                             step, // step
                             &Params {
                                 path: canister_path.to_owned(),
                                 output: wasm_output_path.to_owned(),
                             },
-                            Some(tx),
+                            tx,
```
rikonor: Why not `Option`? This means that tests using `Builder` must mock a `tx` instead of just being able to pass `None`.

crates/icp-cli/src/progress.rs
```diff
@@ -165,7 +163,7 @@ impl ScriptProgressHandler {
 
     /// Create a channel and start handling script output for progress updates
     /// Returns the sender and a join handle for the background receiver task.
-    pub fn setup_output_handler(&self) -> (mpsc::Sender<String>, JoinHandle<()>) {
+    pub fn setup_output_handler(&self) -> (mpsc::Sender<String>, JoinHandle<RollingLines>) {
```
rikonor: Why return a `RollingLines`, which is meant to be an internal detail of this `ScriptHandler`? Can't you just return the error output?

crates/icp-cli/tests/build_tests.rs
```diff
@@ -79,3 +80,97 @@ fn build_adapter_script_multiple() {
...
+        .failure()
+        .stdout(contains("Build output for canister my-canister"))
+        .stdout(contains("Building: script (command: echo \"success 1\") 1 of 3"))
+        .stdout(contains("success 1"))
+        .stdout(contains("Building: script (command: echo \"success 2\") 2 of 3"))
+        .stdout(contains("success 2"))
+        .stdout(contains("Building: script (command: sh -c 'for i in $(seq 1 20); do echo \"failing build step $i\"; done; exit 1') 3 of 3"))
+        .stdout(contains("failing build step 1"))
+        .stdout(contains("failing build step 20"))
+        .stdout(contains("hide this").not());
```
rikonor, line 143: Maybe also use `indoc` here to do a `contains(...)` check against one nicely formatted message?

crates/icp-cli/tests/build_tests.rs
```diff
@@ -79,3 +80,97 @@ fn build_adapter_script_multiple() {
...
+    let pm = r#"
+        canisters:
+          - name: my-canister
+            build:
+              steps:
+                - type: script
+                  command: echo "initial step succeeded"
+                - type: pre-built
+                  path: /nonexistent/path/to/wasm.wasm
+                  sha256: 0000000000000000000000000000000000000000000000000000000000000000
```
rikonor: Do you actually need such a sha256 value? Can't a simpler test value suffice?

crates/icp/src/canister/build.rs
```diff
@@ -66,6 +66,12 @@ pub enum BuildError {
 
     #[error(transparent)]
     Unexpected(#[from] anyhow::Error),
+
+    #[error("failed to send build output")]
+    SendOutput(#[from] SendError<String>),
+
+    #[error("failed to join futures")]
+    JoinError(#[from] tokio::task::JoinError),
```
rikonor, line 71: Maybe keep the `Unexpected` error last in line, since it's a sort of catch-all.

crates/icp/src/canister/prebuilt.rs
```diff
@@ -22,17 +22,18 @@ impl Build for Prebuilt {
...
+        let Step::Prebuilt(adapter) = step else {
+            panic!("expected prebuilt adapter");
         };
 
         let wasm = match &adapter.source {
             // Local path
             SourceField::Local(s) => {
+                stdio
+                    .send(format!("Reading local file: {}", s.path))
+                    .await?;
```
rikonor: This will never actually be displayed to the user, right? Because it will (likely) finish almost instantly.

crates/icp/src/canister/prebuilt.rs
```diff
@@ -73,6 +75,7 @@ impl Build for Prebuilt {
 
         // Verify the checksum if it's provided
         if let Some(expected) = &adapter.sha256 {
+            stdio.send("Verifying checksum".to_string()).await?;
```
rikonor: Same as above - will it actually get shown, if it completes almost instantly?

crates/icp/src/canister/script.rs
```diff
@@ -108,7 +110,7 @@ impl Build for Script {
             // Spawn command and handle stdio
             // We need to join! as opposed to try_join! even if we only care about the result of the task
             // because we want to make sure we finish  reading all of the output
-            let (_, _, status) = join!(
+            let (stdout_result, stderr_result, status) = join!(
```
rikonor: Consider renaming to `out` and `err`.

crates/icp/src/canister/script.rs
```diff
@@ -326,14 +329,15 @@ mod tests {
...
+        let (tx, _rx) = mpsc::channel::<String>(100);
         Script
             .build(
                 &build::Step::Script(v),
                 &build::Params {
                     path: "/".into(),
                     output: "/".into(),
                 },
-                None,
+                tx,
```
rikonor: This is what I was referring to with regards to making `tx` not an `Option` - it serves no purpose for the test here, but adds to the required boilerplate. Why not just keep it `Option`?

crates/icp-cli/src/commands/build/mod.rs
```diff
@@ -177,7 +177,7 @@ pub async fn exec(ctx: &Context, cmd: Cmd) -> Result<(), CommandError> {
 #[derive(Debug)]
 struct StepOutput {
     title: String,
-    output: RollingLines,
+    output: Vec<String>,
```
rikonor: Should this be just `String`?

crates/icp-cli/src/progress.rs
viviveevee: Tested manually. It looks fine

crates/icp-cli/src/commands/build/mod.rs
```diff
@@ -77,83 +77,106 @@ pub async fn exec(ctx: &Context, cmd: Cmd) -> Result<(), CommandError> {
...
+                    |err| format!("Failed to build canister: {err}"),
+                )
+                .await;
+
+                // After progress bar is finished, dump the output if build failed
+                if let Err(e) = &result {
+                    pb.dump_output(ctx);
+                    let _ = ctx
+                        .term
+                        .write_line(&format!("Failed to build canister: {e}"));
```
viviveevee, line 157: My gut feeling says there's still some way for the output to get messed up because of progress bars, but I don't see anything specific.

crates/icp/src/canister/prebuilt.rs
viviveevee: It will not be visible if everything works out fine, but if there is an error it will show up in the full error output

crates/icp-cli/src/commands/build/mod.rs
viviveevee: I reworked output display significantly. It's now all handled by the progress bar

crates/icp-cli/src/progress.rs
```diff
@@ -151,6 +164,135 @@ impl ProgressManager {
...
+                let _ = ctx.term.write_line(&format!("{}", line));
+            }
+            if step_output.output.len() == 0 {
+                let _ = ctx.term.write_line("<no output>");
+            }
+        }
+    }
+}
+
+pub enum SomeProgressBar {
```
rikonor: Why not rename:
```
ProgressBar -> SimpleProgressBar or BasicProgressBar
SomeProgressBar -> ProgressBar
```
`SomeProgressBar` is quite unclear - it could be construed as `type Option<ProgressBar>` unless one checks the enum definition.

crates/icp-cli/tests/build_tests.rs
```diff
@@ -159,20 +170,130 @@ fn build_adapter_display_failing_prebuilt_output() {
...
+        "#};
+
+    write_string(
+        &project_dir.join("icp.yaml"), // path
+        pm,                            // contents
+    )
+    .expect("failed to write project manifest");
+
+    // Invoke build
+    let expected_output = indoc! {r#"
```
rikonor, line 218: This is a really nice pattern 👍

PR #132, opened 2025-10-10
feat: recipe hash and verification

ilbertt:


crates/icp-cli/tests/recipe_tests.rs
```diff
@@ -0,0 +1,247 @@
...
+    let server = spawn_test_server("GET", "/recipe.hbs", recipe_template.as_bytes());
+    let addr = server.addr();
+
+    // Project manifest with invalid checksum
+    let pm = formatdoc! {r#"
+        canister:
+          name: my-canister
+          recipe:
+            type: http://{addr}/recipe.hbs
+            sha256: invalid
```
adamspofford, line 83: This is a failed-to-parse error, yes? (Or would be if we weren't doing hex string comparison.) We should have a test that also checks that a validly written but incorrect checksum fails.

ilbertt: Added

rikonor: The `sha256` is not parsed; the downloaded file is hashed, and the resulting value is compared against the `sha256` value.

crates/icp-cli/tests/recipe_tests.rs
```diff
@@ -0,0 +1,284 @@
...
+            sha256: invalid_checksum_should_be_ignored
+    "#};
+
+    write_string(
+        &project_dir.join("icp.yaml"), // path
+        &pm,                           // contents
+    )
+    .expect("failed to write project manifest");
+
+    // Invoke build - should succeed because local files don't verify checksums
```
raymondk, line 246: Why wouldn't we verify the checksum for a local file?

ilbertt: We could, I just thought since you have it locally, you could verify its integrity manually. Let me add it in another PR

ilbertt: Here we go: https://github.com/dfinity/icp-cli/pull/133

PR #133, opened 2025-10-10
feat: allow hash verification for local recipes

ilbertt:
Follow-up of #132

PR #136, opened 2025-10-13
feat: Remove builtin recipes

raymondk:
This change is to remove the built in recipes so that we only rely on the recipe repository and there are minimal helpers in the icp cli.

For now, there are example recipes in the examples/ directory - they will need to be moved into https://github.com/dfinity/icp-cli-recipes.

The changes adds a hidden `icp project show` command that prints out the manifest it could be used to test out that the recipes are getting computed properly.

This will end up using: https://github.com/dfinity/icp-cli-recipes/pull/3

### DIFF
--- a/crates/icp-cli/src/main.rs
+++ b/crates/icp-cli/src/main.rs
@@ -7,13 +7,8 @@ use console::Term;
 use icp::{
     Directories, agent,
     canister::{
-        self,
-        assets::Assets,
-        build::Builder,
-        prebuilt::Prebuilt,
-        recipe::handlebars::Handlebars,
-        script::Script,
-        sync::Syncer,
+        self, assets::Assets, build::Builder, prebuilt::Prebuilt, recipe::handlebars::Handlebars,
+        script::Script, sync::Syncer,
     },
     identity, manifest, network,
     prelude::*,


crates/icp/src/network/mod.rs
```diff
@@ -22,7 +22,9 @@ mod lock;
 pub mod managed;
 pub mod structure;
 
-#[derive(Clone, Debug, PartialEq, JsonSchema)]
+pub const DEFAULT_IC_GATEWAY: &str = "https://icp0.io";
```
ilbertt: This was removed in #135. Merge conflict?

crates/icp/src/manifest/recipe.rs
ilbertt, line 1: Can you update the doc comment as pointed out in https://github.com/dfinity/icp-cli/pull/133#discussion_r2421995325?

crates/icp/src/canister/recipe/handlebars.rs
```diff
@@ -290,178 +269,6 @@ impl HelperDef for ReplaceHelper {
...
-pub const WASM_INJECT_METADATA_PARTIAL: &str = r#"
-- type: script
-  commands:
-    - command -v ic-wasm >/dev/null 2>&1 || { echo >&2 "ic-wasm not found. To install ic-wasm, see https://github.com/dfinity/ic-wasm \n"; exit 1; }
-    - ic-wasm "$ICP_WASM_OUTPUT_PATH" -o "${ICP_WASM_OUTPUT_PATH}" metadata "{{ name }}" -d "{{ value }}" --keep-name-section
-"#;
-
-/// Collection of reusable Handlebars partials for WASM processing
-/// These partials can be included in templates using {{> partial-name}} syntax
-pub const PARTIALS: [(&str, &str); 4] = [
```
rikonor, line 330: Why remove the partials? These can be used by any recipe, regardless of whether it's built-in or is external.

examples/icp-static-assets-recipe/recipe.hb.yaml
```diff
@@ -0,0 +1,32 @@
...
+{{! `version: string` Optional, The version of the asset canister to use. Defaults to the master branch}}
+{{! `dir: string` Required, the directory of assets to synchronize }}
+{{! `metadata: [name: string, value: string]`: An array of name/value pairs that get injected into the wasm metadata section }}
+
+build:
+
+  steps:
+
+    - type: pre-built
+      {{! There is no convenient way to define defaults }}
```
rikonor, line 11: Maybe using a helper? See https://github.com/dfinity/icp-cli/blob/main/crates/icp/src/canister/recipe/handlebars.rs#L267-L291

crates/icp/src/canister/recipe/handlebars.rs
raymondk: I prefer not to use helpers and partials at first.
It creates a dependency on icp-cli and it makes the sha of your recipe meaningless because how it executes depends on the version of the cli you are using.

Same thing with helpers. I would rather avoid helpers for now - we can consider adding some later if we find it necessary.

examples/icp-static-assets-recipe/recipe.hb.yaml
raymondk: I would prefer not to add more helpers at the moment.

PR #139, opened 2025-10-14
feat: debug logs

ilbertt:
Tested locally, this is what the output looks like when the `--debug` flag is enabled:

```shell
DEBUG icp: Starting icp-cli version="0+rev340282366920938463463374607431768211455.dirty-62d5900" git_sha="62d590076dd42998e4f0d5f9e47fe704451fd5d9" command=Deploy(Cmd { names: [], mode: "auto", subnet_id: None, controller: [], cycles: 2000000000000, identity: IdentityOpt { identity: None }, environment: EnvironmentOpt { environment: None, ic: false } })
INFO icp::progress: Fetching remote file: ...
...
INFO icp: Command executed successfully
```

Uses `tracing` for logging, and removes the `console` dependecy.

### DIFF
--- a/crates/icp-cli/tests/canister_start_tests.rs
+++ b/crates/icp-cli/tests/canister_start_tests.rs
@@ -82,7 +82,7 @@ fn canister_start() {
         ])
         .assert()
         .success()
-        .stderr(
+        .stdout(
             starts_with("Canister Status Report:")
                 .and(contains("Status: Stopped"))
                 .and(contains("Controllers: 2vxsx-fae")),


crates/icp-cli/src/commands/mod.rs
```diff
@@ -60,6 +60,13 @@ pub struct Context {
     pub syncer: Arc<dyn Synchronize>,
 }
 
+impl Context {
+    /// Print a message to the user.
```
viviveevee:
```suggestion
    /// Print a message to the user on stdout.
```

crates/icp-cli/src/progress.rs
```diff
@@ -213,7 +218,10 @@ impl MultiStepProgressBar {
             let mut complete = RollingLines::new(MAX_LINES_PER_STEP); // We need _some_ limit to prevent consuming infinite memory
 
             while let Some(line) = rx.recv().await {
-                debug!(line);
+                // Only log non-empty lines to avoid cluttering debug output
+                if !line.trim().is_empty() {
+                    debug!("{}", line);
+                }
```
viviveevee: Does this forbid the user from adding empty lines to space out their build logs? I don't think fussing around with the user's output is a good idea

PR #141, opened 2025-10-15
chore: update recipe examples to point to the dfinity repo

raymondk:


crates/icp/src/manifest/recipe.rs
```diff
@@ -1,18 +1,79 @@
...
+            let (registry, recipe) = v.split_once("/").expect("delimiter missing");
+
+            return Self::Registry {
+                name: registry.to_owned(),
+                recipe: recipe.to_owned(),
+                version: version.to_owned(),
+            };
+        }
+
+        panic!("Unable to convert {value} into a valid recipe type.")
```
vincent-dfinity: Should we handle the error instead of panic?

crates/icp/src/manifest/recipe.rs
```diff
@@ -1,18 +1,79 @@
...
+        let s: String = self.clone().into();
+        write!(f, "{s}")
+    }
 }
 
 impl From<String> for RecipeType {
     fn from(value: String) -> Self {
-        let other = value.as_str();
-        Self::Unknown(other.to_owned())
+        let v = value.as_str();
```
rikonor: FYI, you can also do `impl From<&str>` instead.

crates/icp/src/manifest/recipe.rs
rikonor: Yeah, it might make sense to have this be a `TryFrom` instead.

examples/icp-motoko-recipe/icp.yaml
```diff
@@ -1,6 +1,6 @@
 canister:
   name: my-canister
   recipe:
-    type: file://recipe.hb.yaml
+    type: "@dfinity/motoko"
```
rikonor, line 4: Maybe we could have an example for using a recipe from a local file, similar to how there's an example for a recipe from a remote url.

crates/icp/src/manifest/recipe.rs
```diff
@@ -1,18 +1,84 @@
...
 
 use schemars::JsonSchema;
 use serde::Deserialize;
 
-#[derive(Clone, Debug, Deserialize, PartialEq, JsonSchema)]
+#[derive(Clone, Debug, PartialEq, JsonSchema)]
 #[serde(rename_all = "lowercase", from = "String")]
 pub enum RecipeType {
-    Unknown(String),
+    File(String),
```
rikonor, line 12: If you're splitting with newlines in-between, might as well ad some docstrings (`///`)

examples/icp-motoko-recipe/icp.yaml
raymondk: I'll add that with: https://github.com/dfinity/icp-cli/pull/142

PR #143, opened 2025-10-15
chore: refactor of command invocation handling

rikonor:
This PR makes several standardizing changes:

1. Removes the various `dispatch` functions in favor of a centralized setup in `main`.
2. Removes various `Command` structs that were actually unused.
3. Standardizes on `Subcommand` enums being called `Command`.
4. Standardizes on `exec` functions taking reference to `args`, instead of a `cmd`, and renames all `Cmd` structs to `<Command>Args`.

### DIFF
--- a/crates/icp-cli/src/commands/canister/mod.rs
+++ b/crates/icp-cli/src/commands/canister/mod.rs
@@ -22,12 +22,13 @@ pub enum Command {
     Info(info::InfoArgs),
     Install(install::InstallArgs),
     List(list::ListArgs),
+
+    #[command(subcommand)]
+    Settings(settings::Command),
+
     Show(show::ShowArgs),
     Start(start::StartArgs),
     Status(status::StatusArgs),
     Stop(stop::StopArgs),
     TopUp(top_up::TopUpArgs),
-
-    #[command(subcommand)]
-    Settings(settings::Command),
 }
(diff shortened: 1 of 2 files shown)


docs/cli-reference.md
```diff
@@ -12,15 +12,15 @@ This document contains the help content for the `icp-cli` command-line program.
...
-* [`icp-cli canister settings`↴](#icp-cli-canister-settings)
-* [`icp-cli canister settings show`↴](#icp-cli-canister-settings-show)
-* [`icp-cli canister settings update`↴](#icp-cli-canister-settings-update)
 * [`icp-cli canister list`↴](#icp-cli-canister-list)
+* [`icp-cli canister show`↴](#icp-cli-canister-show)
 * [`icp-cli canister start`↴](#icp-cli-canister-start)
 * [`icp-cli canister status`↴](#icp-cli-canister-status)
 * [`icp-cli canister stop`↴](#icp-cli-canister-stop)
 * [`icp-cli canister top-up`↴](#icp-cli-canister-top-up)
+* [`icp-cli canister settings`↴](#icp-cli-canister-settings)
```
raymondk: gg why is this stuff not coming up in alphabetical order now?

crates/icp-cli/src/commands/canister/mod.rs
```diff
@@ -1,97 +1,33 @@
...
-    #[snafu(transparent)]
-    Status { source: status::CommandError },
-
-    #[snafu(transparent)]
-    Stop { source: stop::CommandError },
-
-    #[snafu(transparent)]
-    TopUp { source: top_up::CommandError },
+    #[command(subcommand)]
+    Settings(settings::Command),
```
raymondk: maybe it's because these are not sorted?

docs/cli-reference.md
rikonor: Hmm, I reverted it, it should be back to same as before.

crates/icp-cli/src/commands/canister/mod.rs
rikonor: Reverted the ordering change.

PR #148, opened 2025-10-17
feat: operations, argument validation, and args for global mode

rikonor:
This change introduces a few new things:
1. Begins the separation of commands and operations. Commands act as an entrypoint for the cli, and in turn make use of lower-level operations that don't require knowledge of `icp` projects. This allows us to:
2. Introduce global and project mode - this means that one should be able to run the cli tool in both project context, as well as outside of a project. This enables one to use operations that should not necessarily be limited to a project's scope.
3. Introduce a crappy little framework for cli command validation, since more extensive validation is needed (e.g some things are allowed in global mode, but not project mode, and vice-versa).

The above are done for several initial commands, with the assumption that the same work can take place for the rest of the cli footprint (currently global mode and arg validation is implemented for `icp canister start/stop` and `token transfer/balance`).

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -122,8 +122,8 @@ Perform canister operations against a network
 ###### **Options:**
 
 * `--identity <IDENTITY>` — The user identity to run this command as
-* `--environment <ENVIRONMENT>` — Override the environment to connect to. By default, the local environment is used
-* `--ic` — Shorthand for --environment=ic
+* `--network <NETWORK>`
+* `--environment <ENVIRONMENT>`
 
 
 
@@ -138,8 +138,7 @@ Perform canister operations against a network
 ###### **Options:**
 
 * `--identity <IDENTITY>` — The user identity to run this command as
-* `--environment <ENVIRONMENT>` — Override the environment to connect to. By default, the local environment is used
-* `--ic` — Shorthand for --environment=ic
+* `--environment <ENVIRONMENT>`
 * `--controller <CONTROLLER>` — One or more controllers for the canister. Repeat `--controller` to specify multiple
 * `--compute-allocation <COMPUTE_ALLOCATION>` — Optional compute allocation (0 to 100). Represents guaranteed compute capacity
 * `--memory-allocation <MEMORY_ALLOCATION>` — Optional memory allocation in bytes. If unset, memory is allocated dynamically
@@ -164,8 +163,8 @@ Perform canister operations against a network
 ###### **Options:**
 
 * `--identity <IDENTITY>` — The user identity to run this command as
-* `--environment <ENVIRONMENT>` — Override the environment to connect to. By default, the local environment is used
-* `--ic` — Shorthand for --environment=ic
+* `--network <NETWORK>`
+* `--environment <ENVIRONMENT>`
 
 
 
@@ -180,8 +179,8 @@ Perform canister operations against a network
 ###### **Options:**
 
 * `--identity <IDENTITY>` — The user identity to run this command as
-* `--environment <ENVIRONMENT>` — Override the environment to connect to. By default, the local environment is used
-* `--ic` — Shorthand for --environment=ic
+* `--network <NETWORK>`
+* `--environment <ENVIRONMENT>`
 
 
 
@@ -202,8 +201,7 @@ Perform canister operations against a network
   Possible values: `auto`, `install`, `reinstall`, `upgrade`
 
 * `--identity <IDENTITY>` — The user identity to run this command as
-* `--environment <ENVIRONMENT>` — Override the environment to connect to. By default, the local environment is used
-* `--ic` — Shorthand for --environment=ic
+* `--environment <ENVIRONMENT>`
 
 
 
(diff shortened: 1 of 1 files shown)


crates/icp-cli/src/commands/canister/start.rs
```diff
@@ -48,27 +80,53 @@ pub(crate) enum CommandError {
...
+
+            (agent, cid.to_owned())
         }
 
-        Mode::Project(_) => {
+        Mode::Project(pdir) => {
+            // Argument (Canister)
+            let args::Canister::Name(name) = &args.canister else {
+                return Err(CommandError::Args);
+            };
```
viviveevee, line 114: I don't think we should forbid using principals in projects. Otherwise to talk to a specific canister on a non-default network one has to somehow hook up the principal to a name just for one-off calls

crates/icp-cli/src/commands/canister/start.rs
```diff
@@ -81,27 +139,28 @@ pub(crate) async fn exec(ctx: &Context, args: &StartArgs) -> Result<(), CommandE
             }
 
             // Ensure canister is included in the environment
-            if !env.canisters.contains_key(&args.name) {
+            if !env.canisters.contains_key(name) {
```
viviveevee, line 142: Similarly, this should probably only happen if a non-principal was supplied

crates/icp-cli/src/commands/canister/settings/show.rs
```diff
@@ -4,25 +4,30 @@ use ic_management_canister_types::{CanisterStatusResult, LogVisibility};
...
+    pub(crate) network: Option<args::Network>,
+
+    #[arg(long)]
+    pub(crate) environment: Option<args::Environment>,
 }
 
 #[derive(Debug, thiserror::Error)]
 pub(crate) enum CommandError {
+    #[error("an invalid argument was provided")]
+    Args,
```
viviveevee, line 29: I don't think this is enough detail to give a decent user experience. It's ok to leave this as-is, but at least make a ticket to improve the situation. We can easily say 'only provide one of network and environment, not both' here

crates/icp-cli/src/commands/validation.rs
```diff
@@ -0,0 +1,245 @@
...
+    pub(crate) fn all_modes() -> Vec<Mode> {
+        vec![Mode::Global, Mode::Project("dir".into())]
+    }
+
+    // pub(crate) fn all_networks() -> Vec<args::Network> {
+    //     vec![
+    //         args::Network::Name("my-network".to_string()),
+    //         args::Network::Url("http::/www.example.com".to_string()),
+    //     ]
+    // }
```
viviveevee, line 44:
```suggestion
    pub(crate) fn all_modes() -> Vec<Mode> {
        vec![Mode::Global, Mode::Project("dir".into())]
    }
    
```

crates/icp-cli/src/commands/validation.rs
```diff
@@ -0,0 +1,245 @@
...
+    canister: impl Into<(&'a Canister,)>,
+    m: &Mode,
+) -> Option<&'static str> {
+    let (canister,) = canister.into();
+    (matches!(m, Mode::Global) && !matches!(canister, Canister::Principal(_)))
+        .then_some(PLEASE_PROVIDE_A_CANISTER_PRINCIPAL_IN_GLOBAL_MODE)
+}
+
+const PLEASE_PROVIDE_A_CANISTER_PRINCIPAL_IN_GLOBAL_MODE: &str = r#"
+    Please provide a canister principal in global mode.
```
viviveevee, line 57: Not convinced this is the best wording. `global mode` makes me think of a flag that I could use to switch to local mode.
```suggestion
    {user_provided_string} is not a principal. Cannot use non-principals outside of a project.
```

crates/icp-cli/src/commands/validation.rs
```diff
@@ -0,0 +1,245 @@
...
+    environment: impl Into<(&'a Option<Environment>,)>,
+    m: &Mode,
+) -> Option<&'static str> {
+    let (environment,) = environment.into();
+    (matches!(m, Mode::Global) && environment.is_some())
+        .then_some(ENVIRONMENTS_ARE_NOT_AVAILABLE_IN_GLOBAL_MODE)
+}
+
+const ENVIRONMENTS_ARE_NOT_AVAILABLE_IN_GLOBAL_MODE: &str = r#"
+    Environments are not available in global mode.
```
viviveevee, line 83:
```suggestion
    Environments are not available in outside of projects.
```

crates/icp-cli/src/commands/validation.rs
```diff
@@ -0,0 +1,245 @@
...
+    network: impl Into<(&'a Option<Network>,)>,
+    m: &Mode,
+) -> Option<&'static str> {
+    let (network,) = network.into();
+    (matches!(m, Mode::Global) && !matches!(network, Some(Network::Url(_))))
+        .then_some(A_NETWORK_URL_IS_REQUIRED_IN_GLOBAL_MODE)
+}
+
+const A_NETWORK_URL_IS_REQUIRED_IN_GLOBAL_MODE: &str = r#"
+    A network `url` is required in global mode.
```
viviveevee, line 96:
```suggestion
    A network `url` is required outside of a project.
```

crates/icp-cli/src/commands/canister/start.rs
viviveevee: Alternate suggestion: https://github.com/dfinity/icp-cli/pull/157

PR #153, opened 2025-10-19
fix: do not interleave build errors with progress bars

ilbertt:


crates/icp-cli/src/commands/build/mod.rs
```diff
@@ -157,34 +156,45 @@ pub async fn exec(ctx: &Context, args: &BuildArgs) -> Result<(), CommandError> {
...
+                {
+                    failed_outputs.push((output, e));
                 }
             }
 
-            if found_error {
+            // If any builds failed, dump the output and abort
+            if !failed_outputs.is_empty() {
+                for (output, e) in failed_outputs {
+                    for line in output {
```
viviveevee, line 191:
```suggestion
            while let Some((res, output)) = futs.next().await {
                if let Err(e) = res
                    && let Some(output) = output
                {
                    failed_outputs.push((e, output));
}
}

            // If any builds failed, dump the output and abort
            if !failed_outputs.is_empty() {
                for (e, output) in failed_outputs {
                    for line in output {
```
swapping the order looks like a future footgun

ilbertt: Done

PR #165, opened 2025-10-24
chore: poc another context rework

viviveevee:


crates/icp-cli/src/commands/args.rs
```diff
@@ -0,0 +1,350 @@
...
+    #[error("Environment '{environment}' not found.")]
+    EnvironmentNotFound { environment: String },
+
+    #[error("Failed to load project: {0}")]
+    LoadProject(#[from] icp::LoadError),
+}
+
+impl ArgContext {
+    pub(crate) async fn new(
+        ctx: &Context,
```
viviveevee, line 38: Better to not take ctx here

PR #167, opened 2025-10-24
feat: General file locking mechanism

adamspofford-dfinity:
This PR implements locking for the build artifact store. The two other subsystems needing locking are the network data store and the identity store; they will go in separate PRs since both require code restructuring.

### DIFF
--- a/crates/icp/src/fs/lock.rs
+++ b/crates/icp/src/fs/lock.rs
@@ -138,6 +138,7 @@ pub enum LockError {
 }
 
 /// File lock guard. Do not use as a temporary in an expression - if you are making a temporary lock, use `with_*`.
+#[clippy::has_significant_drop]
 pub struct DirectoryStructureGuardOwned<T> {
     paths_access: T,
     guard: File,


crates/icp/src/fs/lock.rs
```diff
@@ -0,0 +1,196 @@
...
+    fn lock_file(&self) -> PathBuf;
+}
+
+impl<T: PathsAccess> DirectoryStructureLock<T> {
+    /// Creates a new lock, implicitly calling [`fs::create_dir_all`] on the parent.
+    pub fn open_or_create(paths_access: T) -> Result<Self, LockError> {
+        let lock_path = paths_access.lock_file();
+        fs::create_dir_all(lock_path.parent().unwrap())?;
+        let lock_file =
+            File::create(&lock_path).context(OpenLockFileFailedSnafu { path: &lock_path })?;
```
viviveevee, line 43:
```suggestion
        // It's ok to truncate the file since we rely on OS access control mechanisms, not content of the file
        let lock_file =
            File::create(&lock_path).context(OpenLockFileFailedSnafu { path: &lock_path })?;
```
would have helped me when reviewing, feel free to ignore

PR #168, opened 2025-10-24
feat: file locking for identities

adamspofford-dfinity:


crates/icp/src/identity/key.rs
```diff
@@ -76,26 +81,28 @@ pub fn load_identity(
...
-            password_func, // password_func
-        ),
+        } => {
+            load_pem_identity(
+                dirs,          // dir
+                name,          // name
+                format,        // format
+                algorithm,     // algorithm
+                password_func, // password_func
+            )
```
viviveevee:
```suggestion
            load_pem_identity(
                dirs,
                name,
                format,
                algorithm,
                password_func,
            )
```

feels a bit redundant

PR #169, opened 2025-10-26
fix: Generate a correct schema for the manifests

raymondk:
Generate an accurate schema for icp.yaml

A couple of things are still missing based on the examples:

- [x] Loading networks by path
- [x] loading environments by path
- [ ] Some tests to validate the schema definition against sample yaml file


crates/schema-gen/README.md
```diff
@@ -18,7 +18,7 @@ cargo run --bin schema-gen
 ./scripts/generate-schema.sh
```
lwshang, line 18: The script name here is not consistent with the file name.
Maybe `generate-schemas` is more accurate now.

PR #174, opened 2025-10-30
feat: canister show works in global mode

raymondk:


crates/icp-cli/src/commands/args.rs
```diff
@@ -39,21 +44,62 @@ pub(crate) enum ArgValidationError {
...
+
+    /// Name of the target environment
+    #[arg(long)]
+    pub(crate) environment: Option<Environment>,
+}
+
 #[derive(Args, Debug)]
 pub(crate) struct CanisterCommandArgs {
-    /// Name of canister to target
+    // Note: Could have flattened CanisterEnvironmentArg to avoid adding child field
```
viviveevee, line 60: I don't understand this comment. LLM leftover?

raymondk: It's not the llm 

Since I have one struct with canister and env and another with (canister, env, network, identity) I thought about flattening One into the other instead of duplicating the  the comments. But I decided against it because I didn't want to do:

args.cmd_args.sub_args.canister

PR #175, opened 2025-10-31
chore: move contxet to icp crate

viviveevee:
Follow-up to #170 

Moving `Context` and associated traits to the `icp` crate as @lwshang suggested


viviveevee:
obsolete

PR #177, opened 2025-10-31
chore: more explicit user opt handling

viviveevee:
Makes separation between user-supplied `SomethingOpt` and logical `SomethingSelection`


viviveevee:
obsolete - already implemented

PR #178, opened 2025-10-31
chore: use context in update settings

viviveevee:
https://dfinity.atlassian.net/browse/SDK-2367


crates/icp-cli/src/commands/canister/settings/update.rs
raymondk: Why not use the shared args for this?

viviveevee: Using it now

PR #183, opened 2025-11-03
docs: add yaml schema to examples

ilbertt:
In VSCode or any of its forks with the [YAML](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) extension installed, adding the schema URL at the start of the file enables linting.

After we publish the docs on a static URL, we should use that URL instead of the raw GitHub file URL.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -79,13 +79,13 @@ Your project contains:
 Let's look at the `icp.yaml` file:
 
 ```yaml
-canister:
-  name: my-canister
-  build:
-    steps:
-      - type: script
-        commands:
-          # Build commands specific to your language
+canisters:
+  - name: my-canister
+    build:
+      steps:
+        - type: script
+          commands:
+            # Build commands specific to your language
 ```
 
 ### 3. Start a Local Network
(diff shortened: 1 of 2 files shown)


examples/icp-motoko-recipe/icp.yaml
ilbertt, line 1: Here we caught the error from the linter after adding the schema

examples/icp-rust/icp.yaml
ilbertt, line 1: Same as https://github.com/dfinity/icp-cli/pull/183#discussion_r2486515986 for this file

ilbertt:
Let's wait for the first release to come out, so that we can pin the URL of the schema to it. For this reason, I'm converting this to draft for now.

examples/icp-rust/icp.yaml
```diff
@@ -1,8 +1,10 @@
...
-        commands:
-          - cargo build --target wasm32-unknown-unknown --release --locked
-          - mv target/wasm32-unknown-unknown/release/icp_template_rust.wasm "$ICP_WASM_OUTPUT_PATH"
+  - name: my-canister
+    build:
+      steps:
+        - type: script
+          commands:
+            - cargo build --target wasm32-unknown-unknown --release --locked
+            - mv target/wasm32-unknown-unknown/release/icp_template_rust.wasm "$ICP_WASM_OUTPUT_PATH"
```
ilbertt, line 10: Also fixed in #194.

raymondk (approved):

PR #190, opened 2025-11-06
chore: split out operations

viviveevee:
Splices out some operations from commands. `deploy` can now use the operations directly instead of having to create `command::Args` and executing other commands for every step

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -5,8 +5,8 @@ This document contains the help content for the `icp-cli` command-line program.
 **Command Overview:**
 
 * [`icp-cli`↴](#icp-cli)
+* [`icp-cli build`↴](#icp-cli-build)
 * [`icp-cli canister`↴](#icp-cli-canister)
-* [`icp-cli canister build`↴](#icp-cli-canister-build)
 * [`icp-cli canister call`↴](#icp-cli-canister-call)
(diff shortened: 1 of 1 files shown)


crates/icp-cli/src/operations/create.rs
```diff
@@ -0,0 +1,213 @@
+use candid::{Decode, Encode, Nat, Principal};
```
raymondk, line 1: I understand this is a temporary situation but why not keep things consistent and put all the operations together in the icp-cli crate?

viviveevee: Fair enough. All operations are in the icp-cli crate for now

docs/cli-reference.md
```diff
@@ -5,8 +5,8 @@ This document contains the help content for the `icp-cli` command-line program.
 **Command Overview:**
 
 * [`icp-cli`↴](#icp-cli)
-* [`icp-cli build`↴](#icp-cli-build)
```
raymondk, line 8: I think we should keep:
`icp build`
`icp deploy`
`icp sync`

And have them work in a consistent way where they take a list of canisters. Those commands will always only work in the context of a project.

i don't think `icp canister build|deploy|sync` are necessary.

raymondk (approved):

PR #198, opened 2025-11-17
feat: settings sync command

adamspofford-dfinity:


crates/icp-cli/src/commands/canister/settings/sync.rs
```diff
@@ -0,0 +1,78 @@
...
+    context::{
+        Context, EnvironmentSelection, GetAgentForEnvError, GetCanisterIdForEnvError,
+        GetEnvironmentError,
+    },
+};
+use snafu::Snafu;
+
+#[derive(Debug, Args)]
+pub(crate) struct SyncArgs {
+    name: String,
```
raymondk: Why not use 
```
    #[command(flatten)]
    pub(crate) cmd_args: args::CanisterCommandArgs,
```

Like in:
- https://github.com/dfinity/icp-cli/blob/main/crates/icp-cli/src/commands/canister/settings/show.rs#L14
- https://github.com/dfinity/icp-cli/blob/main/crates/icp-cli/src/commands/canister/settings/update.rs#L77

crates/icp-cli/src/commands/canister/settings/sync.rs
```diff
@@ -0,0 +1,78 @@
...
+        }
+        .fail();
+    }
+
+    let agent = ctx
+        .get_agent_for_env(&args.identity.clone().into(), &environment_selection)
+        .await?;
+    let cid = ctx
+        .get_canister_id_for_env(&args.name, &environment_selection)
+        .await?;
```
raymondk: Can be replaced by a call to `ctx.get_canister_id_and_agent`

see: https://github.com/dfinity/icp-cli/blob/main/crates/icp-cli/src/commands/canister/settings/update.rs#L136

crates/icp-cli/src/commands/canister/settings/mod.rs
```diff
@@ -1,11 +1,13 @@
...
 pub(crate) mod show;
+pub(crate) mod sync;
 pub(crate) mod update;
 
 #[derive(Subcommand, Debug)]
 #[allow(clippy::large_enum_variant)]
 pub(crate) enum Command {
     Show(show::ShowArgs),
     Update(update::UpdateArgs),
+    Sync(sync::SyncArgs),
```
raymondk, line 15: Maybe we add a line of docs for the cli for those commands?

crates/icp-cli/src/commands/canister/settings/sync.rs
```diff
@@ -0,0 +1,78 @@
...
+        return CanisterNotFoundSnafu { name }.fail();
+    };
+
+    if !env.canisters.contains_key(&args.name) {
+        return EnvironmentCanisterNotFoundSnafu {
+            environment: &env.name,
+            name,
+        }
+        .fail();
+    }
```
viviveevee: This can now be

`ctx.assert_env_contains_canister()`

crates/icp-cli/src/commands/canister/settings/update.rs
```diff
@@ -181,21 +192,51 @@ pub(crate) async fn exec(ctx: &Context, args: &UpdateArgs) -> Result<(), Command
         }
     }
     if let Some(compute_allocation) = args.compute_allocation {
+        if configured_settings.compute_allocation.is_some() {
+            eprintln!(
```
viviveevee: We try to do output through `ctx.term`

crates/icp-cli/src/operations/settings.rs
```diff
@@ -0,0 +1,124 @@
...
+                merged_environment_variables
+                    .into_iter()
+                    .map(|(name, value)| EnvironmentVariable { name, value })
+                    .collect_vec(),
+            )
+        } else {
+            None
+        };
+
+    mgmt.update_settings(cid)
```
viviveevee, line 81: Would be nice if we could skip this step if there are no settings to update

crates/icp-cli/src/commands/canister/settings/sync.rs
adamspofford: When I started it `icp canister create` didn't use it, for the same reason I didn't, which is that it only supports canister names and not IDs. I see now that it has been changed, so I'll change it to match

crates/icp-cli/src/commands/canister/settings/sync.rs
```diff
@@ -0,0 +1,75 @@
...
+    let selections = args.cmd_args.selections();
+    let CanisterSelection::Named(name) = &selections.canister else {
+        return PrincipalCanisterSnafu.fail();
+    };
+
+    let p = ctx.project.load().await?;
+
+    let Some((_, canister)) = p.canisters.get(name) else {
+        return CanisterNotFoundSnafu { name }.fail();
+    };
```
raymondk: I believe `ctx.assert_env_contains_canister` will make the right checks so you can skip this part.

adamspofford: It doesn't actually return the canister object though. It would make sense if it did, but there's enough unrelated changes in here already I'd want that in a separate PR.

raymondk (approved):

PR #200, opened 2025-11-18
feat: change generated file path to .icp/{cache, data}

lwshang:
This PR changes where the networks, artifacts, mappings (canister ID) are generated.
In general, `icp/cache/` stores files that are usually ignored by VCS.
While, `.icp/data/` stores files that are important and should be committed to VSC.

```
# ignored
.icp/cache/
  .icp/cache/artifacts/   # move current .icp/artifacts
    .icp/cache/artifacts/<wasm_file>
  .icp/cache/networks/    # move current .icp/networks
    .icp/cache/networks/<managed-network-name>/pocket-ic/
    .icp/cache/networks/<managed-network-name>/state/
    .icp/cache/networks/<managed-network-name>/descriptor.json
  .icp/cache/mappings/    # move current .icpdata/
    .icp/cache/mappings/<environment-name>.ids.json

# persisted
.icp/data/
  .icp/data/mappings/
    .icp/data/mappings/<environment-name>.ids.json
```

### DIFF
--- a/crates/icp/src/manifest/mod.rs
+++ b/crates/icp/src/manifest/mod.rs
@@ -52,7 +52,7 @@ pub struct ProjectRootLocateImpl {
 impl ProjectRootLocateImpl {
     /// Creates a new instance of `ProjectRootLocateImpl`.
     ///
-    /// - If `override` is specified, it will be used as Project Root directly.
+    /// - If `dir` is specified, it will be used as Project Root directly.
     /// - Otherwise, it will search upwards from `cwd` for the project manifest file (`icp.yaml`).
     pub fn new(cwd: PathBuf, dir: Option<PathBuf>) -> Self {
         Self { cwd, dir }


crates/icp/src/manifest/mod.rs
```diff
@@ -26,38 +26,45 @@ pub enum Item<T> {
...
-    /// Specific directory to look in (overrides `cwd`).
+    /// Specific directory to be used as project root directly.
     dir: Option<PathBuf>,
 }
 
-impl Locator {
+impl ProjectRootLocateImpl {
+    /// Creates a new instance of `ProjectRootLocateImpl`.
+    ///
+    /// - If `override` is specified, it will be used as Project Root directly.
```
viviveevee:
```suggestion
    /// - If `dir` is specified, it will be used as Project Root directly.
```
This refers to the function parameter, right?

PR #207, opened 2025-11-20
feat: allow token/cycles commands to target networks

viviveevee:
With this PR it becomes possible to do `icp cycles` and `icp token` operations outside of a project, e.g. when targeting `--mainnet`

### DIFF
--- a/crates/icp-cli/src/commands/deploy/mod.rs
+++ b/crates/icp-cli/src/commands/deploy/mod.rs
@@ -3,7 +3,7 @@ use clap::Args;
 use futures::{StreamExt, future::try_join_all, stream::FuturesOrdered};
 use ic_agent::export::Principal;
 use icp::{
-    context::{Context, EnvironmentSelection, GetEnvCanisterError},
+    context::{CanisterSelection, Context, EnvironmentSelection, GetEnvCanisterError},
     identity::IdentitySelection,
 };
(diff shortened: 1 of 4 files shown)


crates/icp-cli/src/commands/canister/show.rs
```diff
@@ -13,14 +13,14 @@ pub(crate) struct ShowArgs {
...
-    GetCanisterId(#[from] icp::context::GetCanisterIdError),
+    GetCanisterId(#[from] icp::context::GetCanisterIdForEnvError),
 }
 
 pub(crate) async fn exec(ctx: &Context, args: &ShowArgs) -> Result<(), CommandError> {
     let (canister_selection, environment_selection) = args.cmd_args.selections();
 
     let cid = ctx
-        .get_canister_id(&canister_selection, &environment_selection)
+        .get_canister_id_for_env(&canister_selection, &environment_selection)
```
raymondk, line 23: This is not going to work if the canister selection is a principal.
We could:
- use `ctx.get_canister_id` and pass the default network.
- leave it as is for now - the command doesn't do anything super special at the moment - but perhaps add a comment.

This is a weird command: unlike all the other commands which could be used inside or outside the context of project this one doesn't make much sense outside of a project.

raymondk (approved):

PR #212, opened 2025-11-25
refactor: replace CommandError with anyhow::Error

lwshang:
SDK-2423

`thiserror` is removed from `icp-cli` dependencies.

### DIFF
--- a/crates/icp-cli/src/commands/identity/principal.rs
+++ b/crates/icp-cli/src/commands/identity/principal.rs
@@ -14,7 +14,7 @@ pub(crate) async fn exec(ctx: &Context, args: &PrincipalArgs) -> Result<(), anyh
 
     let principal = id
         .sender()
-        .map_err(|_| anyhow::anyhow!("failed to load identity principal"))?;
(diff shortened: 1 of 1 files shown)


crates/icp-cli/src/commands/identity/principal.rs
```diff
@@ -11,24 +9,12 @@ pub(crate) struct PrincipalArgs {
...
-}
-
-pub(crate) async fn exec(ctx: &Context, args: &PrincipalArgs) -> Result<(), PrincipalError> {
+pub(crate) async fn exec(ctx: &Context, args: &PrincipalArgs) -> Result<(), anyhow::Error> {
     let id = ctx.get_identity(&args.identity.clone().into()).await?;
 
     let principal = id
         .sender()
-        .map_err(|message| PrincipalError::Sender { message })?;
+        .map_err(|_| anyhow::anyhow!("failed to load identity principal"))?;
```
raymondk: Are we removing some info that might be useful to the user here?

raymondk (approved):

PR #216, opened 2025-12-02
refactor: remove thiserror anyhow from icp

lwshang:
SDK-2349

### DIFF
--- a/crates/icp/src/canister/prebuilt.rs
+++ b/crates/icp/src/canister/prebuilt.rs
@@ -95,7 +95,7 @@ impl Prebuilt {
 
                 // Check for success
                 if !status.is_success() {
-                    return Err(PrebuiltError::HttpStatus { status });
+                    return HttpStatusSnafu { status }.fail();
                 }
 
                 // Read response body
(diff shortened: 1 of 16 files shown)


crates/icp/src/canister/prebuilt.rs
```diff
@@ -57,30 +89,27 @@ impl Build for Prebuilt {
...
-                    .await
-                    .context("failed to fetch prebuilt canister file")?;
+                let resp = http_client.execute(req).await.context(HttpRequestSnafu)?;
 
                 let status = resp.status();
 
                 // Check for success
                 if !status.is_success() {
-                    return Err(anyhow!("http request failed {status}").into());
+                    return Err(PrebuiltError::HttpStatus { status });
```
raymondk: I thought when using snafu it was common to do:
```return HttpStatusSnafu { status }.fail();```

raymondk: It looks like we are a little inconsistent with this throughout the code so maybe we can make a separate pass to update it everywhere.

raymondk: https://github.com/dfinity/icp-cli/pull/217

lwshang: I wasn't aware of this. Thx!

PR #223, opened 2025-12-08
feat: use network launcher

adamspofford-dfinity:


crates/icp/src/network/managed/run.rs
```diff
@@ -65,7 +69,7 @@ pub enum RunNetworkError {
     #[snafu(transparent)]
     CreateDirFailed { source: crate::fs::IoError },
 
-    #[snafu(display("ICP_POCKET_IC_PATH environment variable is not set"))]
+    #[snafu(display("ICP_CLI_NETWORK_LAUNCHER_PATH environment variable is not set"))]
     NoPocketIcPath,
```
raymondk: Aren't these network launcher errors now?

crates/icp/src/network/managed/run.rs
```diff
@@ -96,40 +100,29 @@ async fn run_pocketic(
                 .as_ref()
                 .map(|lock| lock.claim_port())
                 .transpose()?;
-            eprintln!("PocketIC path: {pocketic_path}");
+            eprintln!("Network launcher path: {network_launcher_path}");
 
             create_dir_all(&root.pocketic_dir()).context(CreateDirAllSnafu)?;
```
raymondk: This is not pocketic anymore either right?

crates/icp/src/network/managed/run.rs
```diff
@@ -382,6 +369,41 @@ pub async fn initialize_instance(
     Ok(props)
 }
 
+pub async fn seed_instance(
```
raymondk, line 276: This is supposed to be done by the network launcher.
I don't think icp-cli should ever see anything related to the admin url of pocket ic.

crates/icp/src/network/managed/launcher.rs
raymondk, line 1: isn't this the network launcher now?

crates/icp/src/network/managed/run.rs
```diff
@@ -404,119 +332,165 @@ pub enum InitializePocketicError {
...
 
 async fn mint_cycles_to_account(
-    pic: &PocketIc,
+    agent: &Agent,
     account: Principal,
     amount: u128,
     icp_xdr_conversion_rate: u64,
-) -> Result<(), InitializePocketicError> {
+) -> Result<(), InitializeNetworkError> {
+    // First withdraw to a different account because notify_mint_cycles will fail if the depositing transaction is a mint TX
```
viviveevee, line 340: Since `acquire_icp_to_account` does not mint ICP we could drop this step

adamspofford: We still need it because the anonymous principal has maxed cycles, so a different principal must receive it, and you strangely can't mint cycles ledger cycles to a different principal than held the ICP. Removing the step would require loading all the identities so they can sign mint messages, which then gets into password protection, etc.

PR #226, opened 2025-12-11
feat: artifact bundling system

lwshang:
SDK-2456

### DIFF
--- a/crates/icp-cli/build.rs
+++ b/crates/icp-cli/build.rs
@@ -61,10 +61,8 @@ fn get_git_version() -> Result<String, std::io::Error> {
 /// or the latest git tag plus sha of current git HEAD at time of build,
 /// or let the cargo.toml version.
 fn define_icp_cli_version() {
-    // option_env! automatically detect changes and trigger rebuilds.
-    // "cargo:rerun-if-env-changed" is not needed here.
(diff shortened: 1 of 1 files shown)


crates/icp-cli/build.rs
```diff
@@ -59,8 +61,10 @@ fn get_git_version() -> Result<String, std::io::Error> {
 /// or the latest git tag plus sha of current git HEAD at time of build,
 /// or let the cargo.toml version.
 fn define_icp_cli_version() {
-    println!("cargo:rerun-if-env-changed=ICP_CLI_VERSION");
-    if let Ok(v) = std::env::var("ICP_CLI_VERSION") {
+    // option_env! automatically detect changes and trigger rebuilds.
+    // "cargo:rerun-if-env-changed" is not needed here.
+    // https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-env-changed
+    if let Some(v) = option_env!("ICP_CLI_VERSION") {
```
adamspofford: This triggers rebuilds of the build script, whereas the previous version only triggered re*runs* of the build script. Generally `env::var` is preferred because changing this value does not actually require rebuilding the build script. The linked guidance refers to what was previously required if `env!` was used in main package code.

PR #235, opened 2025-12-17
feat: Pull docker image if not found

adamspofford-dfinity:


crates/icp/src/network/managed/docker.rs
```diff
@@ -326,6 +343,11 @@ pub enum DockerLauncherError {
         source: bollard::errors::Error,
         container_id: String,
     },
+    #[snafu(display("failed to locate or pull docker image {image}"))]
+    PullContainer {
```
lwshang: Is `PullImage` more accurate?

PR #237, opened 2025-12-17
feat: install candid UI when start network

lwshang:
SDK-2457

### DIFF
--- a/docs/containers.md
+++ b/docs/containers.md
@@ -6,7 +6,7 @@
 # icp.yaml
 networks:
   - name: my-network
-    kind: managed
+    mode: managed
     image: ghcr.io/dfinity/icp-cli-network-launcher
     port-mapping:
       - 0:4943


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * feat: Add support for launching dockerized local networks (#233)
 * fix: When deleting a canister, also delete the id from the id store.
 * chore!: rename `icp network run` to `icp network start
+* feat: install Candid UI canister when start network
```
raymondk:
```suggestion
* feat: install Candid UI canister after starting a local network
```

crates/icp/src/network/managed/run.rs
```diff
@@ -343,12 +351,26 @@ pub enum WaitForPortError {
     ChildExited { source: ChildExitError },
 }
 
-pub async fn seed_instance(
+/// Initialize the network:
+/// - Seed ICP and cycles to the given accounts
```
raymondk, line 356:
```suggestion
/// - Seed ICP and cycles to the given accounts
/// - Install the candid UI canister
```

crates/icp/src/network/managed/run.rs
```diff
@@ -343,12 +351,27 @@ pub enum WaitForPortError {
...
+    candid_ui_wasm: Option<&[u8]>,
+) -> Result<Option<Principal>, InitializeNetworkError> {
     eprintln!("Seeding ICP and TCYCLES account balances");
+    if !seed_accounts
+        .clone()
+        .into_iter()
+        .any(|account| account == Principal::anonymous())
+    {
+        return SeedTokensSnafu {
+            error: "seed_accounts must include the anonymous principal for initialization"
```
raymondk: Why is this an error?

crates/icp/src/network/managed/run.rs
```diff
@@ -564,43 +586,51 @@ async fn get_icp_xdr_conversion_rate(agent: &Agent) -> Result<u64, InitializeNet
...
-                if retries >= 60 {
-                    break Err(PingAndWaitError::Timeout { source: e });
-                }
-                tokio::time::sleep(Duration::from_secs(1)).await;
-                retries += 1;
-            }
+async fn install_candid_ui(
+    agent: &Agent,
+    candid_ui_wasm: &[u8],
+) -> Result<Principal, InitializeNetworkError> {
```
raymondk, line 584: Can we add a debug! log to track the steps?

crates/icp/src/network/managed/run.rs
```diff
@@ -343,12 +351,27 @@ pub enum WaitForPortError {
     ChildExited { source: ChildExitError },
 }
 
-pub async fn seed_instance(
+/// Initialize the network:
+/// - Seed ICP and cycles to the given accounts
+/// - Install the candid UI canister
```
raymondk, line 357:
```suggestion
/// - Install the candid UI canister
///
/// Returns the canister id of the candid ui canister
```

Or something along those lines.

crates/icp/src/network/managed/run.rs
raymondk: Or maybe an `eprintln` I see that we're using that when we're seeding ICP

crates/icp/src/network/managed/run.rs
lwshang: The install_candid_ui step requires that anonymous has sufficient cycles balance. So we need mint_cycles_to_account for anonymous. Since there is always an anonymous identity be default, the current code base ensure that the seed_accounts passed from the caller always contain anonymous.

I added the check just in case we accidentally change the caller logic. Removing it is harmless, if anonymous doesn't have cycles, the canister creation will still fail, just a bit later.

PR #249, opened 2026-01-07
feat: Support prime256v1 and ed25519 keys

adamspofford-dfinity:


crates/icp/src/identity/manifest.rs
```diff
@@ -144,6 +144,13 @@ pub enum IdentityKeyAlgorithm {
     #[strum(serialize = "secp256k1", serialize = "k256")]
     #[cfg_attr(feature = "clap", value(alias = "k256"))]
     Secp256k1,
+    #[serde(rename = "prime256v1", alias = "p256", alias = "secp256r1")]
+    #[strum(serialize = "prime256v1", serialize = "p256", serialize = "secp256r1")]
+    #[cfg_attr(feature = "clap", value(alias = "p256", alias = "secp256r1"))]
+    Prime256v1,
```
lwshang, line 150: What is the specific benefit of maintaining these aliases?

The interface specification explicitly uses `secp256r1` and `secp256k1`. To ensure consistency and reduce cognitive load across the codebase, I propose we stick exclusively to the formal naming conventions defined in the spec.

adamspofford: I chose the name prime256v1 to make them visually distinct to users who don't know anything about cryptography. A difference of a single letter is easily overlooked, especially when the letter in question doesn't appear in the common abbreviation. The identity type in ic-agent was named that way for the same reason. The aliases then refer to the other official names so you don't have to remember which name the tool uses if you're used to the behavior of another tool (most tools do this, supporting just these three).

PR #255, opened 2026-01-09
feat: validate call argument against fetched candid interface

lwshang:


crates/icp-cli/src/commands/canister/call.rs
```diff
@@ -41,14 +44,21 @@ pub(crate) async fn exec(ctx: &Context, args: &CallArgs) -> Result<(), anyhow::E
...
-        .with_arg(
-            cargs
-                .to_bytes()
-                .context("failed to serialize candid arguments")?,
-        )
-        .await?;
+    let arg_bytes = if let Some((env, method)) = get_candid_type(&agent, cid, &args.method).await {
+        cargs
+            .to_bytes_with_types(&env, &method.args)
+            .context("failed to serialize candid arguments with specific types")?
```
raymondk, line 50: Is this message going to be helpful enough?

lwshang: The underlying candid error will also be displayed as the cause.

An example with the cycles-ledger canister:

```
$ icp canister call --mainnet um5iw-rqaaa-aaaaq-qaaba-cai icrc2_allowance "(1)"
Error: failed to serialize candid arguments with specific types

Caused by:
    type mismatch: 1 can not be of type record { account : Account; spender : Account }
```

I don't think there is any trivial improvement that we can implement on the icp-cli side. It has to be some fundamental change in the `candid` implementation to make the error message more readable.

PR #256, opened 2026-01-09
feat: display canister URLs after deploying

viviveevee:
New deploy output:

```
[...]

Syncing canisters:
[frontend] ✔ Synced successfully: t63gs-up777-77776-aaaba-cai                                                                                                                                                                                                   
Canisters synced successfully


Deployed canisters:
  frontend: http://t63gs-up777-77776-aaaba-cai.localhost:8000
  backend (Candid UI): http://tqzl2-p7777-77776-aaaaa-cai.localhost:8000/?id=txyno-ch777-77776-aaaaq-cai

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 * feat: `icp deploy` now displays URLs to interact with the deployed canister(s)
 * feat: Allow overriding the `local` network in the config file
   * This makes it more convenient to configure the default environment
+* feat: Validate call argument against candid interface
+  * The interface is fetched from canister metadata onchain
 
 # v0.1.0-beta.2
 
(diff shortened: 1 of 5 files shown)


crates/icp-canister-interfaces/src/candid_ui.rs
```diff
@@ -0,0 +1 @@
+pub const CANDID_UI_CID: &str = "a4gq6-oaaaa-aaaab-qaa4q-cai";
```
adamspofford: Maybe this is clearer?
```suggestion
pub const MAINNET_CANDID_UI_CID: &str = "a4gq6-oaaaa-aaaab-qaa4q-cai";
```

viviveevee: good idea!

PR #257, opened 2026-01-09
feat: Add keyring support

adamspofford-dfinity:


crates/icp-cli/src/commands/identity/import.rs
```diff
@@ -20,11 +20,16 @@ use snafu::{OptionExt, ResultExt, Snafu, ensure};
...
 
+use crate::commands::identity::StorageMode;
+
 #[derive(Debug, Args)]
 #[command(group(ArgGroup::new("import-from").required(true)))]
 pub(crate) struct ImportArgs {
     name: String,
 
+    #[arg(long, value_enum, default_value_t)]
+    storage: StorageMode,
```
raymondk, line 33: Add docs?

crates/icp-cli/src/commands/identity/new.rs
```diff
@@ -12,9 +12,13 @@ use icp::{
 
 use icp::context::Context;
 
+use crate::commands::identity::StorageMode;
+
 #[derive(Debug, Args)]
 pub(crate) struct NewArgs {
     name: String,
+    #[arg(long, value_enum, default_value_t)]
```
raymondk, line 23: Add docs?

PR #259, opened 2026-01-09
feat: Accept an environment as argument for network commands

raymondk:


crates/icp-cli/src/commands/network/status.rs
```diff
@@ -1,14 +1,35 @@
...
+///   icp network status mynetwork
+///
+///   # Get status using environment flag
+///   icp network status -e staging
+///
+///   # Get status using ICP_ENVIRONMENT variable
+///   ICP_ENVIRONMENT=staging icp network status
+///
+///   # Name overrides ICP_ENVIRONMENT
+///   ICP_ENVIRONMENT=staging icp network status local
```
viviveevee: I don't think this renders the way you would like it to, both on the CLI and in the .md file:

Image

crates/icp-cli/src/commands/network/ping.rs
```diff
@@ -1,38 +1,54 @@
...
 
+use super::args::NetworkOrEnvironmentArgs;
 use icp::context::Context;
 
 /// Try to connect to a network, and print out its status.
 #[derive(Args, Debug)]
+#[command(after_long_help = "\
+Examples:
+    # Ping default 'local' network
+    icp network ping
```
lwshang, line 17: The first example in the block is not rendered correctly.

Image

An extra newline after `Examples:` does the trick.

PR #261, opened 2026-01-12
feat: Add password-protected identity support

adamspofford-dfinity:


crates/icp/src/identity/mod.rs
```diff
@@ -103,17 +103,22 @@ pub trait Load: Sync + Send {
...
+
 pub struct Loader {
     pub dir: IdentityDirectories,
+    pub password_func: PasswordFunc,
 }
 
 #[async_trait]
 impl Load for Loader {
     async fn load(&self, id: IdentitySelection) -> Result<Arc<dyn Identity>, LoadError> {
+        let password_func = &self.password_func;
```
lwshang, line 117: The `password_func` closure is constructed conditionally in `main.rs` and passed down through `icp::context::initialize(..., password_func)` and `identity::Loader {..., password_func}`.

Can we instead only passing the optional `identity_password_file` path and constructing the `password_func` closure here? Therefore, the `PasswordFunc` type and related logic can be kept private within the `identity` module.

adamspofford: The goal of the parameter was so that icp, the library, would not be doing any terminal-prompting. icp doesn't depend on dialoguer for this reason, only icp-cli does.

lwshang (approved):

PR #263, opened 2026-01-13
feat: pass init args

viviveevee:
- allow specifying canister `init_args` in `icp.yaml`, both on canister and environment level
- fix canister settings override on environment level in `icp.yaml`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -94,6 +94,14 @@ jobs:
           which mops
           mops --version
 
+      - name: Install ic-wasm
+        run: curl --proto '=https' --tlsv1.2 -LsSf https://github.com/dfinity/ic-wasm/releases/download/0.9.10/ic-wasm-installer.sh | sh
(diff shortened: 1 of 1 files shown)


crates/icp-cli/src/operations/init_args.rs
```diff
@@ -0,0 +1,46 @@
...
+    // Try to decode as hex first
+    if let Ok(bytes) = hex::decode(init_args_str) {
+        return Ok(bytes);
+    }
+
+    // Otherwise, parse as Candid text format
+    let args =
+        parse_idl_args(init_args_str).context("Failed to parse init_args as Candid text format")?;
+
+    args.to_bytes()
```
adamspofford, line 17: This should be to_bytes_with_types, same as icp canister call

viviveevee: Sadly we generally do not have access to init types

PR #267, opened 2026-01-14
feat: Windows support

adamspofford-dfinity:


.github/workflows/test.yml
```diff
@@ -11,6 +11,14 @@ env:
   CARGO_NET_RETRY: 10
   # Use the local .curlrc
   CURL_HOME: .
+  # Disable incremental compilation
+  CARGO_INCREMENTAL: 0
```
lwshang, line 15: Any issues with incremental compilation? Disabling it may slow CI.

crates/icp/src/context/mod.rs
```diff
@@ -492,6 +496,49 @@ impl Context {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct TermWriter {
+    pub debug: bool,
```
lwshang, line 501: The `debug` flag semantics are a bit confusing - when `debug: true`, terminal output is suppressed (only goes to logs). Intuitively, I'd expect `--debug` to mean "show more output", not less.

Could you clarify the intent here? If the purpose is "redirect output to logs only when in debug mode", would a name like `log_only` or `suppress_terminal` be clearer?

crates/icp/src/context/mod.rs
```diff
@@ -492,6 +496,49 @@ impl Context {
...
+    }
+}
+
+impl Write for &TermWriter {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        if !self.debug {
+            (&self.raw_term).write(buf)?;
+        }
+        debug!("{}", String::from_utf8_lossy(buf).trim());
+        Ok(buf.len())
```
lwshang:
```suggestion
        let written = if !self.debug {
            (&self.raw_term).write(buf)?
        } else {
            buf.len()
        };
        debug!("{}", String::from_utf8_lossy(&buf[..written]).trim());
        Ok(written)
```

The `Write` trait contract states that `write()` returns the number of bytes **actually** written.

.github/workflows/test.yml
adamspofford: The speed increase is tiny, on the order of seconds to improve development. But the size consumed is much larger. We're adding a third parallel cache, so the size needs to go down to keep evictions uncommon. I plan to shrink it further in another PR.

crates/icp/src/context/mod.rs
adamspofford: This is reproducing the meaning of the original code. No output is suppressed. Instead the debug flag ensures that regular output is rerouted into the logging framework so it matches the formatting.

crates/icp/src/context/mod.rs
```diff
@@ -492,6 +496,51 @@ impl Context {
...
+
+impl Write for &TermWriter {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        let written = if !self.debug {
+            (&self.raw_term).write(buf)?
+        } else {
+            buf.len()
+        };
+        debug!("{}", String::from_utf8_lossy(&buf[..written]).trim());
+        Ok(written)
```
lwshang: The terminal output `(&self.raw_term).write(buf)?` is not executed when `self.debug` is `true`. That's why I said "the output is suppressed".

The log output `debug!("{}", String::from_utf8_lossy(&buf[..written]).trim());` is executed unconditionally.

adamspofford: Right, the either/or there is tacit, since if `debug` is false, debug mode is off, and so `debug!` is a no-op. Not that this makes any sense to leave in, I was just copying the behavior of the existing code.

lwshang (approved):

PR #268, opened 2026-01-15
chore: validate examples workflow

viviveevee:
Add example validation workflow. For every example in `/examples`:
* if `test.sh` is present it runs the script
* otherwise run `icp project show`

### DIFF
--- a/.github/workflows/validate-examples.yml
+++ b/.github/workflows/validate-examples.yml
@@ -44,51 +44,7 @@ jobs:
         run: cargo build
 
       - name: Validate examples
-        run: |
-          # Add icp binary to PATH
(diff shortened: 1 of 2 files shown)


.github/workflows/validate-examples.yml
```diff
@@ -0,0 +1,101 @@
...
+              echo "✅ Project validation successful for $example_name"
+            fi
+            
+            popd
+          done
+          
+          echo ""
+          echo "=========================================="
+          echo "✅ All examples validated successfully!"
+          echo "=========================================="
```
lwshang: What about extracting this script to `scripts/` so that it can also be executed locally?

PR #269, opened 2026-01-15
feat: icp canister metadata

viviveevee:
As requested on Slack: `icp canister metadata`
Also tested manually against mainnet

### DIFF
--- a/.github/workflows/validate-examples.yml
+++ b/.github/workflows/validate-examples.yml
@@ -1,0 +1,57 @@
+name: Validate Examples
+
+on:
+  push:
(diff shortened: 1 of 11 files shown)


raymondk (approved):
LGTM 
Should we consider adding `--json` for consistency with other commands?

raymondk (changes requested):
There are args structs that can be reused from other commands: `args::CanisterCommandArgs`

raymondk:
add json formatting: https://github.com/dfinity/icp-cli/pull/272

PR #274, opened 2026-01-16
feat: Show `name` in `canister status` command

raymondk:


crates/icp-cli/src/commands/canister/status.rs
```diff
@@ -60,16 +60,19 @@ async fn get_principals(
...
         Some(canister) => {
             let canister_selection: CanisterSelection = canister.clone().into();
             let cid = ctx
                 .get_canister_id(&canister_selection, network, environment)
                 .await?;
-            cids.push(cid);
+            match canister {
+                args::Canister::Name(name) => cids.push((Some(name), cid)),
+                args::Canister::Principal(_) => cids.push((None, cid)),
+            };
```
viviveevee, line 75: If I specify a named canister by principal this will not pick up the name. Is that the intended behavior?

raymondk: Yes...
I didn't think about doing a lookup for the name. I assumed if you were specifying the id that you don't care about the name.

Maybe it's useful to add it though

raymondk: After giving it some thought it seemed like there would be weird corner cases to handle.
For eg - you can call `icp canister status` with a network or with an environment. It's fine to reverse lookup the name for an environment but if you do it for a network you'd need to check all the environments for that network and it doesn't seem to be worth it.

PR #275, opened 2026-01-16
chore!: use --tcycles for consistency

viviveevee:
Sometimes users are expected to provide exact cycles amounts, sometimes `TCYCLES` amounts. With this PR it is always `TCYCLES`


viviveevee:
superseded by #293

PR #276, opened 2026-01-16
fix: error if explicit canister path is bad

viviveevee:
When explicit (non-glob) canister paths in icp.yaml didn't exist or lacked a canister.yaml file, they were silently ignored instead of throwing an error.

Now glob patterns get treated differently from explicit paths and if an explicit path does not exist, `icp project show` throws an error

### DIFF
--- a/crates/icp/src/project.rs
+++ b/crates/icp/src/project.rs
@@ -175,14 +175,7 @@ pub async fn consolidate_manifest(
                     // For explicit paths, validate that they exist and contain canister.yaml
                     let mut validated_paths = vec![];
                     for p in paths {
-                        if !p.is_dir() {
-                            return NotFoundSnafu {
-                                kind: "canister".to_string(),
-                                path: pattern.to_string(),
-                            }
-                            .fail();
-                        }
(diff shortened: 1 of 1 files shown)


crates/icp/src/project.rs
```diff
@@ -163,11 +164,35 @@ pub async fn consolidate_manifest(
...
+                    paths
+                        .into_iter()
+                        .filter(|p| p.is_dir())
+                        .filter(|p| p.join(CANISTER_MANIFEST).exists())
+                        .collect::<Vec<_>>()
+                } else {
+                    // For explicit paths, validate that they exist and contain canister.yaml
+                    let mut validated_paths = vec![];
+                    for p in paths {
+                        if !p.is_dir() {
```
raymondk: Can we just use `p.is_file()` ?

raymondk: Otherwise the errors should be fixed, p.is_dir to say the directory doesn't exist and the path joined with `canister.yaml` to say the canister definition doesn't exist.

raymondk (approved):

PR #278, opened 2026-01-16
feat!: rename implicit "mainnet" network to "ic" and define implicit "ic" environment

lwshang:
SDK-2509

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -130,7 +130,7 @@ Make a canister call
 
 ###### **Options:**
 
-* `--network <NETWORK>` — Name of the network to target, conflicts with environment argument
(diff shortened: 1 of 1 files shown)


adamspofford-dfinity (approved):
Missing a CLAUDE.md update

crates/icp-cli/src/options.rs
```diff
@@ -92,17 +69,10 @@ pub(crate) struct NetworkOpt {
     /// Name of the network to target, conflicts with environment argument
     #[arg(long, env = "ICP_NETWORK", group = "network-select", help_heading = heading::NETWORK_PARAMETERS)]
```
raymondk: Add a shorthand `-n` for network ?

PR #287, opened 2026-01-21
docs: add Starlight documentation site and reorganize containerized networks guide

marc0olo:


.github/workflows/docs.yml
```diff
@@ -0,0 +1,71 @@
+name: Documentation
+
+on:
+  push:
+    branches:
+      - main
```
raymondk, line 6: It's probably fine for now but we should eventually change this to update the docs when we push a specific tag.

marc0olo: Another option is to remove the pipeline for now.

raymondk (approved):

PR #288, opened 2026-01-21
feat: support large wasms

### LINKED ISSUE #273: Error during deployment of large size canister modules
The following error occurs when attempting to install a WASM file of several MBs using `icp deploy`.

```
Error: agent error: The replica returned an HTTP Error: Http Error: status 413 Payload Too Large, content type "text/plain; charset=utf-8", content: error: payload_too_large
details: Payload is too large: maximum body size is 4194304 bytes.
```

This can be avoided by changing `ManagementCanister.install_code` to `ManagementCanister.install` in [`install.rs`](https://github.com/dfinity/icp-cli/blob/07ee1d0228002b960e47f1a83668b0de4ae35ef4/crates/icp-cli/src/operations/install.rs), which automatically applies chunked installation based on the size. However, if this is used on canisters that were manually chunk-installed, it will clear the chunked code storage.

Which approach would be preferable: adding an option like `icp deploy --auto-chunk`, or switching the standard behavior to automatic chunking as a breaking change?

viviveevee:
Fixes #273

### DIFF
--- a/docs-site/astro.config.mjs
+++ b/docs-site/astro.config.mjs
@@ -4,8 +4,8 @@ import rehypeExternalLinks from 'rehype-external-links';
 
 // https://astro.build/config
 export default defineConfig({
-  site: 'https://dfinity.github.io/icp-cli',
-  base: '/',
+  site: 'https://dfinity.github.io',
+  base: '/icp-cli/',
   markdown: {
     rehypePlugins: [
       // Open external links in new tab


crates/icp-cli/tests/assets/3mb_hello_world.wasm
lwshang, line 1: I'd like to avoid committing large binaries to the repo if possible.

A few questions:
- How was this 3MB Wasm module generated?
- Could we replace it with a script or dummy project that generates the binary at build/test time?

If we don't need to verify canister installation by making actual calls (i.e., the module doesn't need functional endpoints), we could generate a dummy Wasm module padded to the required size. Something like a minimal module with a large data segment would be valid Wasm but contain no real logic.

viviveevee: I `include_bytes!`'d some random data

We can also generate it at test time, but Rust builds usually take quite some time so I'd like to avoid this. If you know of another way to simply bloat a wasm then that would be great. Calling endpoints is optional, checking that the module hash is good enough in this case

PR #293, opened 2026-01-22
refactor!: Overhaul cycles handling - new amount formats, display changes, and cycles transfer command

viviveevee:
* Get rid of any mentions of `TCYCLES` in favor of `cycles`
* Cycles amounts now display with underscores
* Pretend the cycles ledger has 0 decimals and the symbol `cycles`
* Support new formats in token and cycle inputs:
  * `_` is allowed and ignored: `1_000` parses to `1000`
  * `k`, `m`, `b`, `t` suffixes (case insensitive) to avoid typing a bunch of zeroes
  * allow arbitrary combinations thereof: `1_234.56K` parses to `1234560`

### DIFF
--- a/crates/icp-cli/src/commands/token/mod.rs
+++ b/crates/icp-cli/src/commands/token/mod.rs
@@ -5,9 +5,9 @@ pub(crate) mod transfer;
 
 #[derive(Debug, Parser)]
 pub(crate) struct Command {
-    /// The token or principal to execute the operation on, defaults to `icp`
+    /// The token or ledger canister id to execute the operation on, defaults to `icp`
     #[arg(default_value = "icp")]
-    pub(crate) token_name_or_principal: String,
+    pub(crate) token_name_or_ledger_id: String,
 
     #[command(subcommand)]
     pub(crate) command: Commands,
--- a/crates/icp-cli/src/main.rs
+++ b/crates/icp-cli/src/main.rs
@@ -406,13 +406,13 @@ async fn main() -> Result<(), Error> {
         // Token
         Command::Token(cmd) => match cmd.command {
             commands::token::Commands::Balance(args) => {
-                commands::token::balance::exec(&ctx, &cmd.token_name_or_principal, &args)
+                commands::token::balance::exec(&ctx, &cmd.token_name_or_ledger_id, &args)
                     .instrument(trace_span)
                     .await?
             }
 
             commands::token::Commands::Transfer(args) => {
-                commands::token::transfer::exec(&ctx, &cmd.token_name_or_principal, &args)
+                commands::token::transfer::exec(&ctx, &cmd.token_name_or_ledger_id, &args)
                     .instrument(trace_span)
                     .await?
             }
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -920,7 +920,7 @@ Synchronize canisters
 
 Perform token transactions
 
-**Usage:** `icp token [TOKEN_NAME_OR_PRINCIPAL] <COMMAND>`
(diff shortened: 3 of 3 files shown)


crates/icp-cli/src/commands/parsers.rs
```diff
@@ -0,0 +1,373 @@
...
+            'b' => (&input[..input.len() - 1], 1_000_000_000u128),
+            't' => (&input[..input.len() - 1], 1_000_000_000_000u128),
+            _ => (input, 1u128),
+        }
+    } else {
+        (input, 1u128)
+    };
+
+    // Remove underscores from the number part
+    let cleaned = number_part.replace('_', "");
```
raymondk, line 40: If someone bothers to put underscores, shouldn't we be checking that they're in the correct places?

eg: `10_00` should fail?

lwshang: Rust numeric literal allows underscores at arbitrary positions (e.g., 10_00, 1___0 are all valid), so I'd say we just follow that precedent and keep it simple.

raymondk:
Notes from standup:
- We'll remove the special handling of 'cycles' in the `icp token` command.
- We will add a `icp cycles transfer` command
- `icp token` will be used to handle ICRC1 tokens without and special treatment (no mention of TCYCLES for now)
- clarify the docs for `icp token` to say that the TOKEN arg is a name or an ICRC1 ledger id

crates/icp-cli/src/commands/cycles/mint.rs
```diff
@@ -42,7 +45,7 @@ pub(crate) async fn exec(ctx: &Context, args: &MintArgs) -> Result<(), anyhow::E
 
     // Display results
     let _ = ctx.term.write_line(&format!(
-        "Minted {} TCYCLES to your account, new balance: {} TCYCLES.",
+        "Minted {} to your account, new balance: {}.",
```
raymondk, line 48:
```suggestion
        "Minted {} cycles to your account, new balance: {} cycles.",
```

crates/icp-cli/src/commands/token/transfer.rs
```diff
@@ -39,11 +42,8 @@ pub(crate) async fn exec(
 
     // Output information
     let _ = ctx.term.write_line(&format!(
-        "Transferred {} {} to {} in block {}",
-        transfer_info.amount,
-        transfer_info.symbol,
-        transfer_info.receiver,
-        transfer_info.block_index
+        "Transferred {} to {} in block {}",
+        transfer_info.transferred, transfer_info.receiver, transfer_info.block_index
```
raymondk, line 46: Why not show the symbol here?

crates/icp-cli/src/commands/token/mod.rs
```diff
@@ -5,9 +5,9 @@ pub(crate) mod transfer;
 
 #[derive(Debug, Parser)]
 pub(crate) struct Command {
-    /// The token to execute the operation on, defaults to `icp`
+    /// The token or principal to execute the operation on, defaults to `icp`
```
adamspofford: I would phrase this 'or ledger canister ID', you might misinterpret 'principal' as the tx recipient

crates/icp-cli/src/commands/cycles/mint.rs
viviveevee: The symbol is included in the struct that is displayed

crates/icp-cli/src/commands/token/transfer.rs
viviveevee: Same here, it's in the displayed struct

PR #302, opened 2026-01-23
feat: support AccountIdentifier hex strings in `icp token transfer`

lwshang:
The `icp token transfer` command now accepts AccountIdentifier hex strings as the receiver for ICP ledger transfers. When a 64-character hex string is provided, the command automatically dispatches to the legacy ICP ledger transfer method which uses AccountIdentifier destinations.

AccountIdentifier format is only supported for the ICP ledger; other tokens continue to require Principal format and use ICRC-1 transfer.

Changes:
- Add `icp_legacy_transfer()` function for legacy ICP ledger transfers
- Update transfer command to accept receiver as String instead of Principal
- Automatically detect and parse AccountIdentifier vs Principal format
- Update TransferInfo to use receiver_display field for flexible display
- Add comprehensive test for AccountIdentifier transfers

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -920,16 +920,16 @@ Synchronize canisters
 
 Perform token transactions
 
-**Usage:** `icp token [TOKEN_NAME_OR_LEDGER_ID] <COMMAND>`
+**Usage:** `icp token [TOKEN|LEDGER_ID] <COMMAND>`
 
 ###### **Subcommands:**
 
-* `balance` — 
-* `transfer` — 
+* `balance` — Display the token balance on the ledger (default token: icp)
+* `transfer` — Transfer ICP or ICRC1 tokens through their ledger (default token: icp)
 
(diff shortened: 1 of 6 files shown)


docs/reference/cli.md
raymondk: The docs here are not complete.
They should mention that you can pass an optional TOKEN name or a ledger id.

raymondk: Maybe we can do this in follow up PR because `icp token balance` is also not accurate.

lwshang: That is documented above
```
## `icp token`

Perform token transactions

**Usage:** `icp token [TOKEN_NAME_OR_LEDGER_ID] <COMMAND>`

###### **Subcommands:**

* `balance` — 
* `transfer` — 

###### **Arguments:**

* `<TOKEN_NAME_OR_LEDGER_ID>` — The token or ledger canister id to execute the operation on, defaults to `icp`
```

PR #305, opened 2026-01-26
docs: make installation guide platform-neutral and clarify ic-wasm requirements

marc0olo:
### Platform Neutrality
- Prioritize cross-platform installers (curl/PowerShell) over Homebrew across all docs
- Remove macOS-centric language and positioning
- Treat all platforms (macOS, Linux, Windows) equally

### ic-wasm Clarity
- Document that ic-wasm is required for Rust templates (default recipes use shrinking)
- Clarify it's recommended for Motoko (provides ~10% cycle reduction, ~16% size reduction)
- Note that Homebrew auto-installs ic-wasm as a dependency
- Use `/latest/` URLs to reduce version maintenance burden

### DRY Principle
- Tutorial now references installation guide instead of duplicating instructions
- Single source of truth for installation steps


docs/guides/installation.md
```diff
@@ -61,6 +58,34 @@ npm install -g ic-mops
...
+
+`ic-wasm` is a WebAssembly post-processing tool that optimizes canisters for the Internet Computer. It provides:
+- **Optimization**: ~10% cycle reduction for Motoko, ~4% for Rust
+- **Size reduction**: ~16% smaller binaries for both languages
+- **Metadata**: Embed Candid interfaces and version information
+- **Shrinking**: Remove unused code and debug symbols
+
+**When is it needed?**
+- **Rust projects**: Required for default templates (they use ic-wasm for shrinking)
+- **Motoko projects**: Optional, but recommended for production (significant performance and size benefits)
```
viviveevee: I don't think that's accurate. The Motoko recipe exits if ic-wasm is not installed: https://github.com/dfinity/icp-cli-recipes/blob/6188a49e4993d490c858fed89121f09bfdf2d4ca/recipes/motoko/recipe.hbs#L17

In dfx we just introduced this behavior, so maybe you're mixing it up with that?

marc0olo: yes, I updated the docs accordingly now

PR #307, opened 2026-01-26
feat: call/install argument improvements

viviveevee:
* `icp canister call` now also can take arguments in hex
* Everywhere that takes arguments (`canister call`, `canister install`, `icp.yaml::canister::init_args`) now also accepts file paths that point to files that contain hex or Candid strings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+* feat: Bind Docker networks to 127.0.0.1
 * feat: Add IC options to network manifest (`ii`, `nns`, `subnets`, `artificial-delay-ms`)
 * feat: Release for Windows
 * feat: Add safety controls for `--set-controller` and `--remove-controller`
--- a/crates/icp-cli/src/commands/build.rs
+++ b/crates/icp-cli/src/commands/build.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use clap::Args;
 use futures::future::try_join_all;
 use icp::context::{Context, EnvironmentSelection};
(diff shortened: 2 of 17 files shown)


crates/icp-cli/src/operations/misc.rs
```diff
@@ -15,3 +20,242 @@ pub async fn fetch_canister_metadata(
...
+    let file_path = if args_str.starts_with('/') {
+        // Absolute path
+        PathBuf::from(args_str)
+    } else {
+        // Relative path - resolve relative to base_path
+        base_path.as_ref().join(args_str)
+    };
+
+    // Try to read the file
+    if let Ok(contents) = std::fs::read_to_string(file_path.as_std_path()) {
```
adamspofford: If it was intended to be a file path, but the file doesn't exist, this is going to be a confusing error.

crates/icp-cli/src/operations/misc.rs
```diff
@@ -15,3 +20,242 @@ pub async fn fetch_canister_metadata(
...
+        return Ok(ParsedArguments::Hex(bytes));
+    }
+
+    // Try to parse as Candid text format
+    if let Ok(args) = parse_idl_args(args_str) {
+        return Ok(ParsedArguments::Candid(args));
+    }
+
+    // If neither hex nor Candid, try to read as a file path
+    let file_path = if args_str.starts_with('/') {
```
adamspofford: This isn't cross-platform, and also isn't needed - `join` with an absolute path will just replace the whole thing with the absolute path.

crates/icp-cli/src/operations/misc.rs
viviveevee: I rewrote it a bit, but I don't think the error is a problem. If the `read_to_string` fails we don't propagate the error from this failure

PR #312, opened 2026-01-27
feat: add log_visibility to canister settings manifest

marc0olo:
## Summary

Add support for configuring `log_visibility` in `icp.yaml` and `canister.yaml` manifest files. This allows users to control who can read canister logs directly in their project configuration:

```yaml
settings:
  log_visibility: public  # or "controllers"
```

Previously, `log_visibility` could only be changed via the CLI command `icp canister settings update --log-visibility`.

### Core functionality
- Add `LogVisibility` enum to `icp::canister` module with `JsonSchema` support for manifest validation
- Add `log_visibility` field to `Settings` struct  
- Update `canister settings sync` to apply `log_visibility` during deployment
- Pass `log_visibility` to cycles ledger during `canister create` and `deploy`
- Regenerate JSON schemas

### Files changed
- `crates/icp/src/canister/mod.rs` - Add `LogVisibility` enum and conversions
- `crates/icp-canister-interfaces/src/cycles_ledger.rs` - Add `LogVisibility` to `CanisterSettingsArg`
- `crates/icp-cli/src/commands/canister/create.rs` - Pass `log_visibility` during creation
- `crates/icp-cli/src/operations/settings.rs` - Apply `log_visibility` during sync
- `docs/schemas/*.json` - Updated JSON schemas

## Testing

Added comprehensive tests:
- `canister_create_with_settings` - Tests manifest `log_visibility` with create + sync workflow
- `canister_settings_sync_log_visibility` - Tests syncing `log_visibility` from manifest after deploy

### Note on PocketIC limitation

The test environment (PocketIC) uses a fake-cmc that doesn't support `log_visibility` during canister creation. The **real cycles ledger on mainnet does support it**. The tests work around this by syncing settings after creation, which uses the management canister directly and works correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Unreleased
 
+# v0.1.0-beta.5
+
+* fix: Fix error when loading network descriptors from v0.1.0-beta.3
 * feat: `icp identity delete` and `icp identity rename`
+* feat: `icp identity export` to print the PEM file for the identity
 
 # v0.1.0-beta.4
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3107,7 +3107,7 @@ dependencies = [
 
 [[package]]
 name = "icp"
-version = "0.1.0-beta.4"
+version = "0.1.0-beta.5"
 dependencies = [
  "async-dropper",
  "async-trait",
@@ -3174,7 +3174,7 @@ dependencies = [
 
 [[package]]
 name = "icp-canister-interfaces"
-version = "0.1.0-beta.4"
+version = "0.1.0-beta.5"
 dependencies = [
  "bigdecimal",
  "candid",
@@ -3183,7 +3183,7 @@ dependencies = [
 
 [[package]]
 name = "icp-cli"
-version = "0.1.0-beta.4"
+version = "0.1.0-beta.5"
 dependencies = [
  "anstyle",
  "anyhow",
@@ -5451,7 +5451,7 @@ dependencies = [
 
 [[package]]
 name = "schema-gen"
-version = "0.1.0-beta.4"
+version = "0.1.0-beta.5"
 dependencies = [
  "icp",
  "schemars 1.1.0",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ resolver = "3"
 [workspace.package]
 authors = ["DFINITY Stiftung <sdk@dfinity.org>"]
 edition = "2024"
-version = "0.1.0-beta.4"
+version = "0.1.0-beta.5"
 repository = "https://github.com/dfinity/icp-cli"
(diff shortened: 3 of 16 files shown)


crates/icp-cli/src/operations/settings.rs
```diff
@@ -79,7 +84,10 @@ pub(crate) async fn sync_settings(
         } else {
             None
         };
-    if compute_allocation.is_none_or(|s| s == current_settings.compute_allocation)
+    if log_visibility_setting
+        .as_ref()
+        .is_none_or(|s| *s == current_settings.log_visibility)
```
raymondk: is this going to work when the settings are for allowed principals and the array is not in the same order or do we not care?

marc0olo: actually is was not possible to set allowed viewers via config yet. I just added this, too.

the updated code now also checks if the values for `allowed_viewers` (`log_visibility`) and `environment_variables` have changed so that we avoid unnecessary update calls.

or do we want to always trigger an update call, also if just the order is changed? 🤔

crates/icp/src/canister/mod.rs
```diff
@@ -12,21 +12,119 @@ pub mod sync;
 mod script;
 
 /// Controls who can read canister logs.
-#[derive(Clone, Debug, Default, Deserialize, PartialEq, JsonSchema, Serialize)]
-#[serde(rename_all = "lowercase")]
+#[derive(Clone, Debug, Default, PartialEq, Serialize)]
+#[serde(rename_all = "snake_case")]
 pub enum LogVisibility {
```
adamspofford, line 16: Serialization, deserialization, and the schema shouldn't have three different structures. If there's multiple ways it can look, it should use multiple serde entities.

crates/icp-cli/src/operations/settings.rs
```diff
@@ -79,20 +110,24 @@ pub(crate) async fn sync_settings(
         } else {
             None
         };
-    if compute_allocation.is_none_or(|s| s == current_settings.compute_allocation)
+    if log_visibility_setting
+        .as_ref()
+        .is_none_or(|s| log_visibility_eq(s, &current_settings.log_visibility))
```
adamspofford, line 115: Please also update `icp canister settings update` to warn for this field like the others.

crates/icp-cli/tests/canister_settings_tests.rs
```diff
@@ -1131,3 +1131,132 @@ async fn canister_settings_sync() {
     sync(&ctx, &project_dir);
     confirm_wasm_memory_limit(&ctx, &project_dir, "5_368_709_120");
 }
+
+#[tokio::test]
+async fn canister_settings_sync_log_visibility() {
```
adamspofford, line 1136: Maybe add a test case for allowed_principals?

PR #319, opened 2026-01-28
feat: add icp identity export command 

lwshang:
Allows users to export their identities as plaintext PEM files for backup or migration purposes. Supports all storage types (plaintext, password-protected, keyring) and includes password file input option.

### DIFF
--- a/crates/icp/src/identity/key.rs
+++ b/crates/icp/src/identity/key.rs
@@ -738,11 +738,6 @@ pub fn export_identity(
         .get(name)
         .context(NoSuchIdentityToExportSnafu { name })?;
 
-    // Cannot export anonymous
-    if matches!(spec, IdentitySpec::Anonymous) {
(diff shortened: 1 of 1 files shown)


crates/icp/src/identity/key.rs
```diff
@@ -676,3 +676,118 @@ fn make_pkcs5_encrypted_pem(doc: &SecretDocument, password: &str) -> Zeroizing<S
...
+    let identity_list = IdentityList::load_from(dirs)?;
+
+    // Check the identity exists
+    let spec = identity_list
+        .identities
+        .get(name)
+        .context(NoSuchIdentityToExportSnafu { name })?;
+
+    // Cannot export anonymous
+    if matches!(spec, IdentitySpec::Anonymous) {
```
adamspofford: This can be merged with the match statement below

PR #320, opened 2026-01-29
feat: publish via npm

viviveevee:
New workflow to make npm releases from the repo itself

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3153,6 +3153,7 @@ dependencies = [
  "ic-ed25519 0.5.0",
  "ic-identity-hsm",
  "ic-ledger-types",
+ "ic-management-canister-types 0.4.1",
  "ic-utils",
  "icp-canister-interfaces",
  "icrc-ledger-types",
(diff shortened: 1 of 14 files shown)


.github/workflows/release-npm.yml
```diff
@@ -0,0 +1,56 @@
+name: Publish to npm
+
+on:
+  release:
+    types: [published]
```
viviveevee: Make manual instead for now

.github/workflows/release-npm.yml
```diff
@@ -0,0 +1,56 @@
...
+
+jobs:
+  publish-npm:
```
viviveevee, line 21: Todo: check how cargo dist can do it

viviveevee: Conclusion: not too well, we'll stick with this separate workflow. Issues with trusted publishing, beta releases, and the overall package structure is worse DX

npm/icp-cli-darwin-arm64/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "arm64"
+  ],
+  "main": "index.js",
+  "files": [
+    "bin/icp"
+  ],
+  "keywords": [
+    "icp",
+    "internet-computer"
+  ]
```
adamspofford, line 25: Missing "cli"

PR #322, opened 2026-01-29
feat: HSM identity support

adamspofford-dfinity:


crates/icp-cli/src/commands/identity/mod.rs
raymondk, line 38: Add a note to the help that for hsm users should use `icp link`

crates/icp-cli/src/commands/identity/link/hsm.rs
```diff
@@ -0,0 +1,75 @@
...
+use icp::{context::Context, identity::key::link_hsm_identity, prelude::*};
+use snafu::{ResultExt, Snafu};
+
+#[derive(Debug, Args)]
+pub(crate) struct LinkArgs {
+    /// Name for the linked identity
+    name: String,
+
+    /// Path to the PKCS#11 module (shared library) for the HSM
+    #[arg(long)]
```
raymondk, line 12: We said we would move this to a subcommand to leave room for ledge and ii later.
`icp link hsm ...`

PR #325, opened 2026-01-30
feat: versioned release docs deployment

marc0olo:
## Summary
- **Versioned docs deployment**: Tags (`v*`) deploy to `/major.minor/`, docs branches (`docs/v*`) update existing versions
- **Version switcher**: Dropdown in header fetches `versions.json` at runtime, shows current version with checkmark, marks latest
- **Main branch preview**: Always available at `/main/` for pre-release docs or upcoming changes
- **Fork-friendly**: Configure `PUBLIC_SITE` and `PUBLIC_BASE_PREFIX` env vars in workflow—no hardcoded paths
- **Local testing**: `./docs-site/test-version-switcher.sh` builds multiple versions and serves them locally


docs-site/src/components/VersionSwitcher.astro
```diff
@@ -0,0 +1,367 @@
...
+
+// Extract version from BASE_URL (e.g., "/icp-cli/0.1/" -> "0.1")
+const versionMatch = baseUrl.match(/\/(\d+\.\d+)\//);
+const currentVersion = versionMatch ? versionMatch[1] : 'unknown';
+---
+
+<div class="version-switcher" data-base-prefix={basePrefix}>
+  {isDev ? (
+    <!-- Development: Show dev badge -->
+    <div class="version-badge">
```
raymondk, line 21: Seems like these should be using `<VerisonBadge />`  or that `VersionBadge.astro` should be removed.

.github/workflows/docs.yml
```diff
@@ -53,21 +77,154 @@ jobs:
...
+
+      - name: Prepare root files
+        run: |
+          mkdir -p root
+
+          # Copy versions.json from repo
+          cp docs-site/versions.json root/versions.json
+
+          # Extract latest version (first in list) from versions.json
+          LATEST_VERSION=$(jq -r '.versions[0].version // empty' docs-site/versions.json)
```
raymondk: It's a bit odd that `versions.json` has a `latest=true` according to the docs but we don't seem to be using it here.

raymondk:
```suggestion
          # Extract the latest version from versions.json (if multiple have the flag set, take the first one)
          LATEST_VERSION=$( jq  -r "[.versions[] | select(.latest == true)] | first  | .version // empty"
 docs-site/versions.json)
```

docs-site/src/components/VersionSwitcher.astro
marc0olo: yeah, that was a relict from previous iteration and `VersionBadge.astro` can be removed.

PR #330, opened 2026-02-02
feat: Implement snapshot uploading/downloading

adamspofford-dfinity:


crates/icp-cli/src/operations/snapshot_transfer.rs
```diff
@@ -0,0 +1,880 @@
...
+    Ok(next_report_offset)
+}
+
+/// Upload a single WASM chunk.
+pub async fn upload_wasm_chunk(
+    agent: &Agent,
+    canister_id: Principal,
+    snapshot_id: &[u8],
+    chunk_hash: &[u8],
+    paths: LWrite<&SnapshotPaths>,
```
lwshang, line 756: I initially thought `paths` could use `LRead` instead. However, since this operation runs within a `with_write` closure, and the lock system doesn't support downgrading—i.e., treating a write lock as sufficient when only a read lock is needed—we need to keep it as is.

lwshang (approved):

PR #347, opened 2026-02-06
feat: add `bitcoind-addr` and `dogecoind-addr` to managed launcher config

marc0olo:
## Summary

- Adds `bitcoind-addr` and `dogecoind-addr` fields to the managed network Launcher variant, allowing native launcher mode to connect to Bitcoin/Dogecoin P2P nodes directly from `icp.yaml` configuration
- Updates the Windows auto-Docker fallback (`transform_native_launcher_to_container`) to automatically translate localhost addresses (`127.0.0.1`, `localhost`, `::1`) to `host.docker.internal` when converting launcher config to Docker options — and adds `host.docker.internal:host-gateway` to the container's `extra_hosts` so the hostname resolves on Linux Docker Engine (relevant for manual WSL2 dockerd setups where Docker Desktop isn't handling DNS)
Updates docs, JSON schemas, and adds tests for the new fields and the Windows auto-Docker address translation

**Note**:
- Requires changes in the icp-cli-network-launcher we recently introduced for bitcoin and dogecoin to work properly, see https://github.com/dfinity/icp-cli-network-launcher/pull/22

### DIFF
--- a/crates/icp/src/manifest/network.rs
+++ b/crates/icp/src/manifest/network.rs
@@ -56,6 +56,8 @@ pub enum ManagedMode {
         status_dir: Option<String>,
         /// Bind mounts to add to the container in the format relative_host_path:container_path[:options]
         mounts: Option<Vec<String>>,
+        /// Extra hosts entries for Docker networking (e.g. "host.docker.internal:host-gateway")
+        extra_hosts: Option<Vec<String>>,
     },
     Launcher {
         /// HTTP gateway configuration
@@ -363,6 +365,41 @@ mod tests {
                         bitcoind_addr: None,
                         dogecoind_addr: Some(vec!["127.0.0.1:22556".to_string()]),
                         version: None,
+                    })
+                })
+            },
+        );
+    }
+
+    #[test]
+    fn managed_docker_network_with_extra_hosts() {
+        assert_eq!(
+            validate_network_yaml(indoc! {r#"
+                    name: my-network
+                    mode: managed
+                    image: ghcr.io/dfinity/icp-cli-network-launcher
+                    port-mapping:
+                      - "8000:4943"
+                    extra-hosts:
+                      - "host.docker.internal:host-gateway"
+                "#}),
+            NetworkManifest {
+                name: "my-network".to_string(),
+                configuration: Mode::Managed(Managed {
+                    mode: Box::new(ManagedMode::Image {
+                        image: "ghcr.io/dfinity/icp-cli-network-launcher".to_string(),
+                        port_mapping: vec!["8000:4943".to_string()],
+                        rm_on_exit: None,
+                        args: None,
+                        entrypoint: None,
+                        environment: None,
+                        volumes: None,
+                        platform: None,
+                        user: None,
+                        shm_size: None,
+                        status_dir: None,
+                        mounts: None,
+                        extra_hosts: Some(vec!["host.docker.internal:host-gateway".to_string()]),
                     })
                 })
             },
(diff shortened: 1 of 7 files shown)


crates/icp/src/manifest/network.rs
```diff
@@ -55,6 +55,18 @@ pub enum ManagedMode {
         status_dir: Option<String>,
         /// Bind mounts to add to the container in the format relative_host_path:container_path[:options]
         mounts: Option<Vec<String>>,
+        /// Artificial delay to add to every update call
```
raymondk: I don't think this is the right approach for docker:
- We cannot count on the docker image actually using the network launcher. It could be using anything.
- It's up to the author of the image to make it compatible with the cli: Has the cycles ledger, mints ICP to the anonymous principal and write a compatible status file to the status directory.
- If the image author wants to make their image flexible, the can support turning some features on through environment variables, or by modifying the entrypoint.

crates/icp/src/network/managed/docker.rs
```diff
@@ -187,6 +190,52 @@ pub enum ManagedImageConversionError {
...
+    }
+}
+
+/// Returns extra_hosts entries needed for Docker to resolve `host.docker.internal`.
+/// On Linux, Docker Engine does not provide `host.docker.internal` by default,
+/// so we add `host.docker.internal:host-gateway` when any addresses reference localhost.
+pub(super) fn docker_extra_hosts_for_addrs(addrs: &[String]) -> Vec<String> {
+    let needs_host_gateway = addrs.iter().any(|addr| {
+        addr.rsplit_once(':')
+            .map(|(host, _)| matches!(host, "127.0.0.1" | "localhost" | "::1"))
```
adamspofford: Also needs 0.0.0.0

crates/icp/src/network/managed/docker.rs
```diff
@@ -187,6 +190,52 @@ pub enum ManagedImageConversionError {
     WslPathConvert { source: WslPathConversionError },
 }
 
+/// Translates a host:port address for use inside a Docker container.
+/// Replaces `127.0.0.1`, `localhost`, and `::1` with `host.docker.internal`
+/// so the container can reach services running on the host machine.
+pub(super) fn translate_addr_for_docker(addr: &str) -> String {
+    if let Some((host, port)) = addr.rsplit_once(':') {
+        let translated = match host {
+            "127.0.0.1" | "localhost" | "::1" => "host.docker.internal",
```
adamspofford: Also needs 0.0.0.0

docs/guides/containerized-networks.md
```diff
@@ -158,6 +158,27 @@ networks:
...
+      - "8000:4943"
+    args:
+      - "--ii"
+```
+
+The `args` field passes values directly to the container entrypoint with no processing. The Docker image determines what arguments it accepts — see the image's documentation for available options.
+
+**Comparison with native launcher mode:** When using native managed networks (without `image`), settings like `bitcoind-addr`, `ii`, `nns`, and `subnets` are configured as top-level YAML fields. In Docker image mode, these are passed via `args` instead, since the image could be any Docker image — not necessarily the official network launcher.
+
+> **Docker networking note:** When referencing services running on the host machine from inside a container (e.g., a local Bitcoin node), use `host.docker.internal` instead of `127.0.0.1` or `localhost`. Inside a container, `127.0.0.1` refers to the container's own loopback, not the host. For example: `--bitcoind-addr=host.docker.internal:18444`. Docker Desktop (macOS/Windows) resolves `host.docker.internal` automatically. On Linux Docker Engine, you may need to pass `--add-host=host.docker.internal:host-gateway` or equivalent to ensure it resolves.
```
adamspofford: This implies that you can pass Docker CLI flags, which you can't. `extra-hosts` should instead be surfaced in the docker network options YAML.

PR #356, opened 2026-02-12
feat: add --json output flag to canister call

marc0olo:
## Summary

- Add `--json` flag to `icp canister call` that outputs the response as JSON instead of Candid text format, using the `idl2json` crate (same approach as dfx)
- Consistent with the existing `--json` flag on `canister status` and `network status`
- Addresses user request for JSON output to support scripting and automation workflows


marc0olo:
converted this into a draft because that might also be relevant for input formats and we can have different formats. 

currently we always assume that Candid is returned.

PR #360, opened 2026-02-12
feat: Add more info to the network status

raymondk:


marc0olo:
the `icp project show` still prints `null` for the root-key, do we also want to fix this along with this PR or separately?

```
ic:
    name: ic
    network:
      name: ic
      configuration:
        mode: connected
        url: https://icp-api.io
        root-key: null
```

crates/icp-cli/src/commands/network/status.rs
```diff
@@ -54,34 +59,54 @@ pub(crate) async fn exec(ctx: &Context, args: &StatusArgs) -> Result<(), anyhow:
...
-    // Build status structure
-    let status = NetworkStatus {
-        port: descriptor.gateway.port,
-        root_key: hex::encode(&descriptor.root_key),
-        candid_ui_principal: descriptor.candid_ui_canister_id.map(|p| p.to_string()),
-        proxy_canister_principal: descriptor.proxy_canister_id.map(|p| p.to_string()),
+            // Build status structure
+            NetworkStatus {
+                managed: true,
+                url: network_access.url.to_string(),
```
marc0olo: @adamspofford-dfinity was mentioning we might need both, `api_url` and `gateway_url`.

the todo in the vite config of the `icp-frontend-environment-variables` example was prepared for `api_url`.

I just wanted to bring this up again before we merge. I do not have a strong opinion on this. either way, I think you can also directly update the vite config accordingly to use the dynamic value of the CLI to set the proxyTarget.

crates/icp-cli/src/commands/network/status.rs
```diff
@@ -54,34 +59,54 @@ pub(crate) async fn exec(ctx: &Context, args: &StatusArgs) -> Result<(), anyhow:
...
+                    .port_or_known_default()
+                    .expect("Didn't find a valid default port for url"),
+                root_key: hex::encode(network_access.root_key.unwrap_or(IC_ROOT_KEY.to_vec())),
+                candid_ui_principal: descriptor.candid_ui_canister_id.map(|p| p.to_string()),
+                proxy_canister_principal: descriptor.proxy_canister_id.map(|p| p.to_string()),
+            }
+        }
+        Configuration::Connected { connected: _ } => NetworkStatus {
+            managed: false,
+            url: network_access.url.to_string(),
```
marc0olo: see comment above

crates/icp-cli/src/commands/network/status.rs
adamspofford: To clarify:
- a gateway URL can substitute for an API URL, gateway is a strict increase in capability
- the pocket-ic URL is a gateway URL, so it can be used for both
- the URL we use for mainnet, icp-api.io, is an API url; the corresponding gateway url is icp0.io (and we shouldn't switch to icp0.io). So when reporting mainnet, the two values will be different.

raymondk: You will always need the api endpoint to make calls - that's where agents (including frontend agents) will make calls to usually.
The http gateway you need to navigate to your frontend in your browser but it's difficult to guess that for a connected network unless that's the URL you've been given.

adamspofford: Right. We should decide whether we assume connected network URLs are gateways or API URLs (I think gateway URLs) and report it accordingly (if API urls, then gateway_url is nullable). But we should still note the difference, because it is there for mainnet. (The reason we need to keep the mainnet divergence is because some canisters may be blocked from icp0.io, but all canisters are accessible under icp-api.io.)

adamspofford-dfinity (approved):

PR #361, opened 2026-02-13
fix: Consistent caching and checksumming

adamspofford-dfinity:
Caches URL-based recipes and ensures the checksum is always verified.


crates/icp/src/canister/recipe/handlebars.rs
```diff
@@ -167,6 +187,10 @@ impl Handlebars {
             }
         };
 
+        if let Some(sha256) = &recipe.sha256 {
+            verify_checksum(tmpl.as_bytes(), sha256)?;
+        }
```
lwshang: If the downloaded content doesn't match the expected checksum, we've already cached the "wrong" content. On subsequent runs, we'll retrieve this cached content and fail verification again.

Should we verify the checksum before caching for remote URLs? Or is the current behavior intentional (cache everything, verify later)?

PR #362, opened 2026-02-13
feat: Separate network.url into api-url and gateway-url

adamspofford-dfinity:


crates/icp-cli/src/commands/deploy/mod.rs
```diff
@@ -392,72 +392,81 @@ async fn print_canister_urls(
...
-                    let scheme = if base_url.starts_with("https://") {
-                        "https"
+        if let Some(base_url) = &base_url {
+            // Check if canister has http_request
+            let has_http = has_http_request(&agent, canister_id).await;
+
+            if has_http {
+                // For canisters with http_request, show the direct canister URL
+                let canister_url =
+                    if base_url.contains("localhost") || base_url.contains("127.0.0.1") {
```
raymondk: Out of curiosity for these places where we need to do URL operations - why not use the url crate?

crates/icp/src/manifest/network.rs
```diff
@@ -88,14 +88,31 @@ impl Default for ManagedMode {
...
 
+#[derive(Clone, Debug, Deserialize, PartialEq, JsonSchema)]
+#[serde(untagged, rename_all_fields = "kebab-case")]
+pub enum Endpoints {
+    Explicit {
+        /// The URL of the gateway endpoint. Should support prefixing canister IDs as subdomains.
+        gateway_url: Option<String>,
+        /// The URL of the API endpoint. Should support the standard API routes (e.g. /api/v3).
+        /// If no gateway endpoint is provided, canister URLs will not be printed in deploy operations.
+        api_url: String,
```
raymondk:
```suggestion
        /// The URL of the HTTP gateway. `icp-cli` assumes that it supports prefixing canister IDs as subdomains.
        /// If no http gateway is provided, canister URLs will not be printed in deploy operations.
        http_gateway_url: Option<String>,
        
        /// The URL of api gateway to use. Should support the standard API routes (e.g. /api/v3).
        api_gateway_url: String,
```

In our documentation we refer to those as "http gateway" and "api gateway", see: https://docs.internetcomputer.org/references/http-gateway-protocol-spec#api-gateway-resolution

crates/icp/src/manifest/network.rs
```diff
@@ -88,14 +88,31 @@ impl Default for ManagedMode {
...
+    },
+    Implicit {
+        /// The URL this network can be reached at.
+        /// Assumed to be both gateway (canister-id.domain.com) and API (domain.com/api/v3).
+        url: String,
```
raymondk:
```suggestion
        /// The URL this network can be reached at.
        /// Assumed to be the url of the http gateway (canister-id.domain.com) and the api gateway (domain.com/api/v3).
        url: String,
```

PR #366, opened 2026-02-17
feat: icp canister logs

viviveevee:
Add `icp canister logs` with the options `--follow` for continuous polling and `--interval n` to poll every `n` seconds


crates/icp-cli/src/commands/canister/logs.rs
```diff
@@ -0,0 +1,220 @@
...
+use crate::commands::args;
+
+/// Fetch and display canister logs
+#[derive(Debug, Args)]
+pub(crate) struct LogsArgs {
+    #[command(flatten)]
+    pub(crate) cmd_args: args::CanisterCommandArgs,
+
+    /// Continuously fetch and display new logs until interrupted with Ctrl+C
+    #[arg(long)]
```
raymondk: can we add a short also? so we can do `-f`

PR #374, opened 2026-02-18
docs: add canister snapshots and migration guides, enhance existing docs

marc0olo:
## Summary

- Add new **Canister Snapshots** guide covering create, download, upload, restore, and delete workflows with pre-upgrade backup and state transfer examples
- Add new **Canister Migration** guide with two approaches: snapshot-only transfer (new ID) and full migration via `migrate-id` (preserved ID), including troubleshooting and interruption handling
- Enhance **Tokens and Cycles** guide with subaccount documentation (balance checks, transfers, receiver address formats, account identifiers)
- Enhance **Local Development** guide with `--query` flag, proxy canister usage, and `network status --json` example output
- Enhance **Containerized Networks** guide with launcher version pinning section
- Improve **Environments** concept doc with `api-url` / `http-gateway-url` YAML example
- Replace outdated "delete and redeploy" advice in **Deploying to Specific Subnets** with links to the new migration guides
- Wire up new guides in sidebar navigation, index pages, and cross-links

### DIFF
--- a/docs/guides/tokens-and-cycles.md
+++ b/docs/guides/tokens-and-cycles.md
@@ -289,7 +289,7 @@ icp cycles transfer 1T rrkah-fqaaa-aaaaa-aaaaq-cai --to-subaccount 1 -n ic
 
 Alternatively, you can use the ICRC-1 account format, which embeds the subaccount directly in the address. This format appends a CRC32 checksum to the principal, followed by the subaccount hex after a `.` separator. For example, `2vxsx-fae-22yutvy.1` is the anonymous principal (`2vxsx-fae`) with checksum `22yutvy` and subaccount `1`.
 
-Use `icp identity account-id --of-subaccount <hex>` to get a correctly formatted ICRC-1 address rather than constructing it manually.
+Use `icp identity account-id --of-subaccount <hex>` to get both the ICP ledger account ID and the ICRC-1 account format for a subaccount, rather than constructing them manually.
 
 ### Receiver Address Formats
 


docs/guides/canister-migration.md
```diff
@@ -0,0 +1,348 @@
...
+
+**Snapshot transfer** — When you can accept a new canister ID. Create a new canister on the desired subnet, transfer state via [snapshots](canister-snapshots.md), and switch over. See [Migrating Without Preserving the Canister ID](#migrating-without-preserving-the-canister-id) below.
+
+**Full migration** — When the canister ID must be preserved. This applies when the canister ID is load-bearing:
+
+- **Threshold signatures (tECDSA / tSchnorr):** The IC derives signing keys by cryptographically binding them to the calling canister's principal. A canister's derived keys — and any addresses or public keys derived from them — are permanently tied to its ID. Losing the ID means losing access to those keys and any assets they control, whether those are addresses on other blockchains (Bitcoin, Ethereum, etc.) or ICP principals controlled by the canister.
+- **VetKeys:** VetKey derivation similarly includes the canister's principal. A new ID produces entirely different decryption keys, making previously encrypted data inaccessible.
+- **External references:** Other canisters, frontends, or off-chain systems that reference the canister by ID would break.
```
viviveevee: Maybe worth adding to external references: Internet Identity login through a domain that contains the canister id

docs/guides/canister-migration.md
```diff
@@ -0,0 +1,348 @@
...
+icp canister status my-canister -e ic
+
+# Example: copy non-default settings
+icp canister update-settings <target-id> \
+  --compute-allocation 10 \
+  --freezing-threshold 604800 \
+  -n ic
+```
+
+Settings you may need to copy: `--compute-allocation`, `--memory-allocation`, `--freezing-threshold`, `--reserved-cycles-limit`, `--wasm-memory-limit`, `--wasm-memory-threshold`, `--log-visibility`.
```
viviveevee: I would not reference the full list here. It will likely grow soon and we will forget to update this. Maybe reference `icp canister update-settings --help` to list possible settings?

docs/guides/canister-migration.md
```diff
@@ -0,0 +1,348 @@
...
+```bash
+# Example: copy non-default settings to the target canister
+icp canister update-settings <target-id> \
+  --compute-allocation 10 \
+  --freezing-threshold 604800 \
+  --wasm-memory-limit 2GiB \
+  -n ic
+```
+
+Settings you may need to copy: `--compute-allocation`, `--memory-allocation`, `--freezing-threshold`, `--reserved-cycles-limit`, `--wasm-memory-limit`, `--wasm-memory-threshold`, `--log-visibility`. You do **not** need to copy controllers — those are restored automatically.
```
viviveevee: same here

docs/guides/canister-migration.md
```diff
@@ -0,0 +1,348 @@
...
+```
+
+Verify the canister is on the expected subnet by querying the NNS Registry canister:
+
+```bash
+icp canister call rwlgt-iiaaa-aaaaa-aaaaa-cai get_subnet_for_canister \
+  '(record { "principal" = opt principal "<canister-id>" })' --query -n ic
+```
+
+You can also verify on the [ICP Dashboard](https://dashboard.internetcomputer.org) — search for your canister ID to confirm its subnet.
```
viviveevee: Can you check with the dashboard team how quickly this updates? If people go quick I'm pretty sure they will be served stale data

docs/guides/tokens-and-cycles.md
```diff
@@ -246,6 +246,72 @@ icp token transfer 1 <RECEIVER> --identity my-wallet -n ic
...
+# Send ICP to subaccount 1 of the receiver
+icp token transfer 1 rrkah-fqaaa-aaaaa-aaaaq-cai --to-subaccount 1 -n ic
+
+# Send cycles to subaccount 1 of the receiver
+icp cycles transfer 1T rrkah-fqaaa-aaaaa-aaaaq-cai --to-subaccount 1 -n ic
+```
+
+Alternatively, you can use the ICRC-1 account format, which embeds the subaccount directly in the address. This format appends a CRC32 checksum to the principal, followed by the subaccount hex after a `.` separator. For example, `2vxsx-fae-22yutvy.1` is the anonymous principal (`2vxsx-fae`) with checksum `22yutvy` and subaccount `1`.
+
+Use `icp identity account-id --of-subaccount <hex>` to get a correctly formatted ICRC-1 address rather than constructing it manually.
```
viviveevee: This is wrong, the command produces 'legacy account ids' that are used by e.g. exchanges. ICP only, not supported by ICRC tokens

docs/guides/tokens-and-cycles.md
```diff
@@ -246,6 +246,72 @@ icp token transfer 1 <RECEIVER> --identity my-wallet -n ic
...
+
+### Receiver Address Formats
+
+Transfer commands accept multiple receiver formats:
+
+| Format | Example | Description |
+|--------|---------|-------------|
+| Principal | `rrkah-fqaaa-aaaaa-aaaaq-cai` | Sends to the principal's default subaccount |
+| ICRC-1 account | `2vxsx-fae-22yutvy.1` | Principal + CRC32 checksum + subaccount hex |
+| ICP ledger account ID | `64-character hex string` | 32-byte account identifier (used by exchanges and the NNS) |
```
viviveevee, line 302: here it's right

docs/guides/canister-migration.md
marc0olo: good point, agree!

docs/guides/canister-migration.md
marc0olo: good point, I think I will remove this here. we're focusing on icp-cli mainly. so I do not think we have to talk about this option in the cli docs.

I am quite sure it wouldn't be updated immediately. the list of controllers is also not immediately updated when changing it.

docs/guides/tokens-and-cycles.md
marc0olo: actually it returns both, I just updated it. and generally we should avoid calling the acount identifier `legacy` as it is still valid and sometimes the only option provided by exchanges.

docs/guides/canister-migration.md
marc0olo: I adjusted info and fixed the command which was also wrong 😅

viviveevee (approved):

PR #375, opened 2026-02-19
feat: support cycles amount suffixes in .yaml files

viviveevee:
Support cycles amount suffixes in `.yaml` files


crates/icp/src/canister/mod.rs
```diff
@@ -198,6 +198,34 @@ impl From<LogVisibility> for icp_canister_interfaces::cycles_ledger::LogVisibili
...
+/// Deserializes reserved_cycles_limit from either a number or a string (e.g. "4t") in YAML.
+/// OneOf is needed because YAML can give us a number token or a string token; we then
+/// always parse via parse_cycles_amount so bare numbers and suffix strings are handled in one place.
+fn deserialize_reserved_cycles_limit<'de, D>(d: D) -> Result<Option<u64>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    #[derive(Deserialize)]
+    #[serde(untagged)]
+    enum OneOf {
```
adamspofford: Making this internal means that the schema can't see it. Can this function instead be formatted as a pub wrapper type which impls Deserialize and JsonSchema?

crates/icp/src/canister/mod.rs
```diff
@@ -198,6 +198,34 @@ impl From<LogVisibility> for icp_canister_interfaces::cycles_ledger::LogVisibili
...
+        Number(u64),
+        Str(String),
+    }
+
+    let opt: Option<OneOf> = Option::deserialize(d)?;
+    opt.map(|v| {
+        let s = match &v {
+            OneOf::Number(n) => n.to_string(),
```
adamspofford: nit: this doesn't need to be roundtripped through conversion to/from a string

crates/icp/src/canister/mod.rs
viviveevee: It's now a pub wrapper type. How do you like the new version? Still gives a clunky impression to me, but I'd like to validate the amount the first time the type is constructed. Is there a nicer way to do it?

adamspofford: Looks good as is to me, but if you did want to make it cleaner, you could implement JsonSchema directly on a u64 newtype, since the enum structure is I *think* only there for schemars(untagged).

adamspofford-dfinity (approved):

PR #383, opened 2026-02-23
feat: support suffixes for memory amounts

viviveevee:
Support `kb`, `kib`, `mb`, `mib`, `gb`, `gib` suffixes in `.yaml` files and CLI arguments when specifying memory amounts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@
 * feat: Support `k`, `m`, `b`, `t` suffixes in `.yaml` files when specifying cycles amounts
 * feat: Support `kb`, `kib`, `mb`, `mib`, `gb`, `gib` suffixes in `.yaml` files and CLI arguments when specifying memory amounts
 * feat: Add an optional root-key argument to canister commands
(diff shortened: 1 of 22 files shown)


crates/icp-cli/src/commands/canister/settings/update.rs
```diff
@@ -97,8 +96,9 @@ pub(crate) struct UpdateArgs {
     #[arg(long, value_parser = compute_allocation_parser)]
     compute_allocation: Option<u8>,
 
-    #[arg(long, value_parser = memory_parser)]
-    memory_allocation: Option<Byte>,
+    /// Memory allocation. Supports suffixes: kb, kib, mb, mib, gb, gib (e.g. "4gib" or "2.5kb").
```
raymondk: For all these parameters should we say what unit without suffix is? eg
```suggestion
    /// Memory allocation in bytes. Supports suffixes: kb, kib, mb, mib, gb, gib (e.g. "4gib" or "2.5kb").
```

PR #385, opened 2026-02-23
feat: add the ability to create an unrecorded canister on a network

raymondk:


crates/icp-cli/src/commands/canister/create.rs
```diff
@@ -97,13 +120,88 @@ impl CreateArgs {
...
+        }
+    }
 }
 
 // Creates canister(s) by asking the cycles ledger to create them.
 // The cycles ledger will take cycles out of the user's account, and attaches them to a call to CMC::create_canister.
 // The CMC will then pick a subnet according to the user's preferences and permissions, and create a canister on that subnet.
 pub(crate) async fn exec(ctx: &Context, args: &CreateArgs) -> Result<(), anyhow::Error> {
-    let selections = args.cmd_args.selections();
+    if let Ok(selections) =
```
adamspofford: The detached flag should be used here, not whether a fallible conversion failed. Currently it is an unused field.

crates/icp-cli/src/commands/canister/create.rs
```diff
@@ -62,6 +74,16 @@ pub(crate) struct CreateArgs {
     /// The subnet to create canisters on.
     #[arg(long)]
     pub(crate) subnet: Option<Principal>,
+
+    /// Only print the canister id
+    #[arg(short, long)]
+    pub id_only: bool,
```
adamspofford: This sounds like it is a copy of `--quiet`.

crates/icp-cli/src/commands/canister/create.rs
```diff
@@ -97,13 +120,88 @@ impl CreateArgs {
...
+            .write_line(&format!("Created canister with ID {id}"));
+    }
+
+    Ok(())
+}
+
+// Creates canister(s) by asking the cycles ledger to create them.
+// The cycles ledger will take cycles out of the user's account, and attaches them to a call to CMC::create_canister.
+// The CMC will then pick a subnet according to the user's preferences and permissions, and create a canister on that subnet.
+async fn create_project_canister(
```
adamspofford: Duplicate of the comment block above - should one be deleted?

crates/icp-cli/src/commands/canister/create.rs
raymondk: right - we have an `--id-only` on canister status also...

crates/icp-cli/src/commands/canister/create.rs
raymondk: you're right - with the explicit flag I can remove a bunch of stuff.

crates/icp-cli/src/commands/canister/create.rs
```diff
@@ -190,15 +185,14 @@ async fn create_canister(
...
-    selections: CommandSelections,
-) -> Result<(), anyhow::Error> {
-    let canister = match selections.canister {
+// Attempts to create a canister and record it in the project metadata
+async fn create_project_canister(ctx: &Context, args: &CreateArgs) -> Result<(), anyhow::Error> {
+    let selections = args.cmd_args.selections();
+
+    let canister = match selections
+        .canister
+        .expect("Canister must be Some() when --detached is not used")
```
adamspofford, line 203: This panic is visible to the user when the command is run without arguments

PR #386, opened 2026-02-24
feat: Anonymous usage telemetry

lwshang:
Collects command name, arguments, duration, and outcome
  * Enabled by default; opt out with `icp settings telemetry false`, `DO_NOT_TRACK=1`, or `ICP_TELEMETRY_DISABLED=1`
  * Automatically disabled in CI environments (`CI` env var set)
  * `icp settings telemetry` to view or change the current setting

### DIFF
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -12,6 +12,7 @@ Each command invocation produces a single telemetry record with the following fi
 | `platform` | `macos`, `linux`, `windows`, `wsl` | Platform distribution |
 | `arch` | `aarch64`, `x86_64` | Architecture distribution |
 | `version` | `0.1.0` | Identify version adoption |
+| `date` | `2026-02-24` | UTC date of the event, for timeseries analysis |
 | `command` | `build`, `deploy`, `canister status` | Feature usage |
 | `arguments` | (see below) | Argument usage |
 | `autocontainerize` | `true` / `false` (optional) | Track adoption of autocontainerize setting |
--- a/crates/icp-cli/src/telemetry.rs
+++ b/crates/icp-cli/src/telemetry.rs
@@ -15,6 +15,7 @@ use icp::settings::Settings;
 use icp::telemetry_data::{IdentityStorageType, NetworkType, TelemetryData};
 use rand::Rng as _;
 use serde::{Deserialize, Serialize};
+use time::OffsetDateTime;
 
 use crate::version::icp_cli_version_str;
 
@@ -77,6 +78,9 @@ pub(crate) struct TelemetryRecord {
 
     pub version: String,
 
+    /// UTC date of the event in `YYYY-MM-DD` format, for timeseries analysis.
+    pub date: String,
+
     // --- About command itself
     pub command: String,
     pub arguments: Vec<Argument>,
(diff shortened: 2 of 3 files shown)


docs/telemetry.md
```diff
@@ -0,0 +1,184 @@
...
+| `platform` | `macos`, `linux`, `windows`, `wsl` | Platform distribution |
+| `arch` | `aarch64`, `x86_64` | Architecture distribution |
+| `version` | `0.1.0` | Identify version adoption |
+| `command` | `build`, `deploy`, `canister status` | Feature usage |
+| `arguments` | (see below) | Argument usage |
+| `autocontainerize` | `true` / `false` (optional) | Track adoption of autocontainerize setting |
+| `success` | `true` / `false` | Error rates |
+| `duration_ms` | `4230` | Performance insights |
+| `identity_type` | `pem-file`, `keyring` (optional) | Identity storage method distribution |
+| `network_type` | `local`, `ic`, `playground` (optional) | Network target distribution |
```
viviveevee:
```suggestion
| `network_type` | `managed`, `connected` | Network target distribution |
```
Isn't this the network type?

docs/telemetry.md
```diff
@@ -0,0 +1,184 @@
...
+| `num_canisters` | `3` (optional) | Project size distribution |
```
viviveevee: Are these canisters in the project or canisters targeted by the command?

docs/telemetry.md
```diff
@@ -0,0 +1,184 @@
...
+The `machine_id` is a random UUID generated on first run and stored locally. It is used solely to count unique installations and is not linked to any user identity.
+
+Additional fields may be introduced in future versions. This page will be updated accordingly. The same privacy principles apply: no personally identifiable information, no project data.
+
+## What is not collected
+
+- IP addresses, usernames, or any personally identifiable information
+- Project names, file paths, or file contents
+- Canister IDs, wallet addresses, or cycle balances
+- Free-form argument values (only values from constrained `possible_values` sets are recorded)
```
viviveevee, line 52: Could recipe names be considered free-form values? If someone creates their own recipes it could contain some quite descriptive strings

crates/icp-cli/src/main.rs
```diff
@@ -92,6 +91,18 @@ struct Cli {
 
 #[tokio::main]
 async fn main() -> Result<(), Error> {
+    // -----------------------------------------------------------------------
+    // Hidden telemetry-send-batch mode: spawned as a detached child process.
```
viviveevee:
```suggestion
    // Background telemetry-send-batch mode: spawned as a detached child process.
```
Sounds less like secretly sneaking away with data

crates/icp-cli/src/main.rs
```diff
@@ -174,142 +165,172 @@ async fn main() -> Result<(), Error> {
...
+    let telemetry_session = telemetry::setup(&ctx, &raw_args, &Cli::command()).await;
+
+    // -----------------------------------------------------------------------
+    // Command dispatch
+    // -----------------------------------------------------------------------
+    let result = dispatch(&ctx, command, trace_span).await;
+
+    // -----------------------------------------------------------------------
+    // Record the telemetry event
+    // -----------------------------------------------------------------------
```
viviveevee: There are lots of these headings everywhere. Does it make sense to keep them? To me they look like they add minimal value

docs/telemetry.md
lwshang: Only "registry" recipes are recorded — recipes specified as local file paths or URLs are excluded. Registry recipes are considered public knowledge, making this field analogous to the "canister type" field in dfx.

PR #390, opened 2026-02-24
feat: candid interface compatibility check

adamspofford-dfinity:


crates/icp-cli/src/operations/install.rs
```diff
@@ -38,37 +53,128 @@ struct InstallFailure {
...
+            })
+        }
+        "install" => Ok(CanisterInstallMode::Install),
+        "reinstall" => Ok(CanisterInstallMode::Reinstall),
+        "upgrade" => Ok(CanisterInstallMode::Upgrade(None)),
+        _ => panic!("invalid install mode: {mode}"),
+    }
+}
+
+pub(crate) async fn install_canister(
```
raymondk, line 77: Why not have separate operations:
- `install_canister` will install a canister
- `check_candid_compatibility` will only check the compatiblity

Them the install command will do:
- `check compatiblity`
- 'install canister`

And the deploy command will:
- `check_compatiblity_many`
- 'install_many`

That would be following the same patterns he we have already: https://github.com/dfinity/icp-cli/blob/main/crates/icp-cli/src/commands/deploy/mod.rs#L19-L21

raymondk (approved):

PR #400, opened 2026-03-03
feat: Implement friendly domains

adamspofford-dfinity:


crates/icp/src/context/mod.rs
```diff
@@ -500,6 +500,65 @@ impl Context {
...
+            return;
+        };
+        let Some(status_dir) = &desc.status_dir else {
+            return;
+        };
+        let gateway_url = Url::parse(&format!(
+            "http://{}:{}",
+            desc.gateway.host, desc.gateway.port
+        ))
+        .unwrap();
```
lwshang: The `.unwrap()` here stands out given that everything else in this function silently returns on failure. Since `desc.gateway.host` comes from a persisted descriptor, a malformed value is possible in theory.

Two options worth considering:

**Option A — `expect()`, abort with a clear message:**
```rust
let gateway_url = Url::parse(&format!("http://{}:{}", desc.gateway.host, desc.gateway.port))
    .expect("network descriptor contains invalid gateway host");
```
Keeps the panic but makes the invariant explicit to future readers.

**Option B — best-effort early return with a warning log:**
```rust
let gateway_url_str = format!("http://{}:{}", desc.gateway.host, desc.gateway.port);
let Ok(gateway_url) = Url::parse(&gateway_url_str) else {
    tracing::warn!("Failed to parse gateway URL {gateway_url_str:?} for custom domains");
    return;
};
```
Consistent with how every other failure in this function is handled — errors are logged but don't propagate.

Given the doc comment says "This is a best-effort operation", Option B feels more consistent. But happy to go with whichever you prefer!

raymondk:
Noticed that:

```
networks:
  - name: local
    mode: managed
    gateway: 
      domains:
        - mydomain.com
```

Will come back with a random port instead of the default fixed port.

crates/icp-cli/src/commands/deploy/mod.rs
```diff
@@ -377,21 +379,25 @@ async fn print_canister_urls(
...
         }
-        NetworkConfiguration::Connected { connected } => {
-            // For connected networks, use the configured URL
-            connected.http_gateway_url.clone()
-        }
+        NetworkConfiguration::Connected { connected } => connected.http_gateway_url.clone(),
     };
 
+    // Friendly domains are available for managed networks where we write custom-domains.txt
+    let has_friendly = matches!(
```
raymondk, line 397: This is assuming that all network launchers and all docker images are going to support the custom domains which is not a safe assumption to make IMO.

I think we can simplify this and continue printing the `<canister>.<localhost>:<port>` url because it is more likely to be correct.

It would be documented that some network can use the friendly names - maybe it can even be printed when the network starts if this network will support friendly names.

crates/icp-cli/src/commands/deploy/mod.rs
```diff
@@ -407,58 +413,40 @@ async fn print_canister_urls(
...
-                    let mut candid_url = http_gateway_url.clone();
-                    if let Some(domain) = domain {
-                        candid_url
-                            .set_host(Some(&format!("{ui_id}.{domain}",)))
-                            .unwrap();
+                if let Some(ui_id) = get_candid_ui_id(ctx, environment_selection).await {
+                    let domain = gateway_domain(http_gateway_url);
+                    let mut candid_url = canister_gateway_url(http_gateway_url, ui_id, None);
+                    if domain.is_some() {
                         candid_url.set_query(Some(&format!("id={canister_id}")));
```
raymondk, line 439: as a side note - could be interesting to create a friendly name for the candid ui canister too.

adamspofford: I agree but would rather keep it out of this PR.

crates/icp-cli/src/commands/deploy/mod.rs
adamspofford: I'm going to narrow it to 'is it native and is the version latest' for now, I want to keep the feature advertised to users and this should be enough following the auto-update PR. The launcher notifying the CLI whether it is supported seems like a good idea.

PR #401, opened 2026-03-04
feat: Auto-update network launcher after updating icp-cli

adamspofford-dfinity:


crates/icp/src/network/managed/cache.rs
```diff
@@ -27,12 +27,55 @@ pub fn get_cached_launcher_version(
...
+        Ok(Some(version_path.join(LAUNCHER_NAME)))
+    } else {
+        Ok(None)
+    }
+}
+
+/// Like [`get_cached_launcher_version`], but for the "latest" tag also checks
+/// whether the launcher was downloaded by an older CLI version, returning `None`
+/// if so. Pinned versions are never considered stale.
+pub fn get_cached_launcher_version_if_fresh(
```
lwshang, line 39: `get_cached_launcher_version` is now only called from `crates/icp-cli/tests/common/context.rs`, which has a comment saying it replicates the command's logic. Since the command switched to this function, the test helper should match — which would make the old function dead code and allow its removal.

adamspofford: I think the test should use the new function but I don't think the old function should be deleted, some other piece of code may want it in the future.

PR #406, opened 2026-03-05
ci: add paths-ignore to skip CI for doc-only changes

lwshang:
## Summary

- Add `paths-ignore` to `test.yml`, `checks.yml`, and `validate-examples.yml` to skip expensive CI runs when only non-code files change
- All three workflows now ignore `**.md`, `docs-site/**`, and `npm/**`
- `test.yml` additionally ignores `docs/**` since no test depends on doc content

### DIFF
--- a/.github/workflows/validate-examples.yml
+++ b/.github/workflows/validate-examples.yml
@@ -5,11 +5,16 @@ on:
     branches:
       - main
     paths-ignore:
+      # NOTE: all docs/ markdown files are covered by "**.md" above. The only
+      # docs/ files that still trigger this workflow are docs/schemas/*.json —
+      # intentionally, since this workflow validates project manifests against
+      # those schemas.
       - "**.md"
       - "docs-site/**"
       - "npm/**"
   pull_request:
     paths-ignore:
+      # See push paths-ignore NOTE above.
       - "**.md"
       - "docs-site/**"
       - "npm/**"
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -5,18 +5,17 @@ on:
     branches:
       - main
     paths-ignore:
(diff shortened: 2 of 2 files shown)


.github/workflows/validate-examples.yml
```diff
@@ -6,11 +6,13 @@ on:
       - main
     paths-ignore:
       - "**.md"
-      - "docs/**"
```
raymondk, line 9: why remove `docs/**`?

lwshang: Added inline comments to explain the intent.

.github/workflows/test.yml
```diff
@@ -4,7 +4,17 @@ on:
...
     branches:
       - main
+    paths-ignore:
+      - "**.md"
+      - "docs/**"
+      - "docs-site/**"
+      - "npm/**"
   pull_request:
+    paths-ignore:
+      - "**.md"
```
adamspofford, line 14: Doesn't this mean the tests won't run if CLI args were affected by the PR? Or if any docs were updated in addition to code. I think a whitelist would be better than a blacklist for workflow filtering.

raymondk: Normally you can't change the cli args without changing the rust code right?

lwshang: `docs/reference/cli.md` is auto-generated from Rust source, so manual edits to it don't affect program behavior — the test workflow would have nothing meaningful to catch even if it ran. Ignoring all .md files is safe for test.yml.

The real concern is the cli-ref job in checks.yml, which verifies the file is up to date. I've added a NOTE there explaining the tradeoff.

adamspofford: Oh, I misunderstood the syntax. I thought this excluded all PRs that modify .md files at all.

PR #410, opened 2026-03-05
feat: Implement update nagging

adamspofford-dfinity:


crates/icp-cli/src/commands/network/start.rs
```diff
@@ -120,26 +123,46 @@ pub(crate) async fn exec(ctx: &Context, args: &StartArgs) -> Result<(), anyhow::
...
     let autocontainerize = cfg!(windows) || settings.autocontainerize;
 
-    let network_launcher_path = if let Ok(var) = std::env::var("ICP_CLI_NETWORK_LAUNCHER_PATH") {
-        Some(PathBuf::from(var))
-    } else if !autocontainerize && let ManagedMode::Launcher(managed_cfg) = &cfg.mode {
-        let version = managed_cfg.version.as_deref().unwrap_or("latest");
-        ctx.dirs
+    let (network_launcher_path, launcher_cached_version) =
+        if let Ok(var) = std::env::var("ICP_CLI_NETWORK_LAUNCHER_PATH") {
+            (Some(PathBuf::from(var)), None)
```
raymondk: Debug statement to stay we're using the network launcher for the variable

crates/icp-cli/src/commands/network/start.rs
```diff
@@ -120,26 +123,46 @@ pub(crate) async fn exec(ctx: &Context, args: &StartArgs) -> Result<(), anyhow::
     // On Windows, always use Docker since the native launcher doesn't run there
     let autocontainerize = cfg!(windows) || settings.autocontainerize;
 
-    let network_launcher_path = if let Ok(var) = std::env::var("ICP_CLI_NETWORK_LAUNCHER_PATH") {
-        Some(PathBuf::from(var))
-    } else if !autocontainerize && let ManagedMode::Launcher(managed_cfg) = &cfg.mode {
-        let version = managed_cfg.version.as_deref().unwrap_or("latest");
-        ctx.dirs
+    let (network_launcher_path, launcher_cached_version) =
```
raymondk: if the environment variable is not set, it seems like all the rest can be encapsulated under a function with a comment explaining the expected behavior.

adamspofford: I try not to encapsulate locking in functions, as a pattern, and I don't think there's much more meat on the bone after that. I can add some comments though.

raymondk (approved):

PR #412, opened 2026-03-05
feat: Rely on the network launcher to report whether friendly domains are supported

adamspofford-dfinity:


crates/icp/src/context/tests.rs
```diff
@@ -351,6 +351,7 @@ async fn test_get_agent_for_env_uses_environment_network() {
                         root_key: local_root_key.clone(),
                         api_url: Url::parse("http://localhost:8000").unwrap(),
                         http_gateway_url: None,
+                        use_friendly_domains: false,
```
lwshang, line 354: The test changes only add `use_friendly_domains: false` to existing constructors. There's no test that exercises the `true` path to verify friendly domain URLs are actually printed in `print_canister_urls`.

crates/icp/src/network/managed/docker.rs
```diff
@@ -499,6 +499,10 @@ pub async fn spawn_docker_launcher(
             root_key: hex::decode(&launcher_status.root_key).context(ParseRootKeySnafu {
                 key: &launcher_status.root_key,
             })?,
+            use_friendly_domains: launcher_status
+                .supported_features
+                .iter()
+                .any(|f| f == "custom-domains"),
```
lwshang: The feature name `"custom-domains"` appears as a raw string in both `docker.rs` and `launcher.rs`. Consider extracting to a constant to avoid typo-related bugs.

crates/icp/src/context/tests.rs
adamspofford: Not in the unit tests, this is just filling holes. But the previous PR created an integration test that checks the right domain is printed, and this alters the behavior it's testing.

lwshang (approved):

PR #416, opened 2026-03-06
feat: Candid building command

adamspofford-dfinity:
Also factors out the candid assist related code, and adds support for specifying a candid file to both candid build and canister call.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Unreleased
 
 * feat: `icp candid build` allows using the Candid-assist wizard from `icp canister call` to construct a precompiled argument file
+* feat: `icp canister logs` supports filtering by timestamp (`--since`, `--until`) and log index (`--since-index`, `--until-index`)
+* feat: Support `log_memory_limit` canister setting in `icp canister settings update` and `icp canister settings sync`
 * feat: Leaving off the method name parameter in `icp canister call` prompts you with an interactive list of methods
 * fix: Correct templating of special HTML characters in recipes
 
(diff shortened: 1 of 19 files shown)


crates/icp-cli/src/commands/candid/build.rs
```diff
@@ -0,0 +1,213 @@
...
+                    },
+            }) if matches!(*source, LookupIdError::IdNotFound { .. }) => {
+                bail!(
+                    "Canister {canister_name} has not been deployed in environment {environment_name}. This command requires an active deployment to reference"
+                );
+            }
+            Err(e) => return Err(e).context("failed to look up canister ID"),
+        };
+        let Some(interface) = get_candid_type(&agent, cid).await else {
+            bail!("could not fetch Candid interface from canister {cid}");
```
raymondk:
```suggestion
            bail!("Could not fetch Candid interface from `candid:service` metadata section of canister {cid}");
```

crates/icp-cli/src/commands/candid/build.rs
```diff
@@ -0,0 +1,213 @@
...
+#[command(group(ArgGroup::new("input").required(true).args(["canister", "candid_file"])))]
+pub(crate) struct BuildArgs {
+    #[command(flatten)]
+    pub(crate) cmd_args: args::OptionalCanisterCommandArgs,
+
+    /// Name of canister method to build arguments for.
+    /// If not provided, an interactive prompt will be launched.
+    pub(crate) method: Option<String>,
+
+    /// Output file path. Pass `-` for stdout.
```
raymondk, line 36: Isn't this a little unusual, why not make it optional?

adamspofford: With the default value being bin, this would dump garbage to the terminal if you forgot it. (*Should* it be bin?)

PR #418, opened 2026-03-10
feat: Stop canisters before upgrading

adamspofford-dfinity:


crates/icp-cli/src/operations/install.rs
```diff
@@ -210,6 +228,39 @@ async fn do_install_operation(
...
+    // Stop the canister before proceeding
+    if is_upgrade {
+        mgmt.stop_canister(canister_id)
+            .await
+            .context(StopCanisterSnafu { canister_name })?;
+    }
+    // Install the canister
+    let install_result = f.await;
+    // Restart the canister whether or not the installation succeeded
+    if is_upgrade {
```
lwshang: If a canister was already stopped before this command runs (e.g. the user manually stopped it), it will be started after the upgrade regardless. The code unconditionally calls start_canister without checking the prior state.
Consider querying canister status first and only restarting if it was Running before the stop.

crates/icp-cli/src/operations/install.rs
```diff
@@ -197,9 +212,12 @@ async fn do_install_operation(
...
-            .map_err(|source| InstallOperationError::Agent { source })?;
+        stop_and_start_if_upgrade(&mgmt, canister_id, canister_name, mode, async {
+            builder
+                .await
+                .map_err(|source| InstallOperationError::Agent { source })
+        })
+        .await?;
 
         // Clear chunk store after successful installation to free up storage
         mgmt.clear_chunk_store(canister_id)
```
lwshang: In the chunked path, if the install succeeds but start_canister fails, we return the start error before reaching clear_chunk_store. The chunk store is leaked until the next install. Minor, but worth noting.

crates/icp-cli/src/operations/install.rs
```diff
@@ -210,6 +228,39 @@ async fn do_install_operation(
...
+    if is_upgrade {
+        let start_result = mgmt.start_canister(canister_id).await;
+        if let Err(start_error) = start_result {
+            // If both install and start failed, report the install error since it's more likely to be the root cause
+            if let Err(install_error) = install_result {
+                warn!("Failed to start canister after failed upgrade: {start_error}");
+                return Err(install_error);
+            } else {
+                return Err(start_error).context(StartCanisterSnafu { canister_name });
+            }
```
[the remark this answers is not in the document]
adamspofford: Fine as is, logging to be addressed in another PR

PR #420, opened 2026-03-10
feat: Add support for creating canisters on a cloud engine

adamspofford-dfinity:
Manually tested on a pocket-ic prerelease build.


crates/icp-cli/src/operations/create.rs
```diff
@@ -170,31 +230,6 @@ impl CreateOperation {
...
-        .map_err(|err| CreateOperationError::Registry { message: err })?
-        .subnet_id
-        .ok_or(CreateOperationError::MissingSubnetId)?;
-
-    Ok(out)
-}
-
 async fn get_available_subnets(agent: &Agent) -> Result<Vec<Principal>, CreateOperationError> {
     let bs = agent
         .query(&CYCLES_MINTING_CANISTER_PRINCIPAL, "get_default_subnets")
```
raymondk, line 235: Would this ever return an engine? Do we need to change something in the cycles minting canister?

raymondk (commented):
LGTM - but needs a changelog entry

crates/icp-cli/src/operations/create.rs
viviveevee: This endpoint's purpose is to return subnets that are open for everyone to create canisters on.
Technically it can return an engine, but that would be a bad mistake by proposal reviewers since AFAIU engines are meant to be more private

raymondk (approved):

PR #426, opened 2026-03-12
chore: Fix log filters and refactor code to use log macros

adamspofford-dfinity:
The rule I have followed is: If, to a person who already knows the tool's behavior, the message would communicate nothing that a status code of 0 would not communicate, it becomes `info!`; otherwise, it becomes `println!`.

### DIFF
--- a/crates/icp-cli/src/commands/canister/snapshot/upload.rs
+++ b/crates/icp-cli/src/commands/canister/snapshot/upload.rs
@@ -189,7 +189,7 @@ pub(crate) async fn exec(ctx: &Context, args: &UploadArgs) -> Result<(), anyhow:
             // Clean up progress file on success
             delete_upload_progress(paths)?;
 
-            info!("Snapshot {} uploaded successfully", progress.snapshot_id);
+            println!("Snapshot {} uploaded successfully", progress.snapshot_id);
 
             Ok::<_, anyhow::Error>(progress.snapshot_id)
         })


crates/icp-cli/src/logging.rs
```diff
@@ -1,23 +1,117 @@
...
+    >,
     Targets,
     S,
 >;
 
 pub(crate) fn debug_layer<S: Subscriber + for<'a> LookupSpan<'a>>() -> DebugLayer<S> {
     let workspace_targets = Targets::new()
-        .with_target("icp-cli", Level::DEBUG)
+        .with_default(Level::WARN)
         .with_target("icp", Level::DEBUG);
```
lwshang, line 32:
```suggestion
        .with_default(Level::WARN)
        .with_target("icp", Level::DEBUG)
        .with_target("icp_cli", Level::DEBUG);
```
The "icp" target matches `icp::*` but not` icp_cli::*` — the prefix match is segment-based (split on `::`). Since this PR moves `info!`/`warn!`/`error!` calls from TermWriter (in the icp crate) into `icp_cli::commands::*` directly, they'll now fall through to `with_default(WARN)` in `--debug` mode, making debug output less informative than normal mode.

(Note: the old code had `"icp-cli"` which also didn't match `icp_cli`, but it didn't matter because `TermWriter.write_line()` called `info!()` from the icp crate.)

crates/icp-cli/src/commands/canister/snapshot/upload.rs
```diff
@@ -187,24 +183,19 @@ pub(crate) async fn exec(ctx: &Context, args: &UploadArgs) -> Result<(), anyhow:
...
             }
 
             // Clean up progress file on success
             delete_upload_progress(paths)?;
 
-            ctx.term.write_line(&format!(
-                "Snapshot {} uploaded successfully",
-                progress.snapshot_id
-            ))?;
+            println!("Snapshot {} uploaded successfully", progress.snapshot_id);
```
lwshang: This uses `println!` while the analogous success messages in `download.rs:205`, `delete.rs:48`, and `restore.rs:64` use `info!`. Should this be `info!` as well for consistency?

crates/icp-cli/src/logging.rs
adamspofford: But it is based on crate name, not package name. Both crates are named icp despite one package being named icp-cli, and the proposed line does nothing.

lwshang (approved):

PR #427, opened 2026-03-12
fix: use dorny/paths-filter so required checks pass on docs-only PRs

lwshang:
## Summary
- Replaces workflow-level `paths-ignore` with job-level path filtering using `dorny/paths-filter` in `checks.yml`, `test.yml`, and `validate-examples.yml`
- When a PR only touches docs, markdown, or npm files, the work jobs are skipped but the workflow still triggers — so required status checks (`fmt:required`, `lint:required`, `test:required`) report as "skipped" instead of missing entirely
- This unblocks docs-only PRs like #425 

## Test plan
- [ ] Verify this PR's own CI passes (all required checks should run normally since it touches workflow files)
- [ ] After merging, verify #425 checks pass (required checks should report as skipped)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,6 +86,8 @@ jobs:
     name: ${{ matrix.test }} on ${{ matrix.os }}
     # Wait until unit tests have run so we can reuse the cache
     needs: [discover, unit-tests]
+    # Check discover's result (not changes.outputs.src) because this job
+    # needs discover's matrix output, which is only available when it ran.
     if: needs.discover.result == 'success'
     runs-on: ${{ matrix.os }}
     strategy:


.github/workflows/test.yml
```diff
@@ -71,6 +86,7 @@ jobs:
     name: ${{ matrix.test }} on ${{ matrix.os }}
     # Wait until unit tests have run so we can reuse the cache
     needs: [discover, unit-tests]
+    if: needs.discover.result == 'success'
```
adamspofford, line 91: Why's this one different?

lwshang: Added an inline comment explaining why: this job needs discover's matrix output (via `fromJson`), which is only available when discover actually ran — so we check its result rather than `changes.outputs.src`.

Issue #491, opened 2026-04-09
Improve --help output for agent usability


marc0olo:
## Summary

This issue tracks all `--help` output improvements needed to make icp-cli more usable by agents — particularly those trained primarily on `dfx`. Poor or missing descriptions cause agents to guess flag formats, hang in interactive prompts, or misidentify commands.

All changes are expected to land in a single PR.

---

## 1. `icp canister settings update` — fill in all empty option descriptions

Several flags have completely empty descriptions, making it impossible for an agent to construct a valid call without guessing.

| Flag | Add description |
|---|---|
| `--compute-allocation` | `Compute allocation percentage (0–100). Represents a guaranteed share of a subnet's compute capacity` |
| `--log-visibility` | `Who can read the canister's logs [possible values: controllers, public, allowed-viewers]` |
| `--add-log-viewer` | `Add a principal to the allowed log viewers list` |
| `--remove-log-viewer` | `Remove a principal from the allowed log viewers list` |
| `--set-log-viewer` | `Replace the allowed log viewers list with the specified principals` |
| `--add-environment-variable` | `Add a canister environment variable in KEY=VALUE format` |
| `--remove-environment-variable` | `Remove a canister environment variable by key name` |

---

## 2. `icp canister settings show` — wrong description

Current description says "Show the **status** of a canister" — identical wording to `icp canister status`. An agent trying to replicate `dfx canister status` will be confused about which command to use.

Change to:

> Show a canister's configured settings (controllers, compute allocation, memory limits, log visibility, environment variables).

---

## 3. `icp canister status` — document no-argument behavior and add examples

The `[CANISTER]` argument is listed as optional but there is no explanation of what happens when it is omitted. A dfx-trained agent expects `dfx canister status --all` as an explicit opt-in; here silence means all.

**Update the `[CANISTER]` argument description** to add:

> If omitted, shows status for all canisters in the environment.

**Add an `after_long_help` examples block** (matching the style already used by `icp network status`):

```
# Status of all canisters in local environment
icp canister status

# Status of one canister by name
icp canister status backend -e local

# Print only canister IDs (useful for scripting)
icp canister status -i

# JSON output for all canisters
icp canister status --json
```

---

## 4. `icp network start` — document default port and `--project-root-override` tip

Nothing in the help mentions that the gateway binds to port 8000 by default, that this is configurable, or how to stop a network in another project directory without `cd`.

Add to the command description or `after_long_help`:

[...]

PR #492, opened 2026-04-09
docs: improve --help output for agent usability

marc0olo:
Addresses all items from #491, plus one additional gap found during a full help audit.

## Changes

**`icp canister settings update`** — fill in all previously empty option descriptions:
- `--compute-allocation`: clarify 0–100 percentage range
- `--log-visibility`: clarify valid values are `controllers`/`public` only (not `allowed-viewers`), and point to `--add-log-viewer`/`--set-log-viewer` for per-principal access
- `--add-log-viewer`, `--remove-log-viewer`, `--set-log-viewer`: add descriptions
- `--add-environment-variable`: document `KEY=VALUE` format
- `--remove-environment-variable`: document that it takes a key name

**`icp canister settings show`** — fix description to accurately state the command shows full canister status (same output as `icp canister status`), not just configured settings; note that a canister must always be specified; fix "enviroment" typo in argument description

**`icp canister status`** — document that omitting `[CANISTER]` queries all canisters in the environment (dfx required `--all` as explicit opt-in); fix "enviroment" typo in argument description; add `after_long_help` examples block

**`icp network start`** — document the default port (8000), how to configure a different port via `gateway.port` in `icp.yaml`, and how to stop a network in another project directory using `--project-root-override`

**`icp new`** — replace misleading `template_values_file` reference in `--silent` (icp-cli does not expose `--template-values-file`) with accurate description of non-interactive behaviour; improve `--define` description to mention that variable names are template-specific and that `--define` is required in `--silent` mode for variables with no template default

**`icp canister list`** — document plain-text output (names, one per line) and JSON schema (`{"canisters": [...]}`)

**`icp environment list`** — fix "enviroments" typo; add context pointing to `icp project show` for expanded configuration

**`icp cycles transfer`** — fix `////` (plain comment) to `///` (doc comment) on `--from-subaccount` so the description appears in `--help`

**`icp cycles mint`** — document that exactly one of `--icp` or `--cycles` is required; previously both appeared fully optional in `--help` but a runtime error was returned if neither was provided

**`docs/reference/cli.md`** — regenerated to reflect all the above changes

### DIFF
--- a/crates/icp-cli/src/commands/canister/list.rs
+++ b/crates/icp-cli/src/commands/canister/list.rs
@@ -6,7 +6,10 @@ use serde::Serialize;
 
 use crate::options::EnvironmentOpt;
 
-/// List the canisters in an environment
+/// List the canisters in an environment.
+///
+/// Prints canister names, one per line. Use --json for machine-readable output
+/// (returns {"canisters": ["name1", "name2", ...]})
 #[derive(Debug, Args)]
 pub(crate) struct ListArgs {
     #[command(flatten)]
--- a/crates/icp-cli/src/commands/canister/settings/show.rs
+++ b/crates/icp-cli/src/commands/canister/settings/show.rs
@@ -15,7 +15,7 @@ pub(crate) struct ShowArgs {
     #[command(flatten)]
     pub(crate) cmd_args: CanisterCommandArgs,
 
-    /// Format output in json
+    /// Format output as JSON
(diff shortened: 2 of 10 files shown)


marc0olo:
I guess changes in https://github.com/dfinity/icp-cli/pull/499 also require some update here now.

viviveevee (approved):

PR #495, opened 2026-04-10
feat: Login with Internet Identity

adamspofford-dfinity:
* New identity type for II logins. Attached to a delegation stored in a separate delegations folder.
    * II keys are stored slightly differently in the keyring to ensure no code path will export them.
* New command `icp identity link ii` to create a new II identity. This opens a URL in the user's browser, and then receives a post on an axum server.
* New command `icp identity login` refreshes the delegation of an identity that can be refreshed (not necessarily II-specific). Calls the same browser login function from `icp identity link ii`.


crates/icp-cli/src/operations/ii_poll.rs
```diff
@@ -0,0 +1,198 @@
+use std::net::SocketAddr;
```
raymondk: Wouldn't it make sense to rename this to something else - since it's not technically polling anymore.

Also does it make sense to be an operation? Seems very specific to the `link ii` command.

adamspofford-dfinity: It's shared between link and login.

crates/icp-cli/src/commands/identity/link/ii.rs
```diff
@@ -0,0 +1,106 @@
...
+/// Link an Internet Identity to a new identity
+#[derive(Debug, Args)]
+pub(crate) struct IiArgs {
+    /// Name for the linked identity
+    name: String,
+
+    /// Where to store the session private key
+    #[arg(long, value_enum, default_value_t)]
+    storage: StorageMode,
+
```
raymondk, line 46: Optional `--domain` with a default 
For now we append `/icp-login#....` 

We need to store the domain in the config file so that `icp identity login` can renew from the same domain.

crates/icp-cli/src/commands/identity/link/ii.rs
```diff
@@ -0,0 +1,377 @@
...
+    let with_scheme = if s.contains("://") {
+        s.to_owned()
+    } else {
+        format!("https://{s}")
+    };
+    Url::parse(&with_scheme).map_err(|e| e.to_string())
+}
+
+pub(crate) async fn exec(ctx: &Context, args: &IiArgs) -> Result<(), IiError> {
+    ctx.dirs
```
raymondk, line 62: We can have a follow up PR to do the same thing in the HSM and in `identity new`

crates/icp-cli/src/commands/identity/link/mod.rs
```diff
@@ -1,9 +1,12 @@
 use clap::Subcommand;
 
 pub(crate) mod hsm;
+pub(crate) mod ii;
 
 /// Link an external key to a new identity
 #[derive(Debug, Subcommand)]
 pub(crate) enum Command {
     Hsm(hsm::HsmArgs),
+    #[command(hide = true)]
```
raymondk: TODO comment to say that we will unhide it when the feature is public?

crates/icp-cli/src/commands/identity/link/ii.rs
```diff
@@ -0,0 +1,377 @@
...
+struct CallbackState {
+    chain_tx: std::sync::Mutex<Option<oneshot::Sender<DelegationChain>>>,
+    shutdown_tx: std::sync::Mutex<Option<oneshot::Sender<()>>>,
+}
+
+fn cors_headers() -> HeaderMap {
+    let mut headers = HeaderMap::new();
+    headers.insert(
+        header::ACCESS_CONTROL_ALLOW_ORIGIN,
+        HeaderValue::from_static("*"),
```
raymondk: Shouldn't this be limited to the URL that was opened?

crates/icp-cli/src/commands/identity/link/ii.rs
```diff
@@ -0,0 +1,377 @@
...
+};
+use indicatif::{ProgressBar, ProgressStyle};
+use snafu::{ResultExt, Snafu, ensure};
+use tokio::{net::TcpListener, sync::oneshot};
+use tracing::{info, warn};
+use url::Url;
+
+use crate::commands::identity::StorageMode;
+
+/// Link an Internet Identity to a new identity
```
raymondk, line 33: Maybe we can add more details here in the expanded help about how this can be useful?

adamspofford-dfinity: What were you envisioning?

crates/icp-cli/src/commands/identity/login.rs
```diff
@@ -0,0 +1,101 @@
...
+            let spec = list
+                .identities
+                .get(&args.name)
+                .context(IdentityNotFoundSnafu { name: &args.name })?;
+            match spec {
+                IdentitySpec::InternetIdentity {
+                    algorithm,
+                    storage,
+                    host,
+                    ..
```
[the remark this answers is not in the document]
adamspofford-dfinity: That's the only refreshable identity type presently.

crates/icp-cli/src/commands/identity/link/ii.rs
raymondk:
```suggestion
/// Link an Internet Identity to a new identity
///
/// Creates a local identity that acts on behalf of an Internet Identity via a
/// delegation chain, so you can call canisters as your II principal from the CLI.
///
/// Benefits over a standalone CLI identity:
///
/// - Same principal across surfaces: calls made by the CLI land as the same
///   principal you use in the browser when you signed in with this II anchor on that domain,
///   so your ledger balances, permissions, and canister ownership carry over.
/// - The CLI only ever holds a time-bounded delegation, not your II secret.
/// - Time-bounded: The delegation expires on its own.
///   Re-authenticate later with `icp identity login`.
/// - No seed phrase to manage: unlike `icp identity new`, there is no PEM or
///   mnemonic for this identity — the root of trust lives with II.
```

crates/icp-cli/src/commands/identity/link/ii.rs
```diff
@@ -0,0 +1,392 @@
...
+    let fragment = {
+        let mut scratch = Url::parse("x:?").expect("infallible");
+        scratch
+            .query_pairs_mut()
+            .append_pair("public_key", &key_b64)
+            .append_pair("callback", &callback_url);
+        scratch.query().expect("just set").to_owned()
+    };
+    let mut login_url = host.join(login_path).expect("login_path is a valid path");
+    login_url.set_fragment(Some(&fragment));
```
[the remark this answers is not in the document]
adamspofford-dfinity: Presumably the site would have a good reason for doing so.

crates/icp-cli/src/commands/identity/link/ii.rs
```diff
@@ -0,0 +1,392 @@
...
+                tokio::select! {
+                    _ = enter_rx.recv(), if !browser_opened => {
+                        browser_opened = true;
+                        spinner.set_message("Waiting for Internet Identity authentication...");
+                        spinner.enable_steady_tick(std::time::Duration::from_millis(100));
+                        let _ = open::that(login_url.as_str());
+                    }
+                    // Yield so the other branches in the outer select! can fire.
+                    _ = tokio::task::yield_now() => {}
+                }
```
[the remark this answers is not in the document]
lwshang: Can you fix the yield_now busy loop Copilot flagged?

crates/icp-cli/src/commands/identity/link/ii.rs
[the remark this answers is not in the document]
lwshang: Please add the relative-path validation Copilot suggested.

crates/icp-cli/src/commands/identity/link/ii.rs
```diff
@@ -0,0 +1,392 @@
...
+            return InterruptedSnafu.fail();
+        }
+        _ = async {
+            loop {
+                tokio::select! {
+                    _ = enter_rx.recv(), if !browser_opened => {
+                        browser_opened = true;
+                        spinner.set_message("Waiting for Internet Identity authentication...");
+                        spinner.enable_steady_tick(std::time::Duration::from_millis(100));
+                        let _ = open::that(login_url.as_str());
```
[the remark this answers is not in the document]
lwshang: Agree. In headless environments this silently hangs forever. Please handle the error and print the full login URL so users can open it manually.

crates/icp-cli/src/commands/identity/link/ii.rs
adamspofford-dfinity: This seems like stuff that should go in the prose docs, not the CLI help page.

crates/icp-cli/src/commands/identity/link/ii.rs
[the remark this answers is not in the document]
adamspofford-dfinity: Is there a reason a site should be prevented from doing this?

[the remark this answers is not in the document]
lwshang: Initially concerned a compromised discovery endpoint could redirect the browser to an attacker-controlled login page. But since the user explicitly trusts --host, a compromised host could serve a malicious page directly anyway — path validation wouldn't add meaningful protection. Resolving.

crates/icp-cli/src/commands/identity/link/ii.rs
lwshang: Should we use `#[command(after_long_help = ...)]` here (like some other subcommands do)? That way Raymond's detail usage guide gets included in the generated `cli.md` without cluttering `--help`.

adamspofford-dfinity: It's a good place for examples, but I just don't feel like help is where advertisement goes. Maybe the help line could be rephrased to better explain what is going on and how precisely that works, but the bulleted list of benefits is guide-level while command help should (IMO) be reference-level

raymondk (approved):

lwshang (approved):

PR #510, opened 2026-04-17
ci: auto-sync docs/vX.Y branch when a release tag is pushed

marc0olo:
## Summary

Adds automation to keep \`docs/vX.Y\` branches in sync with release tags, and prevents a redundant docs re-deployment after the sync.

**Files changed:**
- \`.github/workflows/sync-docs-branch.yml\` — new workflow that force-resets \`docs/vX.Y\` on every release tag
- \`.github/workflows/docs.yml\` — skips \`publish-versioned-docs\` when \`github-actions[bot]\` pushes (avoids redundant re-deploy)
- \`.claude/skills/release/task6-docs.md\` — updated release playbook to document the automation

## The problem

Both \`v*\` tags and \`docs/v*\` branches trigger the same \`publish-versioned-docs\` job and deploy to the same \`/X.Y/\` folder on the live site. Without sync automation, \`docs/vX.Y\` silently drifts behind the tag — and any future push to it (e.g. a docs hotfix) would overwrite the live docs with older content. This happened: \`docs/v0.2\` was sitting at its March 30 state while the live \`/0.2/\` docs were correctly deployed from the April 8 \`v0.2.3\` tag.

### `sync-docs-branch.yml`

Triggers on every \`v*\` release tag push (pre-release tags excluded via \`!v*-*\`).

```
Tag v0.2.3 pushed
  └─ Extract minor version: 0.2 → DOCS_BRANCH=docs/v0.2
  └─ docs/v0.2 exists remotely?
       YES → git checkout -b docs/v0.2 v0.2.3 && git push --force
       NO  → git checkout -b docs/v0.2 v0.2.3 && git push
```

The force-push is an atomic reset — no merge commits, no conflict resolution, no manual steps. The branch ends up pointing exactly at the tag commit.

### Skip guard in `docs.yml`

The force-push to \`docs/vX.Y\` triggers \`docs.yml\` again (because docs files changed). Without a guard, \`publish-versioned-docs\` would re-deploy content that is identical to what was just deployed from the tag — wasting CI minutes.

The added condition:
```yaml
if: github.event_name == 'push' && github.actor != 'github-actions[bot]' && (...)
```

skips \`publish-versioned-docs\` only when \`github-actions[bot]\` pushes to \`docs/v*\`. It does **not** affect:
- Tag pushes (always done by humans or release tooling, not the bot)
- Human pushes to \`docs/v*\` (docs hotfixes still deploy normally)
- \`workflow_dispatch\` (condition requires \`event_name == 'push'\`)

[...]

PR #511, opened 2026-04-17
ci: replace docs/v* branches with tags for versioned docs deployment

marc0olo:
## What and why

Replaces `docs/v*` **branches** with `docs/v*` **tags** as the mechanism for controlling versioned documentation deployments, and adds automation to keep them in sync after every release.

### The problem with branches

Both `v*` release tags and `docs/v*` branches triggered the same `publish-versioned-docs` job and deployed to the same `/X.Y/` folder. Branches drift silently — a routine push to a stale `docs/v0.2` branch overwrites the live docs with older content, with no warning. This happened in practice: `docs/v0.2` sat at its March 30 state while the live `/0.2/` was correctly deployed from the April 8 `v0.2.3` tag. Automating a branch-based sync (PR #510) required branch protection bypass configuration and still produced merge conflicts.

### Why tags

Tags are immutable by convention. Moving one requires a deliberate `git push --force` — it cannot happen accidentally. No branch protection rules apply, so no bypass configuration is needed.

---

### Triggers (changed)

| Event | Before | After |
|---|---|---|
| `v0.2.3` release tag pushed | Deploys `/0.2/` | Deploys `/0.2/` (unchanged) |
| `docs/v0.2` **branch** pushed | Deploys `/0.2/` | **No longer triggers deployment** |
| `docs/v0.2` **tag** pushed or force-moved | No effect | **Deploys `/0.2/`** |

### Automatic sync on every release

Two workflows run concurrently when a release tag is pushed:

```
v0.2.3 pushed
  ├─ docs.yml → publish-versioned-docs
  │    └─ Builds and deploys /0.2/ from the v0.2.3 commit ✓
  │
  └─ sync-docs-tag.yml  ← new
       └─ Creates or force-moves docs/v0.2 tag to v0.2.3 commit
            └─ Uses GITHUB_TOKEN → does NOT re-trigger docs.yml ✓
```

GitHub prevents recursive workflow runs from `GITHUB_TOKEN` pushes by design, so the tag move never causes a second deployment.

### Doing a docs-only fix between releases

The rule: **always merge the fix to `main` first.** The `docs/vX.Y` tag is only for immediate deployment — when the next patch is released, `sync-docs-tag.yml` resets `docs/vX.Y` to the release commit. Any commit that exists only on the tag (not in `main`) will be lost at that point.

Workflow for an immediate re-deploy of `/0.2/`:

```bash
# 1. Merge the fix to main via a normal PR  (required — do this first)

# 2. Immediately deploy the fix to /0.2/ without waiting for a release:
git fetch --tags
git checkout -b temp/docs-fix-v0.2 docs/v0.2   # start from current tag state
git cherry-pick <merged-commit-sha>              # the commit already in main

git tag -f docs/v0.2 HEAD
git push origin refs/tags/docs/v0.2 --force     # triggers re-deploy of /0.2/

git branch -D temp/docs-fix-v0.2                # local branch; not pushed
```

The commits remain reachable via the tag — no remote branch is ever pushed.

---

## Files changed

| File | Change |
|---|---|
| `.github/workflows/docs.yml` | Trigger: `docs/v*` branch → tag (`+ !docs/v*-*`); version extraction and job condition updated to `refs/tags/docs/v` |
| `.github/workflows/sync-docs-tag.yml` | **New.** Creates or force-moves `docs/vX.Y` tag to the release commit after each release tag push |
| `docs-site/README.md` | Triggers section updated; docs-only fix workflow and `GITHUB_TOKEN` behavior explained |
| `.claude/docs-guidelines.md` | New section: "Docs-Only Fixes for Released Versions" — the merge-to-main-first rule, full workflow, and why commits must be in `main` before the next release |
| `.claude/skills/release/task6-docs.md` | Documents both automated workflows (`docs.yml` + `sync-docs-tag.yml`); cross-references docs guidelines |
| `.claude/skills/release/rollback.md` | Task 2 rollback: steps to revert `docs/vX.Y` tag; Task 6 rollback: clarified that tag cleanup belongs to Task 2 |

---

## Migration: after this PR merges

Two `docs/v*` branches currently exist (`docs/v0.1`, `docs/v0.2`). Neither will trigger deployments after this merges. Tag each at its current branch tip, then delete the branches:

```bash
git fetch origin

# docs/v0.1
git tag docs/v0.1 origin/docs/v0.1
git push origin refs/tags/docs/v0.1
git push origin --delete docs/v0.1
git branch -D docs/v0.1 2>/dev/null || true

# docs/v0.2
git tag docs/v0.2 origin/docs/v0.2
git push origin refs/tags/docs/v0.2
git push origin --delete docs/v0.2
git branch -D docs/v0.2 2>/dev/null || true
```

The live `/0.1/` and `/0.2/` docs are unaffected — already correctly deployed on `docs-deployment`.

---

## Test plan

- [ ] Merge this PR, then run the migration steps above for `docs/v0.1` and `docs/v0.2`
- [ ] On next release tag push: verify `sync-docs-tag.yml` runs and moves `docs/vX.Y` to the release commit
- [ ] Verify `docs.yml` is **not** re-triggered by the `sync-docs-tag.yml` tag move
- [ ] Manually force-move `docs/v0.2` to a test commit and verify `docs.yml` **is** triggered and redeploys `/0.2/`
- [ ] Verify `docs/v*` branch pushes no longer trigger `publish-versioned-docs`

### DIFF
--- a/.claude/docs-guidelines.md
+++ b/.claude/docs-guidelines.md
@@ -30,6 +30,32 @@ Documentation follows the Diátaxis framework:
 - When referencing alternatives in other docs, maintain this order: "Homebrew, shell script, ..." (e.g., "See the Installation Guide for Homebrew, shell script, or other options")
 - Both `icp-cli` and `ic-wasm` are available as official Homebrew formulas: `brew install icp-cli` and `brew install ic-wasm`
 
+## Docs-Only Fixes for Released Versions
+
+Versioned docs deployments (e.g. `/0.2/`) are controlled by `docs/vX.Y` tags. To fix or improve docs for an already-released version without cutting a new code release:
+
+**Rule: always merge the change to `main` first.** The `docs/vX.Y` tag is only for immediate deployment — when the next patch release is tagged, `sync-docs-tag.yml` resets `docs/vX.Y` to the new release commit. Any commit that exists only on the tag (not in `main`) will be silently lost at that point.
+
+**Workflow:**
+
+```bash
+# 1. Merge the fix to main via a normal PR (always required)
+
+# 2. To immediately deploy the fix to /X.Y/ without waiting for a release:
+git fetch --tags
+git checkout -b temp/docs-fix-vX.Y docs/vX.Y  # start from current tag state
+git cherry-pick <commit-sha-from-main>          # pick the merged commit(s)
+
+git tag -f docs/vX.Y HEAD
+git push origin refs/tags/docs/vX.Y --force    # triggers re-deploy of /X.Y/
+
+git branch -D temp/docs-fix-vX.Y               # local branch no longer needed
+```
+
+The commits remain reachable via the tag — no remote branch is needed.
+
+**On the next release:** `sync-docs-tag.yml` resets `docs/vX.Y` to the release commit. Because the fix was already merged to `main`, the release will contain it, and the reset preserves it automatically.
+
 ## Writing Guidelines
 
 - Use "canister environment variables" (not just "environment variables") when referring to runtime variables stored in canister settings — this distinguishes them from shell/build environment variables
--- a/.claude/skills/release/rollback.md
+++ b/.claude/skills/release/rollback.md
@@ -9,8 +9,20 @@ If something fails mid-release, here's how to clean up depending on how far you 
   git tag -d v$ARGUMENTS
   ```
   If a GitHub Release was already created, delete it first via `gh release delete v$ARGUMENTS --yes`, then delete the tag.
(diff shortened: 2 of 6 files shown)


marc0olo:
Closing in favour of a simpler branch-based approach. Instead of switching to tags, we keep `docs/v*` branches but auto-delete them after each non-beta release — so branches can never go stale. The `docs/v*` branch protection ruleset has been updated to remove the `deletion` rule, allowing the automation to do the cleanup. A new PR will follow with the auto-delete workflow.

PR #515, opened 2026-04-21
chore: add 404.html

viviveevee:
Add a 404.html page to the docs


marc0olo (approved):
Look and feel as well as content can be improved later. Let's first make sure it works 👍

PR #527, opened 2026-04-23
feat: Automatic delegation sessions

adamspofford-dfinity:
Password-protected identities will generate 'session' delegations by default so you don't have to keep entering your password.
* This happens by default whenever you use your password with an identity-based command
* This can be explicitly triggered with `icp identity reauth`

A user setting configures the session duration, with two more minutes tacked on to account for clock drift, or can disable the automatic form of this feature.

Only keyring storage is used, since plaintext would not be secure enough for automation. Keyring errors are not an error in the automatic path; it just uses the identity normally.


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+* feat: Password-protected identities now only need your password once every five minutes (configurable)
```
viviveevee: Probably worth writing a sentence or two more that explains configuration options or where to find out more about the feature

crates/icp-cli/src/commands/settings.rs
```diff
@@ -49,11 +53,52 @@ struct UpdateCheckArgs {
     value: Option<UpdateCheck>,
 }
 
+#[derive(Debug, Args)]
+struct SessionLengthArgs {
+    /// Set to `<N>m` (e.g. `5m`) or `disabled`. If omitted, prints the current value.
```
viviveevee: Inconsistent suffix support with `icp identity login --duration`. Would be nice to support identical syntax

crates/icp/src/identity/key.rs
```diff
@@ -186,11 +204,21 @@ fn load_pem_identity(
...
+        PemFormat::Pbes2 => load_pbes2_identity(&doc, algorithm, password_func, &origin)?,
+        PemFormat::Plaintext => load_plaintext_identity(&doc, algorithm, &origin)?,
+    };
 
-        PemFormat::Plaintext => load_plaintext_identity(&doc, algorithm, &origin),
+    // After unlocking a Pbes2 PEM, create and cache a short-lived session delegation.
+    if *format == PemFormat::Pbes2
+        && let Some(duration) = pem_session_duration
+        && let Ok(delegated) =
+            create_pem_session_and_build_identity(dirs, name, &*identity, duration)
```
viviveevee: Do you think it could be appropriate to log errors in `--verbose` mode? I imagine without some form of logging errors it will be very hard to debug any issues with sessions

crates/icp/src/identity/key.rs
```diff
@@ -266,13 +294,221 @@ const SERVICE_NAME: &str = "icp-cli";
...
 /// export these keys.
 fn dlg_keyring_key(name: &str) -> String {
     format!("delegation:{name}")
 }
 
+/// Tries to load a previously cached PEM session delegation from keyring + disk.
+///
+/// Returns `None` on any failure (missing, expired, or keyring unavailable) so the
+/// caller falls back to normal PEM loading.
+fn try_load_pem_session(dirs: LRead<&IdentityPaths>, name: &str) -> Option<Arc<dyn Identity>> {
```
viviveevee, line 335: Same here w.r.t. logging

crates/icp/src/identity/key.rs
```diff
@@ -781,6 +1019,24 @@ pub fn rename_identity(
             let new_path = dirs.key_pem_path(new_name);
             let contents = fs::read(&old_path).context(CopyKeyFileSnafu)?;
             fs::write(&new_path, &contents).context(CopyKeyFileSnafu)?;
+
+            // Best-effort: migrate any cached session delegation.
```
viviveevee, line 1125: I would add a test to show that it _can_ work, even if it's ok if it doesn't

viviveevee (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ bump. Currently experimental: sync plugins.
 
 # Unreleased
 
+* feat: Password-protected identities now only need your password once per session. The session length defaults to 5 minutes and can be changed with `icp settings session-length <DURATION>` (e.g. `30m`, `1h`) or turned off with `icp settings session-length disabled`. You can also explicitly create or refresh a session with `icp identity login <NAME> [--duration <DURATION>]`.
```
lwshang: This points users to `icp identity login`, but the implemented subcommand is `reauth`. Looks like a leftover from the rename — should be `icp identity reauth`.

crates/icp-cli/src/commands/identity/reauth.rs
```diff
@@ -1,80 +1,132 @@
...
+            info!("Identity `{}` re-authenticated", args.name);
+        }
+
+        IdentitySpec::Pem {
+            format: PemFormat::Pbes2,
+            algorithm,
+            ..
+        } => {
+            let duration = match &args.duration {
+                Some(d) => Duration::from_nanos(d.as_nanos()) + Duration::from_secs(5 * 60),
```
lwshang: Should the boost values be consistent here? This explicit path adds 5 minutes, while the automatic path below (`(m + 2) * 60`), this arg's doc comment, the settings docs, and `create_explicit_pem_session`'s doc all use 2 minutes.

crates/icp/src/identity/key.rs
```diff
@@ -212,17 +230,28 @@ fn load_pem_identity(
...
+        PemFormat::Pbes2 => load_pbes2_identity(&doc, algorithm, password_func, &origin)?,
+        PemFormat::Plaintext => load_plaintext_identity(&doc, algorithm, &origin)?,
+    };
 
-        PemFormat::Plaintext => load_plaintext_identity(&doc, algorithm, &origin),
+    // After unlocking a Pbes2 PEM, create and cache a short-lived session delegation.
+    if *format == PemFormat::Pbes2
+        && let Some(duration) = pem_session_duration
+    {
+        match create_pem_session_and_build_identity(dirs, name, &*identity, duration) {
```
lwshang, line 242: `load_pem_identity` runs inside `Loader::load`'s `with_read` (shared lock), but `create_pem_session_and_build_identity` writes the keyring entry and chain file. Two concurrent `icp` processes can interleave these writes and leave the keyring key and on-disk chain mismatched (chain delegates to key A, keyring holds key B → signatures rejected until expiry). Consider doing the create+persist under a write lock.

crates/icp/src/identity/key.rs
```diff
@@ -376,8 +655,8 @@ fn load_webauth_identity(
     let stored_chain =
         delegation::load(&chain_path).context(LoadDelegationChainSnafu { path: &chain_path })?;
 
-    // Check expiry (5 minutes grace)
-    if delegation::is_expiring_soon(&stored_chain, FIVE_MINUTES_NANOS)
+    // Check expiry (2 minutes grace)
+    if delegation::is_expiring_soon(&stored_chain, TWO_MINUTES_NANOS)
```
lwshang, line 663: Web-auth's expiry grace was 5 min before this PR and is now 2 min via `TWO_MINUTES_NANOS` — was reducing the web-auth grace intentional, or should it stay 5 min?

crates/icp/src/identity/key.rs
```diff
@@ -292,13 +321,263 @@ const SERVICE_NAME: &str = "icp-cli";
...
+
+    let now_nanos = u64::try_from(
+        SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("system clock before unix epoch")
+            .as_nanos(),
+    )
+    .expect("time wrapped around");
+    let expiration =
+        now_nanos.saturating_add(duration.as_nanos().try_into().expect("time wrapped around"));
```
lwshang: `as_nanos()` (u128) → `try_into()` u64 `.expect(...)` can panic on a large `session-length` (the settings parser accepts up to ~u32::MAX minutes, which exceeds u64 nanoseconds). Prefer saturating/bounding over a panic; the "time wrapped around" message is also misleading here.

lwshang (commented):
Reviewed the automatic delegation sessions change — solid and well-tested. A few inline notes below.

crates/icp/src/identity/key.rs
adamspofford-dfinity: Intentional. This only exists for drift, and the generally assumed/permitted drift is one minute - five minutes was an incorrectly used magic number, and this strikes a better balance.

crates/icp/src/identity/key.rs
adamspofford-dfinity: Saturating would produce a silently incorrect value. (I also don't see how that's misleading.) >u64::MAX is an error case that the tool cannot handle.

lwshang (approved):

PR #530, opened 2026-04-24
feat: Project bundling

adamspofford-dfinity:
This PR adds the `icp project bundle` command, creating a project bundle. A bundle contains built canister WASMs, plugins, input directories and files to the plugins, and a rewritten manifest that references the canisters as prebuilt local files. The command is hidden while the feature is in beta. A bunch of schemas are updated for serialization, required for rewriting the manifest.


crates/icp-cli/src/operations/bundle.rs
```diff
@@ -0,0 +1,815 @@
...
+    )
+    .await
+    .context(ResolvePluginSnafu {
+        canister: canister.name.clone(),
+    })?;
+
+    let plugin_bytes = fs::read(&resolved).context(ReadPluginSnafu {
+        canister: canister.name.clone(),
+    })?;
+    let plugin_sha256 = hex::encode(Sha256::digest(&plugin_bytes));
```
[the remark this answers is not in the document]
adamspofford-dfinity: The relative path is not used in the bundle, so there is nothing nonportable about it. Plugins do not play the same role as plugin inputs and do not make sense to forbid having in parent directories.

crates/icp/Cargo.toml
```diff
@@ -39,6 +39,7 @@ ic-utils = { workspace = true }
 icp-canister-interfaces = { workspace = true }
 icp-sync-plugin = { workspace = true }
 icrc-ledger-types = { workspace = true }
+indexmap = { workspace = true }
```
raymondk, line 42: Why is this needed? Do we care about the order at all?

raymondk: for reproducibility perhaps?

adamspofford-dfinity: Yes, this was in response to a Copilot comment about that.

lwshang (approved):
LGTM — careful path handling throughout. Approving for the beta gate. A few non-blocking things to look at before we un-hide it:

1. **Reproducibility** (`crates/icp-cli/src/operations/bundle.rs:681-683` and `append_dir` at `:745-755`): `append_dir_all` walks `read_dir` unsorted, so archives aren't byte-reproducible across machines despite the `HeaderMode::Deterministic` comment. Either sort entries or reword to "metadata-normalized."
2. **Test gaps** (`crates/icp-cli/tests/bundle_tests.rs`): the validation branches `validate_network_for_bundle` (`bundle.rs:857`, `AbsoluteBindMount`), `CanisterNameCollision` (`bundle.rs:772-784`), and `validate_output_path` (`bundle.rs:843`, `OutputOverlapsSyncDir`) have no coverage — cheap to add, easy to break in a refactor.
3. **Minor edges (low priority):**
   - Cross-namespace archive collisions — e.g. a canister named `plugins`/`init-args` overlaps the prefixes built in `prepare_plugin_step` (`bundle.rs:501`) and `inline_environments` (`bundle.rs:642-650`); only canister *names* are deduped (`bundle.rs:772-784`).
   - Duplicate `init-args/{canister}/{path}` across environments isn't guarded (`bundle.rs:642-650`).
   - `unreachable!()` at `bundle.rs:416` relies on an invariant set in `validate_canisters` — returning the error would be safer than a panic.

PR #536, opened 2026-04-27
feat: Add `--of-principal` to `icp token/cycles balance`

adamspofford-dfinity:


crates/icp-cli/src/operations/token/balance.rs
raymondk, line 57: isn't it simpler if `get_balance` takes:
- principal
- Option<subaccount>
- token

And you get the balance for that principal.
Seems like a clearer api than maybe maybe of_principal otherwise the principal of the agent.

adamspofford-dfinity: No, the agent should typically be an input to functions like this. The reason is that we do not want to create more than one HTTP client for keep-alive reasons.

raymondk: Yes - I didn't mean to drop the agent.
I meant 

```suggestion
pub async fn get_balance(
   agent: &Agent,
   of_principal: Principal,
   subaccount: Option<[u8; 32]>,
   token: &str,
) -> Result<TokenAmount, GetBalanceError> {
```

This way you're always getting the balance `of_principal` not falling back on the principal in the agent and it's clearer what the function is doing.

Otherwise - at the very least add a comment to explain the arguments

raymondk (approved):

PR #545, opened 2026-05-02
icp-sync-plugin/build.rs: warn instead of panic when wasm32-wasip2 is missing

### LINKED ISSUE #543: build.rs: skip wasm32-wasip2 test fixture build for non-test builds
## Problem

`crates/icp-sync-plugin/build.rs` unconditionally calls `build_test_fixture()`, which runs `cargo build --target wasm32-wasip2` to compile a test fixture. This fails when the `wasm32-wasip2` target is not installed:

```
error[E0463]: can't find crate for `core`
  = note: the `wasm32-wasip2` target may not be installed

thread 'main' panicked at crates/icp-sync-plugin/build.rs:33:5:
cargo build --target wasm32-wasip2 failed for test fixture
```

The `TEST_PLUGIN_WASM` env var set by this build step is only used in `#[test]` functions in `runtime.rs`, so the fixture is not needed for regular `cargo build` or `cargo install`.

## Suggested fix

Gate `build_test_fixture()` behind a check so it only runs during test builds, e.g.:

```rust
fn main() {
    println!("cargo:rerun-if-changed=sync-plugin.wit");
    println!("cargo:rerun-if-changed=tests/fixtures/test-plugin/src/lib.rs");
    println!("cargo:rerun-if-changed=tests/fixtures/test-plugin/Cargo.toml");

    // Only build the wasm test fixture when running tests
    if std::env::var("CARGO_CFG_TEST").is_ok() {
        build_test_fixture();
    }
}
```

## Context

This came up while packaging icp-cli 0.2.6 for Homebrew (https://github.com/Homebrew/homebrew-core/pull/280458). The Homebrew Rust toolchain does not include the `wasm32-wasip2` target. We are currently working around this with an `inreplace` in the formula.

SAY-5:
Fixes #543.

`crates/icp-sync-plugin/build.rs` unconditionally ran `cargo build --target wasm32-wasip2` for the test fixture and panicked the entire build when the target wasn't installed. Distribution toolchains (Homebrew, distro packagers) often don't ship `wasm32-wasip2`, so `cargo install icp-cli` failed for downstream packagers.

The fixture is only consumed by `#[test]` functions in `runtime.rs`. `CARGO_CFG_TEST` isn't set during build-script execution, so suggested gating doesn't actually distinguish test from non-test invocations. Instead: keep building when the target is available, but on failure emit a `cargo:warning` and set `TEST_PLUGIN_WASM=` (empty). Tests that need the fixture will surface a clearer error at test time; `cargo install` and `cargo build` succeed cleanly on toolchains without `wasm32-wasip2` (e.g. the Homebrew Rust at https://github.com/Homebrew/homebrew-core/pull/280458).

### DIFF
--- a/crates/icp-sync-plugin/build.rs
+++ b/crates/icp-sync-plugin/build.rs
@@ -28,12 +28,29 @@ fn build_test_fixture() {
             "--target-dir",
             fixture_target_dir.as_str(),
         ])
-        .status()
-        .expect("failed to spawn cargo build for test fixture");
-    assert!(
-        status.success(),
-        "cargo build --target wasm32-wasip2 failed for test fixture"
-    );
+        .status();
+
+    // Building the wasm test fixture requires the `wasm32-wasip2` target to be
+    // installed. Distribution toolchains (Homebrew, distro packagers — see
+    // dfinity/icp-cli#543) do not always include it, and the fixture is only
+    // referenced by `#[test]` functions in `runtime.rs`. Emit a warning and
+    // leave `TEST_PLUGIN_WASM` empty rather than panicking the whole build;
+    // any test that actually depends on the fixture will surface a clearer
+    // "fixture not built" error when the empty path is used.
+    let succeeded = matches!(&status, Ok(s) if s.success());
+    if !succeeded {
+        match status {
+            Err(e) => println!(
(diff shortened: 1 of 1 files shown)


lwshang:
Thank you @SAY-5 for the fix and the thorough analysis — we really appreciate it!

We landed on a similar approach in #544, which was just merged. Closing this to avoid a double-merge. Thanks again!

PR #558, opened 2026-05-18
feat(canister-install): add --wasm-memory-persistence flag

timohanke:
`icp canister install -m upgrade` now offers the `--wasm-memory-persistence keep|replace` option where `keep` is the default.

The option isn't available in any other install mode (install, reinstall) and is only available for EOP canisters.

When `replace` is specified then another interactive y/N confirmation is required. It can be overridden with -y/--yes.


raymondk:
@timohanke from the end user perspective, how is this different than doing a reinstall?
If you pick `replace` don't you effectively do a reinstall and wipe out the full state?

lwshang:
Closing in favor of #599, which rebased this work onto `main` (as a single squashed commit) and added integration test coverage. #599 was merged in 404d7c0. Thanks @timohanke for the original implementation!

PR #567, opened 2026-05-26
feat: inject `_.canister.name` into recipe template context

marc0olo:
## Summary

- Adds `{{_.canister.name}}` as a built-in recipe variable, automatically injected into every Handlebars recipe template from the canister name in `icp.yaml`
- The `_` namespace is reserved and cannot be overridden by user-provided `configuration:` values
- Works with all Handlebars helpers, e.g. `{{ replace "-" "_" _.canister.name }}` for Rust WASM artifact filenames
- Updates `docs/concepts/recipes.md` and `docs/guides/creating-recipes.md` to document built-in recipe variables and use `{{_.canister.name}}` throughout all examples
- `docs/guides/using-recipes.md` will be updated separately once the official recipes (`@dfinity/rust`, `@dfinity/motoko`) are updated to use `{{_.canister.name}}` in icp-cli-recipes


crates/icp/src/lib.rs
```diff
@@ -622,7 +622,11 @@ mod tests {
 
     #[async_trait]
     impl Resolve for MockRecipeResolver {
-        async fn resolve(&self, _recipe: &Recipe) -> Result<(BuildSteps, SyncSteps), ResolveError> {
+        async fn resolve(
+            &self,
+            _recipe: &Recipe,
+            _canister_name: &str,
```
raymondk: I would suggest making this a struct because we might want to add more context than just the name.

docs/concepts/recipes.md
```diff
@@ -111,8 +111,26 @@ build:
...
+    shrink: true   # available as {{ shrink }} in the template
+```
+
+**Built-in recipe variables** — automatically provided by icp-cli for every recipe, regardless of what the user puts in `configuration:`:
+
+| Variable | Value |
+|---|---|
+| `{{icp.canister_name}}` | The canister name as defined in `icp.yaml` |
+
+The `icp` namespace is reserved and cannot be overridden by user-provided configuration.
```
raymondk: I think we need a namespace but `icp` doesn't make sense here.
We could something more generic like:

```
_. project.name     # Not sure if that would be used
_. canister.name    # The canister being rendered
```

I don't think the `_` is reserved for anything but it seems like more of a convention.

crates/icp/src/canister/recipe/mod.rs
```diff
@@ -13,7 +13,11 @@ pub mod handlebars;
 #[async_trait]
 pub trait Resolve: Sync + Send {
     #[allow(clippy::result_large_err)]
-    async fn resolve(&self, recipe: &Recipe) -> Result<(BuildSteps, SyncSteps), ResolveError>;
+    async fn resolve(
+        &self,
+        recipe: &Recipe,
+        canister_name: &str,
```
raymondk: This is where we should send a context instead of a name.

raymondk (dismissed):

crates/icp/src/canister/recipe/handlebars.rs
```diff
@@ -186,13 +198,24 @@ impl Handlebars {
...
+        // The _ key is reserved and always overrides any user-supplied value.
+        let mut context = recipe.configuration.clone();
+        context.insert(
+            "_".to_string(),
+            serde_yaml::to_value(InjectedContext {
+                canister: CanisterContext {
+                    name: canister_name,
+                },
+            })
+            .expect("InjectedContext serialization is infallible"),
```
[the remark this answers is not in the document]
marc0olo: please acknowledge following guideline according to `Claude.md`:

> Don't add error handling, fallbacks, or validation for scenarios that can't happen. Trust internal code and framework guarantees. Only validate at system boundaries (user input, external APIs).

lwshang (approved):
LGTM

**Suggestion: make a user-supplied `_` a hard error**

The override protection is correct, but silently dropping a user's `_` value hides a real mistake. Since `_` is reserved, supplying it is never valid — I'd reject it explicitly rather than swallow it:

```rust
if recipe.configuration.contains_key("_") {
    return Err(/* ReservedNamespace error */);
}
```

Fails fast with a clear message instead of leaving the user wondering why their value vanished. The existing test would just flip to asserting the error.

raymondk:
> LGTM
> 
> **Suggestion: make a user-supplied `_` a hard error**
> 
> The override protection is correct, but silently dropping a user's `_` value hides a real mistake. Since `_` is reserved, supplying it is never valid — I'd reject it explicitly rather than swallow it:
> 
> ```rust
> if recipe.configuration.contains_key("_") {
>     return Err(/* ReservedNamespace error */);
> }
> ```
> 
> Fails fast with a clear message instead of leaving the user wondering why their value vanished. The existing test would just flip to asserting the error.

That's a good suggestion - we can add that in a different change.

PR #575, opened 2026-05-28
feat(identity): receive II delegation via form POST and redirect back with status

sea-snake:
## Motivation

Chrome's Local Network Access gates public-origin → loopback subresource requests behind a permission prompt next to the URL bar. The II frontend used to POST the delegation to the CLI's loopback callback with `fetch`, which triggers that prompt. Switching the delivery to a top-level form-POST navigation avoids the prompt, and having the loopback server redirect the browser back to the II frontend keeps id.ai in control of the success/error UI.

That top-level navigation, however, posts from the https login page to the http loopback callback — an https→http downgrade. Under the browser's default `strict-origin-when-cross-origin` referrer policy, a non-CORS navigation has its `Origin` serialized to `null` on a scheme downgrade, so the original exact-match `Origin` check rejected every legitimate POST with 403. The check is replaced with a nonce that doesn't depend on browser `Origin` behavior.

## Command

```
icp identity link ii <name> [--auth <auth domain>] [--app <delegation domain>]
```

- `--auth` — the domain to sign in at. Defaults to `id.ai`; its `/.well-known/cli-auth-config` decides the login path (so `id.ai` → `id.ai/cli`). Use it to point at another II deployment or a custom login page (e.g. `--auth identity.ce1.com`, `--auth special.com`).
- `--app` — the delegation domain to get an identity for (e.g. `--app oisy.com`), emitted in the fragment as `domain`. Omitted, the auth page picks its own default (id.ai uses `cli.id.ai`), so a bare `icp identity link ii <name>` creates a portable id.ai/cli identity.

## Changes

- The loopback callback now accepts a top-level form POST (field `delegation`) instead of a JSON `fetch`, and 303-redirects the browser back to `{auth}{login_path}#status=...`:
  - `success`: the delegation is accepted; the server returns the chain and shuts down.
  - `identity-mismatch` (only on `icp identity login` re-auth): the delegation's self-authenticating principal does not match the linked identity. The server keeps listening (the redirect re-supplies `public_key`/`callback`/`nonce`/`domain` so the user can retry in place) and the CLI prints a message so it is not silent.
  - `error`: malformed delegation; the server shuts down, the redirect carries only the status, and the CLI exits with an error.
- Renames `--host` to `--auth` and adds `--app`. The login fragment now carries `public_key`/`callback`/`nonce`/`domain` (`domain` only when `--app` is given).
- The auth domain and the delegation domain are persisted in the identity manifest, so `icp identity login` re-authenticates against the same auth page and the same delegation domain — resolving the same principal — without re-passing the flags. (`domain` is serde-defaulted for back-compat with already-linked identities.)
- `recv_delegation` gains `delegation_domain` and `expected_principal` parameters: `login` passes the stored principal/domain, `link` passes `None` for the principal (it creates a brand-new identity).
- Discovery reads the `{ "path": "..." }` JSON at `/.well-known/cli-auth-config` to match the II frontend's contract.
- Drops the CORS preflight handling (a top-level navigation needs none).
- **Gates the callback with a single-use nonce instead of `Origin`.** The CLI generates a 32-byte base64url nonce, puts it in the login URL fragment, and requires the frontend to echo it back in the POST body; a missing or mismatched nonce is rejected with 403 without ending the flow. The fragment is never sent over the network and is unreadable cross-origin, so only the page the user logged in through can produce it — the same anti-abort property the `Origin` check was meant to provide, but immune to the https→http downgrade that made `Origin` `null`.
- **Console output** mirrors the id.ai/cli terminal block: after the "Press Enter to log in" prompt it shows a green `✓ Browser opened` line and an animated `Linking Internet Identity` spinner that finishes with a green check, instead of going silent.

Pairs with the II frontend change in dfinity/internet-identity#3956 (form POST + status parsing + success/mismatch/error screens + nonce echo + `domain` derivation).

### DIFF
--- a/.claude/rules/team-general
+++ b/.claude/rules/team-general
@@ -1,1 +1,0 @@
-/home/claude/.identity-team-skills/rules/general
--- a/.claude/rules/team-general
+++ b/.claude/rules/team-general
@@ -1,0 +1,1 @@
+/home/claude/.identity-team-skills/rules/general
--- a/crates/icp-cli/src/commands/identity/link/ii.rs
+++ b/crates/icp-cli/src/commands/identity/link/ii.rs
@@ -1,10 +1,10 @@
 use std::net::SocketAddr;
 
 use axum::{
-    Router,
+    Form, Router,
     extract::State,
     http::{HeaderMap, HeaderValue, StatusCode, header},
-    response::IntoResponse,
+    response::{IntoResponse, Redirect},
     routing::post,
 };
 use base64::engine::{Engine as _, general_purpose::URL_SAFE_NO_PAD};
@@ -23,6 +23,7 @@ use icp::{
     prelude::*,
 };
 use indicatif::{ProgressBar, ProgressStyle};
+use serde::Deserialize;
 use snafu::{OptionExt, ResultExt, Snafu, ensure};
 use tokio::{net::TcpListener, sync::oneshot};
 use tracing::{info, warn};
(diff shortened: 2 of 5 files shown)


adamspofford-dfinity (approved):

raymondk:
@adamspofford-dfinity can we merge this and update the docs and cli-ref in a separate PR?

PR #583, opened 2026-06-04
feat: Rename ii to web and unhide commands

adamspofford-dfinity:


raymondk:
@adamspofford-dfinity can you add an entry into the changelog 🙏

raymondk (approved):

PR #600, opened 2026-06-11
feat: Change default mainnet gateway url to icp.net

adamspofford-dfinity:


raymondk:
We don't need to handle the rewrite anymore - it's going to be handled in II.
We can change this to only use icp.net

crates/icp-cli/src/commands/identity/link/web.rs
```diff
@@ -512,3 +505,110 @@ async fn handle_callback(
     finish(&state, Ok(chain));
     Redirect::to(&state.success_url).into_response()
 }
+
+fn guesstimate_origin(maybe_origin: &str) -> String {
+    let hierarchy = maybe_origin
```
aterga: Nit. Why not use a proper URL library?

aterga (approved):

crates/icp-cli/src/commands/identity/link/web.rs
adamspofford-dfinity: We expect it to arrive as a domain, URL part stripping is only to be helpful. And the `url` library doesn't then give you domain components.

raymondk (approved):

PR #614, opened 2026-06-18
fix(deploy,sync): ensure canister is Running before asset sync

lwshang:
## Problem

`icp deploy`/`icp sync` against the asset canister intermittently failed at the sync step, when the sync plugin's first canister call reached the target while it was not serving:

```
... Canister <id> is stopped and therefore does not have a CallContextManager, error code Some("IC0508")
```

Investigation found **two distinct failure modes**:

- **Mode A — canister genuinely not Running before sync.** `install_code` is status-preserving, so a canister that *enters* the command already `Stopped`/`Stopping` (handed out Stopped from a canister pool, or left Stopped by a deploy that died in the stop→install→start window) is installed but never started, and the plugin's first call hits IC0508. **Persistent** — reruns don't help. (Why developer-docs preview deploys failed repeatedly.)
- **Mode B — query path transiently stale right after a restart.** When the canister *was* Running, deploy correctly does stop→upgrade→start, but `start_canister` only makes it Running in the subnet's *certified* state. IC query calls are eventually-consistent reads served by a single replica that may still lag the restart's commit height; the plugin's first calls are queries, so they can momentarily observe the just-vacated Stopped state. **Transient** — a rerun "fixes" it. (Why the always-running docs canister failed intermittently.)

## Fix

**Mode A — commit `e0797fe3`:**
- `icp deploy` owns the lifecycle (it just installed): after install and before sync it **starts** every sync-target canister, then leaves it Running. `start_canister` is idempotent and synchronous per the IC spec, so no status poll is needed. Scoped to canisters with sync steps.
- `icp sync` does NOT own the lifecycle: it **prechecks** run status and **aborts with an actionable error** if not Running, replacing the cryptic IC0508. It does not auto-start.
- `install.rs` is deliberately unchanged — it stays status-preserving.

**Mode B — commit `eb367aa2`:**
- In `icp deploy`, after starting the canister and before sync, **wait until the query path consistently observes the canister Running**. Probe with a query for a method no canister exports and classify the reject: "is stopped/stopping" (IC0508/IC0509) means the replica still lags; any other reject (e.g. "no query method") or a reply means it is serving. Require a few consecutive ready observations, spaced so they may land on different replicas.
- Confidence-raising, **not** a hard guarantee — query reads are per-node and boundary nodes load-balance across replicas, an inherent property of the decentralized read path — but it makes the post-restart race rare.

## Tests

- `deploy_starts_stopped_canister_before_sync` — deploy, stop, redeploy; deploy starts the canister before the plugin runs and ends Running with synced data. Reproduces Mode A pre-fix.
- `sync_aborts_when_canister_not_running` — `icp sync` against a stopped canister aborts with the actionable message; the sync step never runs.

Mode B is a timing/replica-lag race that a single-node local replica doesn't exhibit, so it has no e2e test. `cargo build`, `cargo clippy --tests`, `cargo fmt --check` clean.

## Out of scope (follow-ups)
- Clearer plugin-side error than raw IC0508.
- Whether `resolve_install_mode_and_status` should repair a `Stopping` canister.

### DIFF
--- a/crates/icp-cli/src/commands/deploy.rs
+++ b/crates/icp-cli/src/commands/deploy.rs
@@ -487,38 +487,47 @@ async fn wait_until_serving_queries(
     canister_id: Principal,
 ) -> Result<(), anyhow::Error> {
     const REQUIRED_CONSECUTIVE: u32 = 2;
-    const MAX_ATTEMPTS: u32 = 60;
+    // Total wall-clock budget for the whole wait — the hard cap on the failure
+    // path. PROBE_TIMEOUT below only bounds a single hung probe (so retries keep
+    // flowing); this outer budget is what guarantees we give up promptly, rather
+    // than attempts * (probe timeout + interval).
+    const READINESS_BUDGET: Duration = Duration::from_secs(30);
     const POLL_INTERVAL: Duration = Duration::from_millis(500);
-    const PROBE_TIMEOUT: Duration = Duration::from_secs(10);
+    const PROBE_TIMEOUT: Duration = Duration::from_secs(2);
 
-    let mut consecutive_ready: u32 = 0;
-    for _ in 0..MAX_ATTEMPTS {
-        let probe = agent
-            .query(&canister_id, READINESS_PROBE_METHOD)
-            .with_arg(Vec::<u8>::new())
-            .call();
-        let ready = match tokio::time::timeout(PROBE_TIMEOUT, probe).await {
-            Ok(Ok(_)) => true,                        // replied -> Running
-            Ok(Err(err)) => !is_stopped_reject(&err), // reject: ready unless "stopped"
-            Err(_elapsed) => false,                   // probe timed out -> inconclusive
-        };
+    let poll = async {
+        let mut consecutive_ready: u32 = 0;
+        loop {
+            let probe = agent
+                .query(&canister_id, READINESS_PROBE_METHOD)
+                .with_arg(Vec::<u8>::new())
+                .call();
+            let ready = match tokio::time::timeout(PROBE_TIMEOUT, probe).await {
+                Ok(Ok(_)) => true,                        // replied -> Running
+                Ok(Err(err)) => !is_stopped_reject(&err), // reject: ready unless "stopped"
+                Err(_elapsed) => false,                   // probe timed out -> inconclusive
+            };
 
-        if ready {
-            consecutive_ready += 1;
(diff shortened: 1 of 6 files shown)


crates/icp-cli/src/commands/deploy.rs
```diff
@@ -414,6 +455,89 @@ pub(crate) async fn exec(ctx: &Context, args: &DeployArgs) -> Result<(), anyhow:
...
+    const POLL_INTERVAL: Duration = Duration::from_millis(500);
+    const PROBE_TIMEOUT: Duration = Duration::from_secs(10);
+
+    let mut consecutive_ready: u32 = 0;
+    for _ in 0..MAX_ATTEMPTS {
+        let probe = agent
+            .query(&canister_id, READINESS_PROBE_METHOD)
+            .with_arg(Vec::<u8>::new())
+            .call();
+        let ready = match tokio::time::timeout(PROBE_TIMEOUT, probe).await {
```
raymondk: isn't 10s for the timeout too high?
If there's something wrong - we're going to end up waiting 60*10.5 seconds for failure.

lwshang: Fixed in 482bdef7: wrapped the whole wait in a single 30s budget (hard cap) and dropped the per-probe timeout to 2s. Failure path is now ≤30s.

crates/icp-cli/src/commands/deploy.rs
```diff
@@ -414,6 +455,98 @@ pub(crate) async fn exec(ctx: &Context, args: &DeployArgs) -> Result<(), anyhow:
...
+        loop {
+            let probe = agent
+                .query(&canister_id, READINESS_PROBE_METHOD)
+                .with_arg(Vec::<u8>::new())
+                .call();
+            let ready = match tokio::time::timeout(PROBE_TIMEOUT, probe).await {
+                Ok(Ok(_)) => true,                        // replied -> Running
+                Ok(Err(err)) => !is_stopped_reject(&err), // reject: ready unless "stopped"
+                Err(_elapsed) => false,                   // probe timed out -> inconclusive
+            };
```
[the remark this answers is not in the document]
lwshang: Fixed in 2bfc742e: ready now requires a reply or a genuine non-stopped reject; all non-reject AgentErrors are inconclusive and retried.

raymondk (approved):

PR #639, opened 2026-07-06
feat: make root key optional for connected networks

lwshang:
## Summary

Makes the root key **optional** for connected networks, resolved by trust scope when the network is accessed:

| Case | Behavior |
|---|---|
| Explicit key (manifest `root-key` / `--root-key`) | used as-is (any network) |
| Omitted + **loopback** URL | fetched from the network's status endpoint |
| Omitted + **remote** URL | hard error, fail fast |

The built-in `ic` network is unchanged — still the hardcoded mainnet key, non-overridable.

## Motivation

Previously `--network <URL>` **required** `--root-key`, and an omitted manifest key silently defaulted to the mainnet key. A local (loopback) network has no stable key to hardcode, so connecting to one by URL meant pasting a 266-char hex key.

- **Fetching on loopback is safe** — there's no meaningful man-in-the-middle surface on localhost.
- **Refusing on remote** surfaces a clear, actionable error at connection time (pass `--root-key`, or use `--network ic` for mainnet) instead of a cryptic certificate-verification failure on the first certified call. A genuine remote network never shares mainnet's key, so the old mainnet-default only deferred the failure — this is a diagnostics improvement, not a security change (both are fail-safe).

## Implementation

- `Connected.root_key`: `Vec<u8>` → `Option<Vec<u8>>` (runtime type only; the manifest field was already optional). The manifest→runtime conversion no longer defaults to the mainnet key.
- Resolution happens in one place — `get_connected_network_access` → `resolve_root_key` (`crates/icp/src/network/access.rs`) — which loopback-checks the URL, then fetches (anonymous agent → `fetch_root_key` → `read_root_key`) or fails with `RootKeyRequiredForRemote`. Runs before any RPC, so it covers both the CLI and manifest paths. `NetworkAccess.root_key` stays `Vec<u8>`, so identity/delegation and `set_root_key` downstream are untouched.
- CLI: `--network <URL>` no longer forces `--root-key`; a named network still rejects it.

## Notable consequence

Connecting to **actual mainnet by raw URL** without a key now errors instead of silently working — the message points users to `--network ic` or `--root-key`.

## Testing

- Added a no-`--root-key` call against a running local network to `canister_call_with_url_and_root_key` (exercises the loopback-fetch path end-to-end).
- Added `canister_call_remote_url_without_root_key_errors` (fast, no live server — the loopback check short-circuits before any request).
- `cargo fmt --check` clean, `cargo clippy` zero warnings, full-workspace `cargo test --no-run` compiles.
- Config schema regenerated → no diff. CLI docs (`docs/reference/cli.md`) regenerated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ bump. Currently experimental: project bundling
 # Unreleased
 
 * feat: `icp canister delete` will now send the canister's remaining cycles to the caller
+* feat: The root key of a `connected` network is now optional. When you connect to a local (loopback) network by URL, the root key is fetched from the network automatically, so `--network <URL>` no longer requires `--root-key`. A remote network must still provide its root key (via the manifest `root-key` or `--root-key`); previously an omitted key silently defaulted to the mainnet key, which now surfaces a clear error instead. The built-in `ic` network is unchanged.
 
 # v1.0.2
 
--- a/crates/icp-cli/src/options.rs
+++ b/crates/icp-cli/src/options.rs
@@ -99,8 +99,9 @@ pub(crate) struct NetworkOptInner {
     #[arg(long, short = 'n', env = "ICP_NETWORK", group = "network-select", help_heading = heading::NETWORK_PARAMETERS, value_parser = parse_network_target)]
     network: Option<NetworkTarget>,
 
-    /// The root key to use if connecting to a network by URL.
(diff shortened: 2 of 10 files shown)


raymondk:
Closing in favor of: https://github.com/dfinity/icp-cli/pull/645

PR #652, opened 2026-07-15
feat: Add `--no-create` flag to `icp deploy`

adamspofford-dfinity:


marc0olo:
This looks solid overall — clean implementation, good test coverage (both the missing and already-exists paths), and CHANGELOG + CLI docs are updated. 👍

Leaving final approval to Linwei/Raymond since this introduces a new flag.

A few optional, non-blocking nits:

1. `Creating canisters:` is logged unconditionally at [`deploy.rs:181`](https://github.com/dfinity/icp-cli/blob/spofford/flag-no-create/crates/icp-cli/src/commands/deploy.rs#L181), so the `--no-create` failure prints `Creating canisters:` immediately followed by the "will not create" error, which reads a little oddly. Could move that log into the `else` create branch.
2. Suppressing the `All canisters already exist` message under `--no-create` is a minor special case — deploy proceeds identically either way when everything exists, so leaving it unconditional would be simpler.
3. Not a request to change here, just noting: the build runs before the `--no-create` check, so the flag doesn't fail-fast before building. Fine given the check needs the env/agent/ids.

None of these block merge.

lwshang:
- Reworded the phase header "Creating canisters:" → "Ensuring canisters exist:". The old header over-claimed — it printed even when nothing would be created, reading oddly ahead of both "All canisters already exist" and the new `--no-create` error. The new wording reads naturally everywhere, and "All canisters already exist" now prints unconditionally again.
- Made `--no-create` conflict with `--subnet`/`--cycles` at parse time (with a test). They only affect creation, so combined with `--no-create` they were silently ignored — now clap rejects the combo up front. `--proxy` is left out since it's used beyond creation.

lwshang (approved):

PR #678, opened 2026-07-27
fix: Identify broken network processes

### LINKED ISSUE #577: Local network reports "already running" after host sleep, but  PocketIC is defunct and gateway is unreachable.
```markdown
   ## Problem

   After the macOS host sleeps and wakes, a local `icp-cli` network running inside an OrbStack Linux machine becomes unusable.
  I know that after "icp network stop," then "icp network start -d" it will be okay, but this will result in the loss of all data generated after the last deploy.  and i can't let my laptop always on.

   `icp-cli-network-launcher` is still running, so `icp network
 start -d` reports:

   ```text
   Error: network 'local' is already running
 ```

 But the actual gateway is unreachable:

 ```bash
   curl http://127.0.0.1:8000/api/v2/status
   # connection refused
 ```

 Process state shows:

 ```text
   icp-cli-network-launcher ... --gateway-port 8000 ...
   [pocket-ic] <defunct>
 ```

 Then icp deploy fails with:

 ```text
   direct update call failed: An error happened during
 communication with the replica:
   error sending request for url
 (http://127.0.0.1:8000/api/v4/canister/.../call)
 ```

 Environment

 - icp-cli: icp 0.2.7
 - Host: macOS
 - Runtime: OrbStack Linux machine
 - Network: local native managed network
 - Gateway: 127.0.0.1:8000

 Steps to reproduce

 ```bash
   icp network start -d
   icp deploy
   # put macOS host to sleep
   # wake host
   icp deploy
 ```

adamspofford-dfinity:
Fixes #577.

### DIFF
--- a/crates/icp-cli/src/commands/network/start.rs
+++ b/crates/icp-cli/src/commands/network/start.rs
@@ -9,9 +9,12 @@ use icp::{
     identity::manifest::IdentityList,
     network::{
         Configuration,
-        managed::cache::{
-            check_launcher_update_available, download_launcher_version,
-            get_cached_launcher_version_if_fresh,
+        managed::{
+            cache::{
+                check_launcher_update_available, download_launcher_version,
+                get_cached_launcher_version_if_fresh,
+            },
+            run::stop_network,
         },
         run_network,
     },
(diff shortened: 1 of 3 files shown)


marc0olo (approved):
LGTM — reviewed the logic (port-freeing via `stop_launcher`, descriptor-write-after-init so booting networks can't be misclassified) and the test. Solid, well-documented, and conventioned.

One small note, assuming it's a conscious decision: `is_responsive()` only maps connect/timeout errors to `Ok(false)`, so any *other* probe error (e.g. a malformed HTTP response, or DNS failure for a non-localhost host) makes `network start` abort rather than fall through to recovery. That's clearly deliberate per the doc comment ("leaves the question unanswered"), just flagging it since the whole point of the feature is to recover from a wedged launcher. Happy either way.

Issue #597, opened 2026-06-09
icp network start silently discards launcher stderr on port conflict


marc0olo:
When the network launcher exits prematurely (e.g. port 8000 already in use by Docker), the error surfaced to the user is just `exited prematurely with status 101` with no actionable detail.

There are two separate problems:

- **Background mode (`-d`):** stderr is written to `.icp/cache/networks/local/network-launcher/stderr.log` but the path is never shown to the user.
- **Foreground non-verbose mode (default):** stderr is piped to `/dev/null` — the error is completely lost, not even written to a file.

The fix is in `crates/icp/src/network/managed/launcher.rs`: always capture stderr to the log file (including in non-background mode), and when `LauncherExitedPrematurely` fires, read the file tail and include it in the error message.

Reported by a community member in dfinity/examples#1356, who worked around it with a documentation PR.

lwshang:
Fixed by #675 and #676

PR #679, opened 2026-07-28
fix(network): make docker network log behavior match the native launcher

### LINKED ISSUE #677: network start: Docker launcher discards container output on premature exit (same as #597 for native)
## Problem

#597 / #675 / #676 fix the **native** launcher so that when it exits prematurely (e.g. a fixed gateway port is already taken) the user sees the actual cause instead of just `exited prematurely with status 101`.

The **Docker launcher path has the identical bug and is untouched.** When `network start` runs the launcher in a container — always on Windows, and on unix with `autocontainerize=true` or a docker-mode network — it goes through `spawn_docker_launcher`, not `spawn_network_launcher`. On premature exit it returns [`ContainerExitedPrematurely`](https://github.com/dfinity/icp-cli/blob/main/crates/icp/src/network/managed/docker.rs) with only the exit status, and never surfaces the container's output — even though it attaches stdout/stderr (`attach_stdout/attach_stderr: Some(true)`).

## Suggested fix

In the premature-exit arm of the `select!` in `spawn_docker_launcher` (mirroring `premature_exit_detail` from the native path), fetch the container log tail via bollard's `logs()` API and fold it into `ContainerExitedPrematurely`. Note this is a **different mechanism** from the native fix (no local `stderr.log` file; the daemon owns the stdio), so it does not reuse `premature_exit_detail`.

## Notes

- Deliberately scoped out of #676, which is unix/native-only and covered by an integration test that can't exercise the Docker daemon.
- Testing this path needs a working Docker daemon and is harder to make deterministic in CI.

Discovered while reviewing #675 and implementing #676.

lwshang:
## Why

Docker-based networks never showed the launcher's output. Nothing ever attached to the container's stdio, so on a normal start you saw nothing, and on a failed start you got only:

```
Error: failed to watch docker container <id> for exit
Caused by:
    Docker container wait error:
```

On Windows, where the launcher always runs in a container, that meant `icp network start` was silent. #675/#676 fixed this class of bug for the native launcher; this does the same for the container path.

Two root causes on the error path: bollard's `wait_container` turns any nonzero exit into `DockerContainerWaitError` rather than an `Ok` response, so `ContainerExitedPrematurely` was unreachable for exactly the failures that matter — and even when reached, it carried no output, because the daemon owns the container's stdio and there is no local `stderr.log` to read back.

## How

- **Premature exit** — both `wait_container` shapes funnel into `ContainerExitedPrematurely`; the cause is fetched over bollard's `logs()`, bounded by the same 50-line/8&nbsp;KB caps as the native path (`stderr_tail` → shared `output_tail`).
- **Foreground** — a task follows the container and demuxes stdout/stderr onto ours, the equivalent of the native path's `Stdio::inherit()`. `premature_exit_detail` returns early in foreground so the cause isn't printed twice, mirroring the native split; the follower is drained first (2&nbsp;s bound) so its output can't be outrun by the error explaining it.
- **Background** — can't work the same way: a native child keeps filling its log files after the CLI exits because its fds point at them, whereas here nothing is left to pump the daemon-owned stdio. Since the daemon already persists it, `network start -d` prints `docker logs -f <id>` — the counterpart of the native "stderr: `<path>`" message.
- **`--verbose`** — the native path adds it from the CLI's debug flag rather than from `launcher_settings_flags`, so an autocontainerized launcher stayed quiet under `-d`.

Fixes #677. Also closes out #597, whose remaining scope was this path.

## Tests

Two `#[tag(docker)]` tests using a manifest whose container prints a marker to stderr and exits 3. The real launcher can't be made to fail from outside the container (it tolerates unknown arguments, and its ports live in the container's namespace), and what's under test is the log plumbing, not the launcher. The foreground test asserts the `Container output` header is **absent** — that pins the marker as arriving over the live stream, and that the two paths don't both fire.

Successful-start streaming was verified by hand, not in CI; asserting on a healthy foreground network means matching launcher wording, which #676 deliberately avoided.

## Note

Streaming immediately surfaced a launcher bug — every containerized shutdown now ends with `failed to remove status directory / Device or resource busy`, because the launcher tries to `rmdir` its bind-mounted status dir. Pre-existing and previously invisible; icp-cli still exits 0. Filed as dfinity/icp-cli-network-launcher#78.

### DIFF
--- a/crates/icp/src/network/managed/docker.rs
+++ b/crates/icp/src/network/managed/docker.rs
@@ -7,7 +7,7 @@ use bollard::{
     errors::Error as BollardError,
     models::{ContainerCreateBody, HostConfig, Mount, MountTypeEnum, PortBinding},
     query_parameters::{
-        CreateContainerOptions, CreateImageOptions, InspectContainerOptions,
+        CreateContainerOptions, CreateImageOptions, InspectContainerOptions, LogsOptions,
         RemoveContainerOptions, StartContainerOptions, StopContainerOptions, WaitContainerOptions,
     },
 };
@@ -26,7 +26,7 @@ use crate::network::{
 };
 use crate::prelude::*;
 
-use super::launcher::wait_for_launcher_status;
+use super::launcher::{MAX_OUTPUT_TAIL_LINES, output_tail, wait_for_launcher_status};
 
 /// Error converting a path for WSL2.
 #[derive(Debug, Snafu)]
@@ -384,7 +384,7 @@ pub async fn spawn_docker_launcher(
         .await
         .context(CreateContainerSnafu { image_name: image })?;
     let container_id = container_resp.id;
-    info!("Created container {}", &container_id[..12]);
+    info!("Created container {}", short_id(&container_id));
     let guard = AsyncDropper::new(DockerDropGuard {
         container_id: Some(container_id),
         docker: Some(docker),
@@ -401,18 +401,28 @@ pub async fn spawn_docker_launcher(
     let launcher_status = select! {
         content = watcher => content?,
         res = wait_container.try_next() => {
-            let exit = res.context(WatchContainerSnafu { container_id })?;
-            if let Some(exit) = exit {
-                return ContainerExitedPrematurelySnafu {
-                    container_id,
-                    exit_status: exit.status_code,
-                }.fail();
-            } else {
-                return RequiredFieldMissingSnafu {
+            // bollard turns any nonzero exit into `DockerContainerWaitError` instead of an `Ok`
+            // response carrying the status code, so both shapes have to reach the same error.
+            // Otherwise the case that matters — a container that *failed* — surfaced as an
+            // opaque "failed to watch container for exit" with an empty cause.
+            let (exit_status, wait_error) = match res {
+                Ok(Some(exit)) => (
+                    exit.status_code,
+                    exit.error.and_then(|e| e.message).unwrap_or_default(),
+                ),
+                Ok(None) => return RequiredFieldMissingSnafu {
                     field: "StatusCode",
                     route: "wait_container",
-                }.fail()
-            }
+                }.fail(),
+                Err(BollardError::DockerContainerWaitError { code, error }) => (code, error),
+                Err(source) => return Err(source).context(WatchContainerSnafu { container_id }),
+            };
+            let detail = premature_exit_detail(docker, container_id, &wait_error).await;
+            return ContainerExitedPrematurelySnafu {
+                container_id,
+                exit_status,
+                detail,
+            }.fail();
         },
     };
     let container_info = docker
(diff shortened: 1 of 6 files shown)


crates/icp/src/network/managed/docker.rs
```diff
@@ -512,6 +541,142 @@ pub async fn spawn_docker_launcher(
...
+/// premature-exit error is rendered. Bounded, and deliberately ignores both outcomes: the error
+/// is reported either way, at worst without the container's last few lines.
+async fn drain_log_follower(follower: JoinHandle<()>) {
+    if tokio::time::timeout(LOG_DRAIN_TIMEOUT, follower)
+        .await
+        .is_err()
+    {
+        debug!("timed out draining container log output");
+    }
+}
```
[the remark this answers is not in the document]
lwshang: Correct, fixed in 1f324dbb. `timeout` takes the handle by value and dropping a `JoinHandle` detaches rather than cancels, so on timeout the follower really was left free to write after the error.

Awaiting by `&mut` and aborting now, with one adjustment to the suggestion: awaiting cancellation unbounded could hang `network start` if the task is stuck in an in-flight blocking write to stdout, which is worse than the late output it prevents. The abort-await is bounded by `LOG_DRAIN_TIMEOUT` again, so a wedged write degrades to a late error instead.

crates/icp/src/network/managed/docker.rs
```diff
@@ -512,6 +541,142 @@ pub async fn spawn_docker_launcher(
...
+            stdout: true,
+            stderr: true,
+            tail: MAX_OUTPUT_TAIL_LINES.to_string(),
+            ..<_>::default()
+        }),
+    ));
+    let mut output = String::new();
+    while let Some(chunk) = logs.next().await {
+        match chunk {
+            Ok(chunk) => output.push_str(&chunk.to_string()),
```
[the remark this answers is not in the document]
lwshang: Also correct, fixed in 1f324dbb — `tail` caps lines, not their length, so the buffer was unbounded until `output_tail` ran at the end.

Now trimmed to a bounded suffix while consuming the stream (`MAX_LOG_BUFFER_BYTES`, twice the final cap so the line-based trim still has its full budget). The byte cut is factored out of `output_tail` as `tail_bytes` rather than reusing `output_tail` directly: that one trims trailing whitespace, which mid-stream would glue the next chunk onto the end of the current line.

Verified against the scenario you describe — a container emitting a 5 MB single line then a marker:

```
total command output: 8498 bytes
contains TAIL-MARKER: 1
```

so it stays bounded and the tail that explains the failure is what survives.

crates/icp/src/network/managed/docker.rs
```diff
@@ -512,6 +543,162 @@ pub async fn spawn_docker_launcher(
...
+        if output.len() > MAX_LOG_BUFFER_BYTES {
+            output = tail_bytes(&output, MAX_LOG_BUFFER_BYTES).to_string();
+        }
+    }
+    // The full id is already on the error's first line, so the hint uses the short form.
+    let short_id = short_id(container_id);
+    let tail = output_tail(&output);
+    if tail.is_empty() {
+        detail.push_str(&format!(
+            "\nNo output was captured; see `docker logs {short_id}` for details."
```
adamspofford-dfinity: The container's about to be deleted on exit, unless the user has configured otherwise. Not a hugely useful command.

crates/icp/src/network/managed/docker.rs
```diff
@@ -512,6 +543,162 @@ pub async fn spawn_docker_launcher(
...
+            stdout: true,
+            stderr: true,
+            tail: MAX_OUTPUT_TAIL_LINES.to_string(),
+            ..<_>::default()
+        }),
+    ));
+    let mut output = String::new();
+    while let Some(chunk) = logs.next().await {
+        match chunk {
+            Ok(chunk) => output.push_str(&chunk.to_string()),
```
adamspofford-dfinity: Ignorable if it's too much effort: Multi-byte characters can be split across chunk boundaries, so a chunk can contain invalid UTF-8, resulting in that character being replaced with a couple U+FFFDs.

crates/icp/src/network/managed/docker.rs
lwshang: Good catch, and it's worse than "unless the user has configured otherwise" — autocontainerizing hardcodes `rm_on_exit: true` (`run.rs:371`), so on Windows the container is *always* removed, and the drop guard does it as this very error propagates. Fixed in 1670fde7:

- Inline output drops the attribution: `Container output:` rather than `Container output (from \`docker logs X\`)`. It's right there, so naming a command that may not work only misleads.
- No-output case splits on `rm_on_exit`: `The container produced no output, and was removed on exit.` versus `...; see \`docker logs <id>\` for details.` only when the container will actually survive.
- The background hint is now deferred until the network is up, for the same reason — advertised before the wait, it also printed on the failure path.

```
Error: docker container ca12eef0cda5... exited prematurely with status 7
The container produced no output, and was removed on exit.
```

crates/icp/src/network/managed/docker.rs
lwshang: Chased this one down, and the symptom is real but the cause is upstream of us — so it isn't fixable here.

The json-file driver stores each log entry as a JSON string, which must be valid UTF-8, so the daemon substitutes U+FFFD when its own 16 KB chunking bisects a character. That happens at write time, before bollard or icp-cli see anything. Plain `docker logs` on a container emitting one long line of U+20AC, no icp-cli involved:

```
bytes             : 32778
U+FFFD occurrences: 5
U+20AC occurrences: 10921
```

So per-chunk decoding wasn't adding the damage — it was already baked in. (bollard isn't the culprit either: despite the internal name `process_into_stream_string`, `NewlineLogOutputDecoder` hands over raw `Bytes` per frame.)

What I did change in 1670fde7 is to accumulate bytes and decode once, so framing can't influence the result at all, and to make the incremental trim cut on a character boundary — that one *could* have bisected a character and added a replacement char of our own. The daemon's replacements remain.

adamspofford-dfinity (approved):

PR #683, opened 2026-07-29
feat: deploy directly to cloud engines via engine operators

NikolaMilosa:
# Motivation

Before this PR, to deploy to a cloud engine we had to directly call the mgmt canister. It would accept the call only if the caller is a subnet admin. Subnet admins are scars resources (we allow 10 per cloud engine) and they give way more power than just creating canisters. For example a subnet admin can go and delete any canister in the cloud engine. To tackle that problem the team added the engine operator canister. Each engine has a dedicated _system_ engine operator canister which is meant to have its own dedicated ACL and do permission checking. 

The [engine canister](https://dashboard.internetcomputer.org/canister/q6cfj-fyaaa-aaaar-qb77q-cai), which is the central registry for engines, will provide the mapping from `subnet-id` to a dedicated `engine-operator-id` which the icp cli should use to reach the the correct engine operator for a subnet the user wants to deploy to.  

Since different environments may deploy the central canister at different locations it is possible to override the engine canister id (the registry canister for engines) by specifying the `ENGINE_CANISTER_ID=...` variable. 

Example call:
```bash
ENGINE_CANISTER_ID=5s2ji-faaaa-aaaaa-qaaaq-cai ~/icp-cli-real/target/debug/icp canister create -e dev --detached --subnet 7pnox-zummy-6hxlq-u6qwm-nloqq-wz3bx-slcow-l3vri-2g6iz-ln6wh-zae
WARN fetching the root key from https://opencloud-internal.demo.farm.dfinity.systems/; its provenance is not verified (trust-on-first-use)
Created canister with ID jwksz-eqaaa-aaaab-aaaaq-cai
```

With this, we can get rid of subnet admins.


crates/icp-canister-interfaces/src/engine_canister.rs
```diff
@@ -0,0 +1,79 @@
...
+/// non-empty, otherwise falls back to [`ENGINE_CANISTER_CID`]. Returns an error
+/// only when the override is set but not a valid principal.
+pub fn engine_canister_id() -> Result<Principal, String> {
+    match env::var(ENGINE_CANISTER_ID_ENV) {
+        Ok(value) if !value.trim().is_empty() => Principal::from_text(value.trim())
+            .map_err(|e| format!("invalid {ENGINE_CANISTER_ID_ENV}: {e}")),
+        // Unset, empty, or non-UTF-8: use the built-in default.
+        _ => Ok(Principal::from_text(ENGINE_CANISTER_CID)
+            .expect("ENGINE_CANISTER_CID is a valid principal")),
+    }
```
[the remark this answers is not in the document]
NikolaMilosa: Addressed

crates/icp-cli/src/operations/create.rs
```diff
@@ -214,7 +232,19 @@ impl CreateOperation {
...
+            // legacy management-canister path until we're sure the latter is no
+            // longer needed.
+            match self.create_on_cloud_engine(settings, selected_subnet).await {
+                Ok(cid) => cid,
+                Err(e) => {
+                    warn!(
+                        "engine-operator creation failed ({e}); \
+                         falling back to management-canister creation"
+                    );
+                    self.create_mgmt(settings, selected_subnet).await?
```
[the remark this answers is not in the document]
NikolaMilosa: I addressed this but I don't really know if it is valid. Would be nice to have a second opinion from the reviewer. Also to see if my addressing was correct.

NikolaMilosa:
Opencloud stack has had this deployed to production so now a cloud engine owner can link their identity and do something like this:
```bash
󰣇 (on omarchy) ~/Downloads ❯ ./icp-test identity list
* opencloud-prod-nikola 5rbsx-2blhm-7gpfb-ekmug-3jb64-4bzdx-kzpqo-yisej-3zgdu-ce2ce-pae
  engine-admin-2        5rbsx-2blhm-7gpfb-ekmug-3jb64-4bzdx-kzpqo-yisej-3zgdu-ce2ce-pae
  opencloud-dev         begoe-c6xh2-q3pwn-dd6hb-sitwl-zw2o7-imsaq-q2npf-253kp-bwlar-aqe
  anonymous             2vxsx-fae
# They would have to find their subnet id from settings
󰣇 (on omarchy) ~/Downloads ❯ ./icp-test canister create --detached -e ic --subnet fj2ae-lldsj-bshv6-najg4-b7d4a-dba6x-xrf5n-qav4g-2en7f-w6in7-yqe
Created canister with ID vj246-waaaa-aaabn-aaaca-cai
```

After that the canister is visible in the UI and is present on the engine
image

crates/icp-cli/src/operations/create.rs
```diff
@@ -214,7 +232,28 @@ impl CreateOperation {
...
+            // mapped to `NoEngineOperator` in `resolve_engine_operator`). Every
+            // other resolution error (invalid registry id, other query/decode
+            // failures) is propagated. Crucially, the fallback is decided
+            // *before* any operator `create_canister` update is submitted: once
+            // we hand off to the operator, a failure may mean the canister was
+            // already created, so retrying via the management canister could
+            // double-create and double-charge. Those errors propagate too.
+            match self.resolve_engine_operator(selected_subnet).await {
+                Ok(operator) => self.create_via_operator(settings, operator).await?,
+                Err(e @ CreateOperationError::NoEngineOperator { .. }) => {
```
raymondk, line 249: Why isn't this a hard error if we're counting on the operator to be installed on all the cloud engines?

crates/icp-canister-interfaces/src/engine_canister.rs
```diff
@@ -0,0 +1,87 @@
...
+/// Resolve the engine-canister principal to talk to.
+///
+/// Uses the value of the `ENGINE_CANISTER_ID` environment variable when set and
+/// non-empty, otherwise falls back to [`ENGINE_CANISTER_CID`]. Returns an error
+/// when the override is set but invalid — either not a valid principal, or not
+/// valid Unicode. Only an absent or empty (whitespace-only) variable uses the
+/// default: a configured-but-invalid value must never silently route to the
+/// built-in registry (and thus a different environment).
+pub fn engine_canister_id() -> Result<Principal, String> {
+    match env::var(ENGINE_CANISTER_ID_ENV) {
```
raymondk, line 40: We usually collect all the known environment variables very early and avoid having deeply nested modules changes their behavior based on an environment variable.

Also, I assume you want an environment variable because for testnets you might have a different canister id - isn't it possible to force the canister id to be the same on testnets as it is on mainnet like we do for other well known canisters?

crates/icp-canister-interfaces/src/engine_canister.rs
```diff
@@ -0,0 +1,87 @@
...
+use std::env::{self, VarError};
+
+use candid::{CandidType, Principal};
+use serde::Deserialize;
+
+/// Default engine-canister principal (staging/prod registry).
+///
+/// Overridable at runtime via the [`ENGINE_CANISTER_ID_ENV`] environment
+/// variable — see [`engine_canister_id`].
+pub const ENGINE_CANISTER_CID: &str = "q6cfj-fyaaa-aaaar-qb77q-cai";
```
raymondk, line 23: is there a chance this canister will move to the system subnet at some point and that we will be forced to change its id?

crates/icp-cli/src/operations/create.rs
raymondk: On second thought, maybe it makes sense to keep the fallback while things are still in flux.

NikolaMilosa: Since our canisters are very new we don't have them deep integrated in the IC ecosystem. For example you can have a perfectly fine running engine (from the protocol perspective) that doesn't have any of these canisters. This is an _optimistic_ handling where we still try to fulfill the users request. In reality, we will change that no-one on mainnet (who is managed by our canisters) will have subnet admin powers. 

If we make these hard errors we assume that cloud engines are coupled to these canisters. In reality they are, but if you used proposals today you could do the same without our canisters.

Icp tool should ideally not fail if someone else creates an engine that is completely valid. 

Would you prefer it to be different?

crates/icp-canister-interfaces/src/engine_canister.rs
NikolaMilosa: The chance is at 100% but we don't know when and we don't know what canister id will it have.

crates/icp-canister-interfaces/src/engine_canister.rs
NikolaMilosa: There is a way, I can look into it and maybe we can have it, but still overriding would be useful. For example we may deploy a staging canister to a different id to do some testing ahead of time. 

Don't you think?

crates/icp-canister-interfaces/src/engine_canister.rs
raymondk: The concern is that when it does move, the error message won't be very informative and it will silently fallback on the fallback. After chatting with @lwshang our suggestion is:
1. If there is a place where we can get the canister id (and perhaps cache it for some time) then we should do that, ortherwise
2. We should have an error message if that canister id is not found anymore saying something along the lines of:

> By default the icp-cli will lookup the cloud engine control canister in canister `qa6c...` but it seems to be unavailable. You can set the env var to the correct canister or try to upgrade icp-cli

NikolaMilosa: What we won't do:
We will never migrate from the canister and delete it. The only thing that could happen is that we move it to the NNS (it will for sure happen but we don't know when). In the NNS we may reuse the id that we have already or we create a completely new one, this is what we need to decide.

What we will do in case we migrate:
Deploy a shim canister that will panic loudly saying what to do.

I am not sure if there is a place to fetch this canister from but we can build something. Maybe we can put a file that is queried once a day from git and cached locally? That way if we merge a pr with this change the next day it will pick it up.

adamspofford-dfinity (commented):
LGTM besides Raymond's comments

adamspofford-dfinity (approved):

PR #693, opened 2026-08-04
feat: Allow environment variables to be specified in a file

adamspofford-dfinity:


crates/icp-cli/src/operations/bundle.rs
```diff
@@ -873,6 +887,32 @@ async fn inline_environments(
...
+                };
+                for (variable, var) in vars.iter_mut() {
+                    if let ManifestEnvVar::Path { path } = &*var {
+                        let value = fs::read_to_string(&override_dir.join(path)).context(
+                            ReadEnvVarSnafu {
+                                canister: canister_name,
+                                variable,
+                            },
+                        )?;
+                        *var = ManifestEnvVar::Value(value.trim().to_owned());
```
lwshang, line 946: Unlike the `init_args` override just above, this has no `canonicalize_within_project` check. I reproduced `path: ../outside-secret.txt` bundling successfully and inlining the host file's contents into the distributed `icp.yaml` — for both a canister's own settings and an environment override. The same path under `init_args` is rejected with "bundles cannot reference files outside the project". Worth the same check here, plus one in `resolve_manifest_settings`, since a canister's own vars are resolved during consolidation before bundle sees them.

crates/icp/src/project.rs
```diff
@@ -1002,7 +1093,7 @@ fn build_environment_canisters(
     if let Some(settings) = root_settings {
         for (name, s) in settings {
             if let Some((_p, canister)) = cs.get_mut(name) {
-                canister.settings = s.clone();
+                canister.settings = resolve_manifest_settings(s, root_settings_dir, name)?;
```
lwshang: `settings` paths resolve against the declaring manifest's directory; `init_args` paths on L1103 resolve against the referenced canister's directory. Two rules inside one override block will surprise people. Align them, or contrast them explicitly in the Environment Overrides docs.

crates/icp/src/canister/mod.rs
```diff
@@ -207,9 +210,48 @@ impl schemars::JsonSchema for ControllerRef {
...
+/// The object form reads the value from a file, relative to the directory of the
+/// manifest declaring it — a canister's own directory for `canister.yaml`, and
+/// the project or environment manifest's directory for an environment override:
+/// ```yaml
+/// environment_variables:
+///   API_KEY:
+///     path: ./secrets/api-key
+/// ```
+#[derive(Clone, Debug, PartialEq, Deserialize, JsonSchema, Serialize)]
+#[serde(untagged)]
```
lwshang: `PORT: 8080` now fails with `data did not match any variant of untagged enum ManifestEnvVar` instead of the old `invalid type: integer 8080, expected a string`. `#[serde(expecting = "a string, or `{ path: <file> }`")]` on the enum replaces that message.

lwshang (commented):
Nice feature, and the shared `environment_override_dir` rule is the right shape. Checked out locally: builds, fmt/clippy clean, schemas regenerate with no drift, `cargo test -p icp --lib` and the new bundle test pass. Three notes inline — the first one I'd want resolved before merge.

crates/icp/src/project.rs
adamspofford-dfinity: Chose to align them.

lwshang (approved):

PR #696, opened 2026-08-06
feat: env-specific builds

adamspofford-dfinity:


crates/icp-cli/src/operations/build.rs
```diff
@@ -69,6 +70,7 @@ pub(crate) async fn build(
                 &Params {
                     path: canister_path.to_owned(),
                     output: wasm_output_path.to_owned(),
+                    environment: environment.to_owned(),
                 },
```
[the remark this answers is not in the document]
adamspofford-dfinity: Not worth thinking about

crates/icp-cli/tests/build_tests.rs
```diff
@@ -83,6 +83,62 @@ fn build_adapter_script_multiple() {
...
+    // Project manifest: the build step records the environment it was built for
+    let pm = formatdoc! {r#"
+        canisters:
+          - name: my-canister
+            build:
+              steps:
+                - type: script
+                  commands:
+                    - echo "$ICP_CLI_ENVIRONMENT" > '{recorded}'
+                    - touch "$ICP_WASM_OUTPUT_PATH"
```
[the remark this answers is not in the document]
adamspofford-dfinity: This is false.

crates/icp-cli/src/commands/build.rs
raymondk: Should probably add docs and the option to receive the environment from an environment variable no?

crates/icp-cli/src/commands/deploy.rs
raymondk: Update the docs?

docs/reference/cli.md
```diff
@@ -921,7 +921,7 @@ using --args or --args-file:
 * `--no-create` — If any canisters do not exist, error instead of creating them
 * `-y`, `--yes` — Skip confirmation prompts, including the Candid interface compatibility check
 * `--identity <IDENTITY>` — The user identity to run this command as
-* `-e`, `--environment <ENVIRONMENT>` — Override the environment to connect to. By default, the local environment is used
+* `-e`, `--environment <ENVIRONMENT>` — Override the environment to build for. By default, the local environment is used.
```
raymondk: to build for & deploy to

crates/icp-cli/src/operations/build.rs
```diff
@@ -69,6 +70,7 @@ pub(crate) async fn build(
                 &Params {
                     path: canister_path.to_owned(),
                     output: wasm_output_path.to_owned(),
+                    environment: environment.to_owned(),
```
lwshang, line 73: Build output can now vary by environment, but the artifact store still has one slot per canister (`artifacts.save(&canister.name, …)` 18 lines down, `.icp/cache/artifacts/<name>`). That was safe under the reproducibility rule this PR removes from `build-deploy-sync.md`. Now `icp canister install <c>` without `--wasm` serves whatever was built last — so `deploy -e staging` then `install -e local` installs the staging wasm, and `project bundle` (defaults to `ic`) silently clobbers the local artifacts for every canister.

Key the artifact dir by environment, or keep a caveat in the docs in place of the deleted bullet.

crates/icp-cli/tests/bundle_tests.rs
```diff
@@ -817,6 +817,60 @@ fn bundle_rejects_source_outside_project() {
 /// exist when bundling validates the sync sources, before the build. Validation
 /// must resolve sync paths lexically (no canonicalization) so a not-yet-built
 /// directory is accepted; the build then creates it before it is archived.
+/// The environment reaching build steps as `ICP_CLI_ENVIRONMENT` defaults to `ic`
```
lwshang, line 815: This test landed between the doc comment and the fn it documents — the "resolve sync paths lexically" comment now describes `bundle_builds_for_ic_by_default`, and `bundle_accepts_synced_dir_created_by_build_step` has none. Move the new test above the comment block.

crates/icp-cli/src/commands/build.rs
```diff
@@ -13,12 +16,18 @@ pub(crate) struct BuildArgs {
...
 
     #[command(flatten)]
-    pub(crate) environment: EnvironmentOpt,
+    pub(crate) environment: BuildEnvironmentOpt,
 }
 
+arg_struct_change_help!(
+    EnvironmentOpt => BuildEnvironmentOpt,
+    arg = "environment",
+    help = "Override the environment to build for. By default, the local environment is used."
```
lwshang: Drop the trailing period — clap strips it from doc comments, so every other `--environment` renders without one, and it shows up in the generated `cli.md`. Same on `deploy.rs:104`. Regenerate `cli.md` after.

crates/icp-cli/src/operations/build.rs
adamspofford-dfinity: Resolution: we do not want environment-specific build caches, because that will just lead to installing stale artifacts given the same command pattern, plus change slightly how `icp canister install` works.

lwshang (approved):
Docs-only follow-up pushed in f95d5163 covering the artifact-store caveat. Verified on a91b7adb: clippy clean, bundle_tests 23/23, build_tests 11/11, and generate-cli-docs.sh produces no diff.

Issue #493, opened 2026-04-09
RFC: document --json output schemas in --help and/or generated docs


marc0olo:
## Background

Many `icp` commands support a `--json` flag for machine-readable output, but `--help` currently says nothing about what the output looks like beyond "Output command results as JSON". For an agent (or a developer writing scripts) this means the JSON schema has to be discovered by running the command live or reading the source code.

This RFC asks: **where and how should JSON output schemas be documented?**

---

## Current state

The following commands produce structured JSON output with no schema documentation in `--help`:

| Command | Schema (verified from source) |
|---|---|
| `icp deploy --json` | `{"canisters": [{"name": string, "canister_id": string, "url": string\|null}]}` |
| `icp network status --json` | `{"managed": bool, "api_url": string, "gateway_url": string\|null, "root_key": string, "candid_ui_principal": string\|null, "proxy_canister_principal": string\|null}` |
| `icp canister call --json` | `{"response_bytes": string, "response_text": string\|null, "response_candid": string\|null}` |
| `icp canister create --json` | `{"canister_id": string, "canister_name": string\|null}` |
| `icp canister list --json` | `{"canisters": [string]}` |
| `icp canister logs --json` | `{"log_records": [{"timestamp": u64, "index": u64, "content": string}]}` |
| `icp canister snapshot create --json` | `{"snapshot_id": string, "taken_at_timestamp": u64, "total_size_bytes": u64}` |
| `icp canister snapshot list --json` | `{"snapshots": [{"snapshot_id": string, "taken_at_timestamp": u64, "total_size_bytes": u64}]}` |
| `icp cycles balance --json` | `{"balance": string}` |
| `icp cycles mint --json` | `{"deposited": string, "new_balance": string}` |
| `icp cycles transfer --json` | `{"block_index": string}` |
| `icp token transfer --json` | `{"block_index": string}` |
| `icp identity list --json` | `{"default_identity": string, "identities": [{"name": string, "principal": string}]}` |
| `icp identity new --json` | `{"seed_phrase": string}` |

Note: `icp canister status --json` produces a more complex structure (full status object) that is not listed here as it is not a simple flat schema.

---

## Why it matters

Two specific workflows already referenced in the `icp-cli` skill documentation rely on parsing JSON output without having the schema documented:

- `icp network status --json` — used in dev server configuration to extract `api_url` and `root_key`
- `icp deploy --json` — used to extract `canister_id` and `url` per deployed canister

[...]

PR #705, opened 2026-08-12
fix: emit newline-delimited JSON from `icp canister logs --follow --json`

raymondk:
## What this fixes

Two bugs in `icp canister logs --follow`, both in `crates/icp-cli/src/commands/canister/logs.rs`.

**1. `--follow --json` was neither delimited nor streaming.** Each record was written with a bare `serde_json::to_writer(stdout(), ...)` and no newline, so:

- records ran together as `{...}{...}{...}` on one line, which line-oriented consumers (`head`, `while read`, per-line `json.loads`) cannot split into records — `jq` itself tolerates concatenated values, so it is the rest of the toolchain that breaks — and
- with no newline ever written, nothing left the line-buffered stdout buffer until it filled (~1KB) or the process exited — so a live tail showed *nothing*.

This was pre-existing and deliberately out of scope for the `--json` orientation fix in #704, which left follow mode untouched.

**2. A closed output pipe was reported as an error.** `... --follow | head -1` makes the next write fail with `EPIPE`. On the `--json` path that propagated out as `Error: Broken pipe (os error 32)` and a non-zero exit; on the human-readable path `println!` *panicked*. A consumer that stops reading is an ordinary way to stop a tail, not a failure. This one predates #704 as well, but piping NDJSON into `head`/`jq` is exactly the pattern fix 1 invites, so it became far more reachable.

## Approach

Each record is now serialized to a `String` and written with a single `writeln!`, and the follow loop is labelled so a broken pipe can break out of it:

```rust
match writeln!(stdout().lock(), "{line}") {
    Ok(()) => {}
    Err(e) if e.kind() == ErrorKind::BrokenPipe => break 'follow,
    Err(e) => return Err(e.into()),
}
```

Notes on the two judgement calls in there:

- **Newline, not an explicit `flush()`.** Rust's `std::io::stdout()` is a `LineWriter` unconditionally — unlike C stdio, it does not switch to block buffering when stdout is not a terminal — so writing the newline flushes the record. That is verified by the tests below rather than taken on trust, so no `flush()` call was added.
- **One write site for both output modes.** Collapsing the `--json` and human-readable branches means the broken-pipe handling covers both, which is what removes the pre-existing `println!` panic. The emitted text is byte-identical in both modes.

Also drops a comment in `fetch_and_display_logs` that claimed follow mode does *not* terminate its records — true when written, false as of this PR — while keeping the explanation of why the non-follow document needs its own trailing newline.

## Compatibility

`--follow --json` changes from concatenated JSON objects to newline-delimited JSON. Anything parsing the old output must now read one object per line. Called out in `CHANGELOG.md` under `# Unreleased`.

The non-follow paths are untouched: `--json` is still a single `JsonListRecord` document terminated by one newline, the default is still the human-readable `[idx. timestamp]: content` lines, and neither `JsonFollowRecord` nor `JsonListRecord` changed shape.

## Test coverage

All of it lives in `canister_logs_follow_mode` in `crates/icp-cli/tests/canister_logs_tests.rs`, as three phases sharing one network start and one `moc` deploy — the pattern `canister_logs_single_fetch` already uses. Each phase triggers its own fresh `log_repeated`, so nothing is replayed from the 1-hour lookback.

1. **Polling** (pre-existing) — the human-readable output picks up logs that did not exist when the follow started.
2. **NDJSON streaming** — covers delimiting *and* streaming, not just delimiting. `assert_cmd`'s `.timeout()` SIGKILLs the process, so anything still in the stdout buffer is lost; observing output at all therefore proves it was flushed as records arrived. Each line is parsed on its own, which also pins the delimiting, since `serde_json::from_str` rejects the trailing content of a concatenated `{...}{...}`.
3. **Closed pipe**, both output modes — spawn `--follow` with a piped stdout, read one record, drop the reader (what `head -1` does to the pipe), then emit one more log so the next poll *must* write to the now-closed pipe, and assert a clean exit. That last step is what makes it deterministic rather than racing the canister.

Both fixes were verified to **fail without them**, not just pass with them: `write!` in place of `writeln!` fails the run, as does dropping the `ErrorKind::BrokenPipe` arm.

Phase 3 needs a raw `std::process::Command` to control the child's stdio, so `TestContext::icp_std()` was added and `icp()` is now a thin `Command::from_std` wrapper over it — no duplicated env-isolation block.

`cargo test`, `cargo fmt`, and `cargo clippy --all-targets` are clean. (`canister_install_large_wasm_chunked` fails in my local environment for want of `wasm-tools`; unrelated, and it passes in CI.)

## Not addressed here

**Trailing newlines CLI-wide.** `canister logs` is now the only command whose `--json` output ends in a newline; `canister list`/`metadata`/`create`/`call`, `deploy`, all of `token`/`cycles`/`identity` and the `snapshot` commands write a bare document with no terminator. The justification in the comment this PR rewrites — it is the entirety of stdout, so the prompt lands mid-line — applies verbatim to all of them. Cosmetic rather than a parsing bug, so it belongs in a separate consistency sweep, pairing naturally with #493. Worth noting the sentence this PR deletes was that divergence's only in-code marker; deleting it is still correct, since within this module the contrast is now false.

**Schema documentation.** The `--json` output schema still has no owning documentation surface — `docs/reference/cli.md` is generated from clap help and describes flags only. Creating one is out of scope for a bug fix; #493 already tracks it, and the NDJSON contract for `--follow --json` should be written down when that happens.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,7 @@ bump. Currently experimental: project bundling, project dependencies
 
 # Unreleased
 
-* fix: `icp canister logs` no longer has its output formats swapped. `--json` emitted the human-readable lines and the default emitted JSON; now `--json` emits machine-readable JSON (terminated by a newline) and the default emits the human-readable lines. `--follow` already had the right orientation (see the next entry for a separate `--follow --json` fix). Scripts that relied on the inverted behavior — parsing the default output as JSON, or omitting `--json` to get JSON — must now pass `--json`.
-* fix: `icp canister logs --follow --json` now emits newline-delimited JSON (one record per line) and each record reaches stdout as it arrives. Previously the records were concatenated with no separator (`{...}{...}`) and, with no newline ever written, stayed in the line-buffered stdout buffer — so a live tail produced no output until the buffer filled or the process ended. Anything parsing the old concatenated output must now read one JSON object per line.
+* fix: `icp canister logs` output formats are corrected. `--json` now emits machine-readable JSON and the default emits the human-readable lines (the two were swapped), and `--follow --json` emits newline-delimited JSON, one record per line, streamed as each record arrives. This is breaking for scripts: parsing the default output as JSON now requires `--json`, and consumers of `--follow --json` must read one JSON object per line.
 
 # v1.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ bump. Currently experimental: project bundling, project dependencies
 # Unreleased
 
 * fix: `icp canister logs` no longer has its output formats swapped. `--json` emitted the human-readable lines and the default emitted JSON; now `--json` emits machine-readable JSON (terminated by a newline) and the default emits the human-readable lines. `--follow` was already correct and is unchanged. Scripts that relied on the inverted behavior — parsing the default output as JSON, or omitting `--json` to get JSON — must now pass `--json`.
+* fix: `icp canister logs --follow --json` now emits newline-delimited JSON (one record per line) and each record reaches stdout as it arrives. Previously the records were concatenated with no separator (`{...}{...}`) and, with no newline ever written, stayed in the line-buffered stdout buffer — so a live tail produced no output until the buffer filled or the process ended. Anything parsing the old concatenated output must now read one JSON object per line.
 
 # v1.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ bump. Currently experimental: project bundling, project dependencies
 
 # Unreleased
 
-* fix: `icp canister logs` no longer has its output formats swapped. `--json` emitted the human-readable lines and the default emitted JSON; now `--json` emits machine-readable JSON (terminated by a newline) and the default emits the human-readable lines. `--follow` was already correct and is unchanged. Scripts that relied on the inverted behavior — parsing the default output as JSON, or omitting `--json` to get JSON — must now pass `--json`.
+* fix: `icp canister logs` no longer has its output formats swapped. `--json` emitted the human-readable lines and the default emitted JSON; now `--json` emits machine-readable JSON (terminated by a newline) and the default emits the human-readable lines. `--follow` already had the right orientation (see the next entry for a separate `--follow --json` fix). Scripts that relied on the inverted behavior — parsing the default output as JSON, or omitting `--json` to get JSON — must now pass `--json`.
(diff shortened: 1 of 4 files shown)


marc0olo (approved):
Approving. I checked the load-bearing claim rather than taking it on faith: Rust's `Stdout` wraps a `LineWriter` unconditionally (not TTY-dependent, unlike C stdio), so the newline really does flush and an explicit `flush()` would be noise. Two supporting details that aren't in the description but hold up — `serde_json` escapes newlines inside strings as `\n`, so a log message containing a newline can't trigger a premature partial flush mid-record; and scoping `stdout().lock()` to a single record rather than hoisting it above the loop is the right call, since hoisting would hold the lock across the `.await` on the next fetch.

The test is honest about what it covers. `.timeout()` SIGKILLs, so observed output must have been flushed on arrival — and the Motoko `log_repeated` uses a `recurringTimer` and returns immediately, so the five records really are produced after follow starts rather than being replayed from the 1-hour lookback. Delimiting is covered too, even though the description doesn't claim it: `serde_json::from_str` rejects trailing content, so the old `{...}{...}` on one line would fail to parse.

Three non-blocking notes:

1. **Broken pipe.** `--follow --json | head -1` now hits EPIPE on the next `writeln!` and exits non-zero — previously masked by the buffering. Not a regression (the human-readable path already *panics* via `println!` on the same condition), but piping NDJSON into `head`/`jq` is exactly the pattern this change invites, so it's far more reachable now. Treating `ErrorKind::BrokenPipe` as a normal loop break at `logs.rs:271` would be a small follow-up.

2. **Test cost.** `canister_logs_follow_mode_json` duplicates the full network-start + `moc` deploy + 7s follow window of `canister_logs_follow_mode`. A second invocation inside the existing test — the pattern `canister_logs_single_fetch` already uses — would be equally valid evidence (the SIGKILL argument holds either way) at roughly half the wall clock.

3. **Trailing newlines elsewhere.** Worth flagging for a follow-up, not this PR: `canister logs` is now the *only* command whose `--json` output ends in a newline. `canister list`, `metadata`, `create`, `call`, `deploy`, all of `token`/`cycles`/`identity`, and the `snapshot` commands all write a bare document with no terminator. The justification in the rewritten comment — "it is the entirety of stdout, so without it the shell prompt lands mid-line" — applies verbatim to all ~18 of them. Cosmetic rather than a parsing bug, so a separate consistency sweep; it pairs naturally with #493.

Small irony on that last point: the sentence this PR deletes ("Unlike the other `--json` sites in this module…") was pointing at exactly that divergence. Deleting it is correct as written, since within this module the contrast is now false — but the CLI-wide inconsistency loses its only in-code marker.

The streaming fix itself is correctly scoped to this command: `--follow` is the only streaming flag in the CLI, and every other `--json` site writes one document immediately before exit.

PR #709, opened 2026-08-13
refactor: report progress through a new icp-events crate instead of the terminal

raymondk:
## Why

Progress reporting is what welded `crates/icp-cli/src/operations/` to the binary: six of those files imported `crate::progress`, a seventh drove `indicatif` directly, and two `crates/icp` ports took a `tokio::sync::mpsc::Sender<String>` that only existed because the binary had a progress bar to feed. Nothing else in the crate split can start until that is inverted, so it goes first — and it lands as one change rather than a partial one, so the tree is never half-converted.

## What

**New crate `crates/icp-events`** — progress and user-facing notices as data. Depends on **serde and futures only**; not on `icp`, not on an async runtime, not on anything terminal-shaped (`indicatif`, `dialoguer`, `clap`, `console`).

- `Event` — `TaskStarted`/`TaskMessage`/`TaskPosition`/`TaskFinished`, `StepStarted`/`StepOutput`/`StepFinished`, and `Notice { level, message }`
- `TaskKind` — closed enum carrying the three shapes that exist: `Spinner`, `Steps { output_label }`, `Bytes { total }`
- `Reporter` / `Task` — `Task::run(fut, success, error)` replaces `ProgressManager::execute_with_progress`. A dropped `Task` still emits `TaskFinished { Neutral }`, so an early return cannot leave a sink holding a bar that spins forever.
- `OutputWriter` — a cheap, cloneable handle for the lines a running step produces. Each line becomes an `Event::StepOutput` *and* is kept in the task's step log (capped at `MAX_RECORDED_LINES_PER_STEP`), so an operation that fails can replay the failing step once the bars are down.
- `EventSink`, `DiscardSink`, `RecordingSink`, `CancelToken`

**`crates/icp-cli/src/events.rs`** — `IndicatifSink`, now the only place in the workspace that knows about `indicatif`, and where the styles live.

**Every operation converted.** `install.rs`, `settings.rs`, `binding_env_vars.rs`, `candid_compat.rs`, `build.rs`, `sync.rs`, `snapshot_transfer.rs` and `bundle.rs` take a `&Reporter`; none takes `debug: bool`. The commands that drew their own bars — `deploy.rs`, `network/start.rs`, `network/update.rs`, `canister snapshot upload`/`download` — go through the reporter too.

`build_many`/`sync_many`/`create_bundle` keep one bool, renamed `all_step_output`: replaying every step of a failure rather than only the failing one is the only thing `debug` still decided once bar-hiding moved into the reporter.

**Both port signatures changed.** `Build::build` and `Synchronize::sync` — and the script, prebuilt, wasm-resolve, plugin and script-runner paths beneath them — take `Option<OutputWriter>` instead of `Option<Sender<String>>`. `crates/icp` gains no terminal-shaped dependency; it depends on `icp-events`, which is serde + futures.

Two things fall out of that:

- `MultiStepProgressBar`'s channel plus the background task that drained it are gone, and with them `end_step().await`, which existed only to wait for that drain. Recording is now synchronous.
- the sync plugin runtime reported lines with `try_send` on a bounded channel, so a noisy plugin silently dropped output when the renderer fell behind. It no longer can.

**`progress.rs` is deleted.** The compiler was the gate, as the architecture note asked: it was removed and `cargo check -p icp-cli --all-targets` had to come back clean.

## The two macOS failures

`events::rendering_equivalence` failed on `macos-15` only, with an extra bare `" ~ "` frame captured before the `[backend]` prefix appeared.

Not the rebase, and not a new defect: `ProgressBar::enable_steady_tick` spawns a thread whose first act is to draw a frame (`TickerControl::run` ticks *before* its first wait), so a prefix set after that call races that tick. Both renderers had that ordering, so both could produce the frame; Linux and Windows happened to win the race and macOS did not. Reproduced deterministically by inserting a sleep into the pre-inversion path, which produced the exact failing value:

```
[" ~ ", "[backend] ~ ", "[backend]   Skipped (not an upgrade)"]
```

Fixed in the sink rather than in the assertion: every bar is now fully styled and labelled before it is added to the `MultiProgress` and, for spinners, before the ticker starts, so no frame can be drawn unlabelled. `tests::a_spinner_is_labelled_before_its_first_tick` waits out several tick intervals and asserts every raw frame carries the prefix.

## Terminal output is unchanged — how that was verified

The previous revision compared old and new renderers frame-by-frame against a recording `indicatif::TermLike`. Those renderers no longer exist, so before deleting them their output was captured and pinned as literals in `events.rs::rendering`:

- spinner success and failure — full three-frame sequences, including the final `✔`/`✘`
- the `candid_compat` skip, which used `finish_with_message` and drew exactly one frame — so `Outcome::Neutral` must call `finish_with_message` too, not `set_message` + `finish`, or it inserts an extra redraw
- multi-canister bar ordering
- the byte bar's at-rest line, character for character

Four things are normalized away, each because it is a property of the machine rather than of what was drawn: the blank padding row indicatif writes to fill the terminal line (a function of the frame it follows); the timer-driven animation glyph; SGR colour codes (a `TermLike` draw target is never stderr, so indicatif leaves the styled fields following `console`'s stdout-based detection — the frames carry colour when `cargo test` runs from a terminal and not when it is piped); and a spinner frame with no message yet, which is the animation thread getting a frame in before the operation has said anything. The last is what the macOS failure was made of; the underlying missing-prefix defect is fixed, not hidden. `normalization_strips_the_colour_codes_a_terminal_would_add` covers the colour rule directly rather than leaving it assumed.

Bar *widths* are deliberately not pinned once bytes have moved: `{wide_bar}` takes whatever width the rest of the line leaves over, which includes a transfer rate whose text is as long as the machine happened to be fast. Those tests pin the glyph set and the `done/total` counter instead.

One intentional difference, called out rather than smoothed over: the byte-transfer bars in `canister snapshot upload`/`download` are still drawn under `--debug`, matching today's behaviour, where the spinners are hidden. They are the only indication a long transfer is moving.

## Testability

Every converted operation is unit-tested by running it for real against a `RecordingSink` and asserting on the resulting `Vec<Event>` — task order, exact messages, outcomes, step titles, step output. The three added in this pass:

- `build.rs` — a build's whole reported shape, a failing step's output and failure outcome, canister ordering
- `sync.rs` — the same three, driven through an injected `ScriptRunner` that reports lines instead of spawning a shell
- `snapshot_transfer.rs` — a resumed upload reports the offset it recovered before anything else, so the bar opens where the last attempt stopped rather than at zero

`icp`'s own suite gains one: a build step's output arrives through the writer and is retained for replay, with no terminal involved.

## Deliberately not in this PR

- The 29 user-facing `tracing` calls inside `operations/` are still `tracing` calls. `Event::Notice` exists and the sink routes it, so converting them needs no redesign. (INFO-level `tracing` is *product output* in this CLI — `logging.rs` installs a `UserLayer` that prints `Level::INFO` to stderr unprefixed.)
- Two bespoke spinners still build their own `indicatif` bars: `commands/canister/migrate_id.rs` and `commands/identity/link/web.rs`. Neither ever went through `progress.rs`. Their look is genuinely different — no prefix, their own tick sets and colours, `println` through the bar, `finish_and_clear` — so converting them would mean either changing what the user sees or growing the event model to carry per-call-site styling for two call sites. Both are recorded in `.claude/architecture.md` with a grep to check the boundary holds.

## Stability

The event model is intentionally **not** semver-stable: `publish = false`, `0.x` in lockstep with `icp-cli`, all enums `#[non_exhaustive]`, `TaskKind` closed rather than an open string. The event stream does **not** drive `--json` — `--json` keeps meaning the command's final result, and no flag or feature is added that would change that.

## Checks

`cargo build`, `cargo fmt --check`, `cargo clippy --workspace --all-targets` and `cargo test --workspace` are clean, locally and through the validation pipeline (review, test, document, lint). The integration suites need the test network launcher fixture; the `identity_link_hsm*` tests additionally need SoftHSM2 and the `canister_snapshot_*_resume` tests need `mitmproxy`. One integration assertion moved with the code: the `debug!` line that echoes step output now has the tracing target `icp::events` rather than `icp::progress`.

## Pipeline fixes folded in

Three findings the original change missed, fixed during validation:

- **`goldens-ansi-sensitive`** — the goldens above were captured with colour off and would have failed whenever the test binary's stdout was a terminal. `RecordingTerm` now strips SGR codes in both its normalized and raw views, with a test that feeds escaped frames through and asserts both.
- **`architecture-grep-false-hits`** — the invariant check documented in `.claude/architecture.md` matched `events::indicatif_reporter` too, so it listed 11 files while the prose claimed 3. Narrowed to `grep -rl 'use indicatif'`.
- **`sink-emit-repeated-lock`** — five arms of `IndicatifSink::emit` each repeated the same lock-and-look-up; collapsed into one `with_bar` helper, so "events for an unknown task are ignored" lives in one place.

## Follow-up: `--debug` output says which canister printed each line

A bug found after the inversion landed on this branch, fixed here rather than in a separate PR. Under `--debug` the bars are hidden, so the `debug!` line the sink logs for each step-output event was the only record of a build's output — and it carried just the line, so canisters building in parallel interleaved into an unattributable stream. The bars have always shown the name in their prefix, so the two paths disagreed.

`BarState` now keeps the prefix it gave the bar and the log line reuses it, so `[backend] compiling` reads the way the bar does. The prefix is kept at construction rather than read back off the bar. Byte tasks label their bars undecorated and keep doing so here; they report no step output today, so this settles the shape rather than changing anything. A line with no label, or one arriving for a task that has already finished, is still logged — bare — rather than dropped for want of a prefix.

Verified on a real parallel build, because interleaving is the whole point and a single-canister check would pass while proving nothing: five canisters under `--debug` produced 20 output lines, every one attributed, none misattributed, with the canister changing between all 19 consecutive pairs. Sink-level tests cover the interleaved case, the unlabelled and unknown-task cases, and the byte-task shape; the interleaving test was mutation-checked against the bare `debug!` it replaced.

Updating the integration assertions turned up a trap worth noting: `sync_tests` had a negative assertion that no sync step ran (`"DEBUG icp::events: syncing"` must not appear), which the new prefix would have made vacuously true — green while testing nothing. It now matches `"] syncing"`, a shape only the runtime log produces and which the `--debug` config dump does not.

The pipeline also fixed a flaw in the new test helper: it captured `debug!` lines through a thread-local subscriber, but `tracing` resolves and caches whether a callsite is enabled the first time *any* thread reaches it, so another test getting there first could cache "disabled" and leave the capture silent. It now installs one global subscriber per test binary and keys captured lines by thread id; `.claude/testing.md` records that constraint.


adamspofford-dfinity (approved):

crates/icp-events/src/cancel.rs
```diff
@@ -0,0 +1,332 @@
...
+
+use futures::future::{Either, select};
+
+/// A cooperative cancellation signal shared by everything a [`Reporter`](crate::Reporter)
+/// hands out.
+///
+/// Deliberately runtime-agnostic: it is a flag plus a waker list, so operations can
+/// wait on cancellation without pulling in an async runtime.
+#[derive(Debug, Clone, Default)]
+pub struct CancelToken {
```
raymondk, line 20: Nothing outside this crate uses `CancelToken`, `Cancelled`, `run_until`, `Reporter::with_cancel_token`, or `cancel_token()` — the module's own tests are its only consumers.

That's 332 lines including a keyed waker registry with drop-time deregistration, plus 10 tests, for machinery no operation exercises. The subtle parts (waker replacement on re-poll, the re-check under the lock) are exactly what rots when nothing calls them.

I'd land this with the increment that actually needs cancellation.

crates/icp-events/src/reporter.rs
```diff
@@ -0,0 +1,697 @@
...
+    pub fn with_cancel_token(sink: Arc<dyn EventSink>, cancel: CancelToken) -> Self {
+        Self {
+            sink,
+            next_id: Arc::new(AtomicU64::new(0)),
+            cancel,
+        }
+    }
+
+    /// A reporter that throws everything away.
+    pub fn discard() -> Self {
```
raymondk, line 54: `Reporter::discard()` and `DiscardSink` have no callers outside the crate. The two tests that cover them (`a_discarding_reporter_still_works` here, `discard_sink_keeps_nothing` in `sink.rs`) are both assertion-free.

crates/icp-events/src/reporter.rs
```diff
@@ -0,0 +1,697 @@
...
+        &self.cancel
+    }
+
+    /// Send one event straight through to the sink.
+    pub fn emit(&self, event: Event) {
+        self.sink.emit(event);
+    }
+
+    /// Emit a user-facing message that does not belong to any task.
+    pub fn notice(&self, level: NoticeLevel, message: impl Into<String>) {
```
raymondk, line 69: `notice`/`info`/`warn`/`error` — and `Event::Notice` / `NoticeLevel` behind them — have no callers. Nothing in the workspace constructs a `Notice`; `events.rs` has a receiving arm that the comment there describes as a placeholder for a separate work item.

Five methods, an enum, and two tests for a path that doesn't exist yet. Worth landing with the conversion that uses it.

raymondk (commented):
Review focused on test redundancy and comment noise. The design is sound and everything passes locally (`cargo test -p icp-events -p icp-cli --bins`: 40 + 105 pass; `cargo clippy --all-targets` clean). Inline comments cover the file-specific items; the cross-cutting ones are below.

## 1. A large part of the new crate has no callers

This is the biggest "unnecessary" item. Grepping the workspace outside `crates/icp-events/src`:

| API | Callers |
| --- | --- |
| `cancel.rs` — `CancelToken`, `Cancelled`, `run_until`, `Reporter::with_cancel_token`, `cancel_token()` | **none** |
| `Event::Notice` + `NoticeLevel` + `Reporter::notice/info/warn/error` | **none** — nothing constructs a `Notice`; the sink arm is an admitted placeholder |
| `Reporter::discard()` / `DiscardSink` | **none** |
| `Task::step_output` | **none** (call sites all use `task.output().line()`) |
| `RecordingSink::take` | **none** |
| `MAX_RECORDED_LINES_PER_STEP` as `pub` | internal only |

`cancel.rs` is the costly one: 332 lines including a keyed waker registry with drop-time deregistration, plus 10 tests, none of which any operation exercises. Its subtleties (re-poll waker replacement, re-check under lock) are exactly the kind of thing that rots when unexercised.

My suggestion is to cut `cancel.rs` and the `Notice`/`discard` surface from this increment and land each with the change that needs it. If you'd rather keep the model complete up front, that's a defensible call — but then it's worth saying so in the PR body, because as it stands a reviewer can't tell "planned" from "orphaned".

## 2. Six copies of the same ordering test

`operations/build.rs`, `sync.rs`, `install.rs`, `candid_compat.rs`, `binding_env_vars.rs`, `settings.rs` all assert `task_labels(...) == [frontend, backend]`, each ~25 lines of call-site boilerplate. Plus `events.rs`'s `bars_are_drawn_in_the_order_the_canisters_were_given` and `reporter.rs`'s `task_ids_are_handed_out_in_creation_order` for the same property one layer down. The invariant is per-call-site so it isn't pure duplication, but eight tests for "don't move `reporter.task()` inside the async block" is a lot — two (one steps-op, one spinner-op) would cover the regression at a fraction of the size. Details inline on `operations/build.rs`.

## 3. The same fact explained three times

Two subtle facts each get explained in three places, which means three places to update when either changes:

- **ANSI/colour detection**: `strip_ansi`'s doc, `RecordingTerm::frames`' 15-line doc, and `normalization_strips_the_colour_codes_a_terminal_would_add`'s doc.
- **The `enable_steady_tick` race**: `IndicatifSink::start`, `a_spinner_is_labelled_before_its_first_tick`, and `.claude/architecture.md`.

One site should carry each explanation; the others can point at it.

Separately, this comment is duplicated verbatim in five files:

> `// Started up front so the tasks appear in the order the canisters were given, regardless of the order the futures below are first polled in.`

It's a good comment — it just doesn't need to be five comments. One in `test_support.rs` or the architecture doc, referenced from the call sites, would do.

## What's good

The frame-golden approach in `mod rendering` is the right call for a refactor that deletes its own baseline; capturing the literal pre-inversion output is far more convincing than eyeballing it. `step_replay.rs` and its tests are tight — five tests, no overlap, each pinning one formatting rule. The `Task` drop-closes-neutrally guarantee is the correct invariant and is tested at the right level. And the `--debug` canister-prefix fix is a genuine user-visible improvement that the CHANGELOG entry describes plainly.

crates/icp-events/src/reporter.rs
```diff
@@ -0,0 +1,697 @@
...
+
+        self.reporter.emit(Event::StepStarted {
+            id: self.id,
+            index,
+            title,
+        });
+    }
+
+    /// Report a line of output from the step in progress.
+    pub fn step_output(&self, line: impl Into<String>) {
```
raymondk, line 182: `step_output` has no callers outside this crate's tests — every real call site goes through `task.output().line()`. It's a one-line convenience wrapper; probably fine to drop until something wants it.

crates/icp-events/src/reporter.rs
```diff
@@ -0,0 +1,697 @@
...
+                _ => None,
+            })
+            .collect();
+        assert_eq!(lines, vec!["out".to_owned(), "err".to_owned()]);
+    }
+
+    /// The writer outlives the borrow of the task, since it is handed to code that
+    /// keeps reporting while the operation moves on.
+    #[test]
+    fn a_writer_can_be_sent_to_another_thread() {
```
raymondk, line 495: This test, and `sinks_are_usable_from_several_threads` in `sink.rs`, assert `Send`/`Sync` and that `Mutex` works. Both are compile-time properties — a `fn _assert_send<T: Send>() {}` plus a call site says the same thing without spawning threads, and fails at compile time rather than at test time.

crates/icp-events/src/sink.rs
```diff
@@ -0,0 +1,134 @@
...
+        Self::default()
+    }
+
+    /// A snapshot of everything recorded so far, in emission order.
+    pub fn events(&self) -> Vec<Event> {
+        self.events.lock().expect("recording sink poisoned").clone()
+    }
+
+    /// Drain everything recorded so far, leaving the recorder empty.
+    pub fn take(&self) -> Vec<Event> {
```
raymondk, line 55: `take` has no callers — `take_drains_the_recorder` is the only thing that exercises it.

crates/icp-events/src/sink.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn take_drains_the_recorder() {
+        let sink = RecordingSink::new();
+        sink.emit(notice("one"));
+
+        assert_eq!(sink.take(), vec![notice("one")]);
+        assert!(sink.events().is_empty());
+    }
+
+    #[test]
+    fn discard_sink_keeps_nothing() {
```
raymondk, line 100: No assertions, and it covers API (`DiscardSink`) that nothing calls. Between this and `a_discarding_reporter_still_works` in `reporter.rs` there are two assertion-free tests for the same unused type.

crates/icp-events/src/output.rs
```diff
@@ -0,0 +1,178 @@
...
+    event::{Event, TaskId},
+    reporter::Reporter,
+};
+
+/// How many of a step's output lines are kept for replay.
+///
+/// Some limit is needed: a runaway command can print without end. Once a step is
+/// over this, its oldest lines are dropped and the most recent ones survive, which
+/// is the half worth showing after a failure.
+pub const MAX_RECORDED_LINES_PER_STEP: usize = 10_000;
```
raymondk, line 16: `MAX_RECORDED_LINES_PER_STEP` is exported from `lib.rs` but only read inside `output.rs` and its tests. Could be private until something outside needs to reason about the cap.

crates/icp-events/src/event.rs
```diff
@@ -0,0 +1,229 @@
...
+                level: NoticeLevel::Info,
+                message: "hi".into()
+            }
+            .task_id(),
+            None
+        );
+    }
+
+    #[test]
+    fn notice_levels_are_ordered_by_severity() {
```
raymondk, line 225: This tests a derived `Ord` that nothing reads — no code compares `NoticeLevel`s. If the ordering isn't load-bearing, consider dropping the derive along with the test.

crates/icp-cli/src/events.rs
```diff
@@ -0,0 +1,1170 @@
...
+    ProgressStyle::default_bar()
+        .template("{prefix} [{elapsed_precise}] [{wide_bar:.cyan/blue}] {bytes}/{total_bytes} ({bytes_per_sec}, {eta})")
+        .expect("invalid progress bar template")
+        .progress_chars("#>-")
+}
+
+// Creates a progress bar style with a spinner that transitions to a final tick symbol
+// - end_tick: the symbol to display when the progress completes (success, failure, etc.)
+// - color: the color theme for the spinner and text
+fn make_style(end_tick: &str, color: &str) -> ProgressStyle {
```
raymondk, line 63: Three of the comments in and around this function just restate the code:

```rust
// Animation frames for the spinner - creates a rotating star effect   (l.21)
// Template format: "[prefix] [spinner] [message]"                     (l.64)
// Combine animation frames with the final completion symbol           (l.69)
```

The `format!` and `tick_strings` calls say all three things already. Carried over from `progress.rs`, so a good moment to drop them.

crates/icp-cli/src/events.rs
```diff
@@ -0,0 +1,1170 @@
...
+
+    ProgressStyle::with_template(&tmpl)
+        .expect("invalid style template")
+        // Combine animation frames with the final completion symbol
+        .tick_strings(&[TICKS, &[end_tick]].concat())
+}
+
+/// A fixed-capacity rolling buffer that always holds the last `capacity` items.
+#[derive(Debug)]
+struct RollingLines {
```
raymondk, line 75: `RollingLines` and the `VecDeque` cap inside `StepLog` (`icp-events/src/output.rs`) are two implementations of the same "keep the last N" behaviour, in two crates. Different caps and purposes, so not necessarily worth unifying — but worth a look before a third one appears.

crates/icp-cli/src/events.rs
```diff
@@ -0,0 +1,1170 @@
...
+        let multi = MultiProgress::with_draw_target(target);
+
+        Self {
+            multi,
+            bars: Mutex::new(HashMap::new()),
+        }
+    }
+
+    fn start(&self, id: TaskId, kind: TaskKind, label: Option<String>) {
+        // Every bar is fully styled and labelled *before* it can draw anything: only
```
raymondk, line 171: Two things here.

**The comment** is 10 lines covering three separate points (the ticker race, how the prefix is decorated, why it's kept on `BarState`). The ticker race in particular is also explained at length in `a_spinner_is_labelled_before_its_first_tick` and again in `.claude/architecture.md` — three copies of one subtle fact. One should carry it and the others point at it.

**The code**: both match arms repeat the `match &prefix { Some(p) => bar.with_prefix(p.clone()), None => bar }` dance and construct the same five-field `BarState`. Only the bar constructor, `styled_spinner`, and the steady tick actually differ — the prefix application and the struct literal could be hoisted out of the match.

crates/icp-cli/src/events.rs
```diff
@@ -0,0 +1,1170 @@
...
+                outcome,
+                message,
+            } => self.finish(id, outcome, message),
+
+            // User-facing notices go through `tracing`, which the CLI's `UserLayer`
+            // prints as product output. Converting the operations' `info!`/`warn!`/
+            // `error!` calls to notices is a separate work item; this arm is what
+            // will receive them.
+            Event::Notice { level, message } => match level {
+                icp_events::NoticeLevel::Warn => tracing::warn!("{message}"),
```
raymondk, line 337: The outer `_ => {}` on line 342 is forced by `#[non_exhaustive]`, fine. But this inner `_ => tracing::info!` is not — a new `NoticeLevel` would silently print as info. Worth spelling out `NoticeLevel::Info` so the wildcard sits alone and a new level is a compile error here rather than a silent misclassification.

crates/icp-cli/src/events.rs
```diff
@@ -0,0 +1,1170 @@
...
+
+    /// Drop the colour codes from a frame.
+    ///
+    /// `indicatif` only calls `ProgressStyle::set_for_stderr` when its draw target
+    /// *is* stderr, which a [`TermLike`] target never is, so the styled template
+    /// fields fall back to `console`'s stdout-based colour detection: run from a
+    /// terminal the frames arrive wrapped in SGR codes, run from a pipe they do not.
+    /// What is drawn does not depend on how the developer started `cargo test`, so
+    /// the codes come off before anything else looks at the frame.
+    fn strip_ansi(frame: &str) -> String {
```
raymondk, line 381: The ANSI / colour-detection rationale is explained three times: here in `strip_ansi`'s doc, again in `RecordingTerm::frames`' 15-line doc, and a third time in `normalization_strips_the_colour_codes_a_terminal_would_add`'s doc. All three say the same thing — a `TermLike` target is never stderr, so `console` falls back to stdout-based detection.

One of them should own the explanation and the other two reference it.

crates/icp-cli/src/events.rs
```diff
@@ -0,0 +1,1170 @@
...
+        let reporter = new_reporter(&term);
+        for name in ["frontend", "backend"] {
+            reporter.task(TaskKind::Spinner, name).skip("done");
+        }
+
+        assert_eq!(term.frames(), ["[frontend]   done", "[backend]   done"]);
+    }
+
+    /// Every frame a [`TaskKind::Bytes`] task draws, with the clock blanked out.
+    fn byte_frames(positions: &[u64], finish: bool) -> Vec<String> {
```
raymondk, line 637: Dead branch — both paths return `term.timeless_frames()`, so the early `return` does nothing:

```rust
if finish {
    task.succeed("done");
    return term.timeless_frames();   // identical to the tail expression
}
term.timeless_frames()
```

Collapses to `if finish { task.succeed("done"); }` followed by the single return. (The comment about reading frames while the task is alive is still true and still worth keeping — the task drops after the tail expression either way.)

crates/icp-cli/src/events.rs
```diff
@@ -0,0 +1,1170 @@
...
+        });
+
+        let bars = sink.bars.lock().unwrap();
+        assert_eq!(bars[&TaskId(0)].bar.message(), "second\n│ new\n└\n\n");
+    }
+
+    /// The bar is gone from the map by the time `finish` returns, so the message has
+    /// to be observed where it actually lands: on the terminal.
+    #[test]
+    fn a_neutral_finish_draws_its_message_before_closing_the_bar_out() {
```
raymondk, line 925: This overlaps `a_skip_draws_what_finish_with_message_drew` in `mod rendering`, which pins the exact frame (`["[backend]   Skipped (not an upgrade)"]`). This one asserts the strictly weaker "some frame contains both substrings".

Its only extra assertion, `live_bars == 0`, is already covered by `a_spinner_lives_from_start_to_finish`.

crates/icp-cli/src/events.rs
```diff
@@ -0,0 +1,1170 @@
...
+            "frames: {frames:?}"
+        );
+    }
+
+    /// Every `DEBUG` message emitted while `f` runs, in order.
+    ///
+    /// The step-output log line is the only record of a build's output under
+    /// `--debug`, where the bars are hidden, so what it says is worth asserting on
+    /// directly rather than through the bar it is standing in for.
+    fn captured_debug_lines(f: impl FnOnce()) -> Vec<String> {
```
raymondk, line 984: 77-line helper carrying ~25 lines of comment, inline in the middle of the rendering tests — and the same explanation is repeated in `.claude/testing.md`.

Worth moving into a test-support module so this file's test section stays about rendering. The global-subscriber constraint is real and the explanation is good; it just doesn't need to live here and in the testing doc both.

crates/icp-cli/src/events.rs
```diff
@@ -0,0 +1,1170 @@
...
+        });
+
+        assert_eq!(lines, ["no one to blame", "late line"]);
+    }
+
+    /// A byte task's label is undecorated on its bar, so it stays undecorated here
+    /// too. Byte transfers report no step output today, so this only fixes what the
+    /// shape would be rather than changing anything visible.
+    #[test]
+    fn a_byte_task_attributes_output_with_its_undecorated_label() {
```
raymondk, line 1137: Two tests worth reconsidering:

- **This one** tests a path whose own doc comment says "byte transfers report no step output today" — it fixes the shape of something that doesn't happen.
- **`events_for_an_unknown_task_are_ignored`** (l.1156) has no assertions, and `an_unlabelled_task_logs_its_output_unattributed` already emits a `StepOutput` for the unknown `TaskId(9)`.

crates/icp-cli/src/operations/build.rs
```diff
@@ -174,3 +200,170 @@ pub(crate) async fn build_many_with_progress_bar(
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::operations::test_support::{
+        EmptyArtifacts, bare_canister, recording_reporter, task_labels,
+    };
+    use icp::manifest::{BuildStep, script};
+    use icp_events::{Event, Outcome, TaskId};
+
+    fn pkg_cache() -> (camino_tempfile::Utf8TempDir, PackageCache) {
```
raymondk, line 213: `pkg_cache()` is defined identically here and in `sync.rs`, and `canister_id()` is defined identically in `install.rs`, `candid_compat.rs`, `binding_env_vars.rs`, and `settings.rs` — four copies of the same `Principal::from_text("ryjl3-...")`.

`test_support.rs` exists for exactly this and both would fit there.

crates/icp-cli/src/operations/build.rs
```diff
@@ -174,3 +200,170 @@ pub(crate) async fn build_many_with_progress_bar(
...
+                .as_deref()
+                .is_some_and(|m| m.starts_with("Failed to build canister:")),
+            "{message:?}"
+        );
+    }
+
+    /// Tasks appear in the order the canisters were given, whichever order their
+    /// builds happen to finish in.
+    #[tokio::test]
+    async fn tasks_are_started_in_the_order_the_canisters_were_given() {
```
raymondk, line 337: This test exists six times: here, `sync.rs`, `install.rs`, `candid_compat.rs`, `binding_env_vars.rs`, `settings.rs` — all asserting `task_labels(...) == [frontend, backend]`, each ~25 lines of call-site boilerplate. Plus `bars_are_drawn_in_the_order_the_canisters_were_given` (events.rs) and `task_ids_are_handed_out_in_creation_order` (reporter.rs) for the same property one layer down.

The invariant genuinely is per-call-site, so it's not pure duplication — but eight tests guarding "don't move `reporter.task()` inside the async block" is a lot of surface. Two of them (one steps-op, one spinner-op) would cover the regression at a fraction of the size.

The two-line comment above the `reporter.task()` call is likewise duplicated verbatim across five files.

crates/icp-cli/src/operations/install.rs
```diff
@@ -444,3 +444,117 @@ pub(crate) async fn install_many(
...
+                Some("database".to_string()),
+            ]
+        );
+        for id in 0..3 {
+            assert_eq!(outcome_of(&events, TaskId(id)).0, Outcome::Failure);
+        }
+    }
+
+    #[tokio::test]
+    async fn nothing_to_install_reports_nothing() {
```
raymondk, line 545: This, and `nothing_to_sync_reports_nothing` in `settings.rs`, can't fail: empty input means the loop body never runs, so no events. No regression is being guarded here.

crates/icp-cli/src/operations/settings.rs
```diff
@@ -563,3 +564,98 @@ mod tests {
         assert!(environment_variables_eq(&vars1, &vars2));
     }
 }
+
+#[cfg(test)]
+mod reporting_tests {
```
raymondk, line 569: A second test module next to the existing `mod tests` — the new tests could just go in `tests`. It's the only file in the PR that splits them.

crates/icp-cli/src/operations/sync.rs
```diff
@@ -192,3 +213,200 @@ pub(crate) async fn sync_many(
...
+                .is_some_and(|m| m.starts_with("Failed to sync canister:")),
+            "{message:?}"
+        );
+    }
+
+    /// Tasks appear in the order the canisters were given, whichever order their
+    /// syncs happen to finish in.
+    #[tokio::test]
+    async fn tasks_are_started_in_the_order_the_canisters_were_given() {
+        use crate::operations::test_support::task_labels;
```
raymondk, line 380: `task_labels` is imported inside the function body here, while every sibling test module imports it at the top with the rest of `test_support`. Worth making consistent.

crates/icp-cli/src/commands/build.rs
```diff
@@ -57,12 +58,13 @@ pub(crate) async fn exec(ctx: &Context, args: &BuildArgs) -> Result<(), anyhow::
...
 
-    build_many_with_progress_bar(
+    build_many(
         canisters_to_build,
         environment_selection.name(),
         ctx.builder.clone(),
         ctx.artifacts.clone(),
         &ctx.dirs.package_cache()?,
+        &indicatif_reporter(ctx.debug),
         ctx.debug,
```
raymondk, line 68: These two adjacent arguments are both derived from `ctx.debug` but mean opposite things — hide the bars vs. replay *more* output on failure. Same shape in `deploy.rs` and `commands/sync.rs`.

The rename to `all_step_output` helps at the definition; the call site still reads as `ctx.debug` twice, which is easy to get wrong when someone edits this later.

crates/icp/src/canister/script.rs
```diff
@@ -113,13 +113,13 @@ pub(super) async fn execute_commands(
             //
             // Stdout
             tokio::spawn({
-                // Clone the stdio sender for use in the stdout handling task
+                // Clone the stdio writer for use in the stdout handling task
```
raymondk, line 116: This comment (and the matching one on l.131) restates the line below it — `let stdio = stdio.clone();` already says it clones the writer for the task.

Pre-existing, but the PR touches these exact lines, so it's a free cleanup.

crates/icp-cli/src/operations/snapshot_transfer.rs
```diff
@@ -611,8 +598,8 @@ pub async fn download_blob_to_file(
             .mark_complete(chunk_offset, total_size);
         save_download_progress(progress, paths)?;
 
-        // Update progress bar to show frontier position
-        progress_bar.set_position(progress.blob_progress(blob_type).frontier);
+        // Report the frontier position
```
raymondk, line 601: `// Report the frontier position` restates `task.position(progress.blob_progress(blob_type).frontier)` directly below it. The companion comment on the initial position ("a resumed download starts partway in") earns its place because it explains *why*; this one doesn't add anything.

crates/icp-cli/src/events.rs
raymondk: Not unifying these two, and here is the reasoning so the decision is on the record rather than silently skipped.

They look alike but they are different things:

- `RollingLines` here is a **display window**: four lines, rewritten into the bar's message on every push, and its contents are what a reader sees *while* a step runs. Its size is a property of the terminal.
- the `VecDeque` inside `StepLog` is a **retention buffer**: ten thousand lines, never read until an operation fails, and read whole. Its size is a property of how much output is worth keeping to explain a failure.

Unifying them means exporting a ring-buffer type from `icp-events` (the crate whose whole point is that it depends on serde and futures and knows nothing about rendering) so that the CLI's four-line display window can share it — new public API in the model crate, to save about fifteen lines in the sink. The coupling costs more than the duplication.

Your "before a third one appears" caveat is the right trigger, though: if a third turns up, that is when the shared type earns its place. I have not tried to encode that anywhere — it would be a comment predicting the future — but it is here for whoever hits it.

crates/icp-cli/src/events.rs
raymondk: Reconciled with the request to remove `Event::Notice`/`NoticeLevel` rather than done mechanically: this whole arm is gone, so the inner wildcard you flagged no longer exists. Nothing is left to misclassify.

What remains at that spot is the outer `_ => {}` only, and it now says what it is for — `Event` is `#[non_exhaustive]`, so a variant added upstream has to compile here rather than break the build, and it draws nothing until someone handles it.

When the 29 user-facing `tracing` sites are converted and a notice shape comes back, the receiving arm comes back with it — and it will spell out each level explicitly, which is the point you were making.

crates/icp-cli/src/operations/sync.rs
raymondk: Resolved by deletion rather than by moving the import: this is one of the six near-identical `tasks_are_started_in_the_order_the_canisters_were_given` tests, and per your other comment I kept two — `build.rs` (a steps operation) and `install.rs` (a spinner operation) — and dropped the rest, this one included.

So there is no longer an inconsistent inline import here. The two survivors both import from `test_support` at the top of their module with everything else.

PR #713, opened 2026-08-14
feat: deploy lib

adamspofford-dfinity:


raymondk:
Done as part of this stack: https://github.com/dfinity/icp-cli/pull/776

PR #720, opened 2026-08-20
feat: Expose the project canister ID table to sync plugins

adamspofford-dfinity:
Lets plugins access canister ID mappings (regardless of dependency declarations).

---

### DIFF
--- a/crates/icp-sync-plugin/sync-plugin.wit
+++ b/crates/icp-sync-plugin/sync-plugin.wit
@@ -20,13 +20,14 @@ interface types {
         /// The canister's fully-qualified project key: the subproject it belongs
         /// to (a path relative to the app root) joined to its local name by a
         /// single colon, e.g. "services/open-accounts:backend". A canister
-        /// defined directly in the project being synced has no subproject prefix
-        /// and appears as its bare local name, e.g. "backend".
+        /// defined directly in the app root has no subproject prefix and appears
+        /// as its bare local name, e.g. "backend".
         ///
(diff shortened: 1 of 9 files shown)


crates/icp-sync-plugin/sync-plugin.wit
```diff
@@ -14,6 +14,24 @@ interface types {
         content: string,
     }
 
+    /// An entry in the project's canister ID mapping table: a canister name
+    /// and the textual principal it resolves to in the environment being synced.
+    record canister-id-entry {
```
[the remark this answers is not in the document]
adamspofford-dfinity: This is unreleased.

crates/icp-sync-plugin/sync-plugin.wit
```diff
@@ -14,6 +14,25 @@ interface types {
         content: string,
     }
 
+    /// An entry in the project's canister ID mapping table: a canister name
```
raymondk, line 17: is this always going to be the fully qualified name?
If a canister is in a subproject, it will receive its  name and sibling names as the short name or the fully qualified name?

raymondk (approved):
The plugin is getting access to `fully qualified name --> id`
It's fine for now, but I wonder if it's going to be an issue in practice because for a subproject will get its sibling name differently than if it was being run indepedently.

crates/icp-sync-plugin/sync-plugin.wit
adamspofford-dfinity: It gets both.

PR #723, opened 2026-08-20
feat: Key-value names for files

adamspofford-dfinity:
Currently, plugins can only accept an unstructured list of files and/or directories:

```
files:
  - config.json
  - a.txt
  - b.txt
dirs:
  - ./generated
```

This PR allows plugins to assign semantic meaning to files and directories with named fields for a key-value structure. Keys are arbitrary. Values can be either single entries or lists.

Looks like:

```yaml
type: plugin
files:
  config: config.json
  templates: [a.txt, b.txt]
directories:
  upload: ./generated
```

File and directory entries in the wit interface gain a `key: option<string>`. When a name contains a list, all the entries will have the same key. The old unstructured list form is valid and results in the `option` being unset.

---


docs/guides/managing-environments.md
```diff
@@ -173,8 +173,8 @@ canisters:
       steps:
         - type: plugin
           path: ./plugins/upload-assets.wasm
-          dirs:
-            - dist
+          files:
```
raymondk, line 176: This doesn't seem like it needed to change since the static site canister will still work with `dirs`

crates/icp-sync-plugin/sync-plugin.wit
raymondk, line 100: now I can't help but think this should have been called `params` instead of `fields`

raymondk (approved):

PR #726, opened 2026-08-23
feat: Add canister-metadata-section function

adamspofford-dfinity:
This PR adds a function to the sync plugin interface to fetch metadata sections from canisters. This is useful for type-checking canister calls by fetching `candid:service`. (Does this function need to be so specific instead of fetching state-tree content? Yes, because plugins should be usable from a canister host, and canisters cannot fetch the state tree, but rather get metadata sections by calling a management canister method only available to canisters.)


crates/icp-sync-plugin/src/runtime.rs
```diff
@@ -201,12 +213,124 @@ impl HostState {
...
+                let controllers_path: Vec<Label<Vec<u8>>> = vec![
+                    "canister".into(),
+                    Label::from_bytes(target.as_slice()),
+                    "controllers".into(),
+                ];
+                let cert = agent
+                    .read_state_raw(
+                        vec![metadata_path.clone(), controllers_path.clone()],
+                        target,
+                    )
```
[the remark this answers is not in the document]
adamspofford-dfinity: Controllers are not privileged

crates/icp-sync-plugin/src/runtime.rs
```diff
@@ -101,6 +104,87 @@ pub struct CallableCanisters {
...
+/// Ask the target's subnet to certify a metadata section, reporting only what
+/// the certificate proves.
+///
+/// The section path is requested together with `controllers`, because a
+/// metadata path proven absent is equally what a canister that was never created
+/// looks like — `controllers` is written at creation, so its presence is what
+/// separates the two. A canister with no module installed has no sections at
+/// all, which the certificate reports as an absent path under a canister that
+/// exists, and so as [`CertifiedSection::Absent`].
+async fn certified_metadata_section(
```
raymondk, line 124: can't this reuse `icp::operations::misc::fetch_canister_metadata` ?

raymondk: or: https://github.com/dfinity/icp-cli/blob/main/crates/icp/src/operations/wasm.rs#L12 ?

raymondk: .... it would introduce a cycle....

raymondk (approved):

PR #729, opened 2026-08-25
refactor: move the operations layer into the icp crate

raymondk:
**Stacked on #725** — base is `rk/events`, so review this diff on its own; it becomes a diff against `main` once #725 merges.

## Why

#725 removed the last reason the operations layer had to live in the binary. Before it, `crates/icp-cli/src/operations/` imported `crate::progress`, drove `indicatif` bars, and formatted user-facing strings — it *couldn't* move. Now those files import no renderer, no progress bar, and no line channel: they take a `Reporter` and return typed errors.

So this moves `crates/icp-cli/src/operations/` → `crates/icp/src/operations/` (~7k lines). The library owns the work — building, syncing, installing, creating, settings, snapshots, tokens, cycles — and the binary owns argument parsing and display.

## What

Mostly code motion. Three couplings pointed back at the binary and are resolved rather than carried across:

| Coupling | Resolution |
| --- | --- |
| `recover_cycles` read its wasm from the CLI's build-time `artifacts` module | takes the bytes as a parameter; `canister delete` passes `crate::artifacts::get_recover_cycles_wasm()` |
| `token::transfer` imported `FlexibleAccountId` from `commands::args` | moved to `icp::parsers` alongside the other value parsers; `commands::args` re-exports it so command signatures are unchanged |
| `WasmMemoryPersistenceOpt` derived `clap::ValueEnum` unconditionally | now `#[cfg_attr(feature = "clap", derive(...))]`, using `icp`'s existing optional `clap` feature |

`call_output.rs` deliberately stays in the binary — it formats and prints a call response, so it's presentation, not an operation. It moved to `crates/icp-cli/src/call_output.rs`.

Eight dependencies leave `icp-cli`'s manifest (`backoff`, `phf`, `wasmparser`, `flate2`, `tar`, `sha2`, `num-bigint`, `camino-tempfile`); `flate2`, `tar`, and `sha2` come back as dev-dependencies for `bundle_tests`.

## Verification

- `cargo test -p icp -p icp-cli -p icp-events -p icp-sync-plugin`: 24 suites pass. The 5 `identity_link_hsm*` failures are environmental (SoftHSM2 isn't installed on this machine) and fail identically on `main`.
- `cargo clippy --all-targets` clean; `cargo fmt` applied.
- `cargo check -p icp --no-default-features` passes — the library still builds without `clap`, so the optional-dependency boundary is real.
- `cargo check -p schema-gen` passes (the other `icp` consumer).
- `./scripts/generate-cli-docs.sh` produces no diff — no user-visible command surface changed.

## Notes for review

- The move rewrote `pub(crate)` → `pub` across the moved files, so operations' internals are now `pub` on an internal (`publish = false`) crate. Tightening that to only what commands actually call is a reasonable follow-up; I kept this PR to code motion.
- Worth reviewing as one commit — `git log --stat` shows it as renames, and `git diff -M` keeps it readable.

https://claude.ai/code/session_01QUZHfx2ng97WM2j9XoA3rr


marc0olo:
Reviewed this on its own diff against `rk/events`, as suggested.

**It is pure code motion.** I normalized every moved file — rewriting `icp::` → `crate::` and `pub(crate) ` → `pub ` — and diffed old against new. The residual delta is exactly the three couplings in your table, nothing else. Specifically:

- `recover_cycles_before_delete`'s new `recover_cycles_wasm: &[u8]` receives the same `crate::artifacts::get_recover_cycles_wasm()` bytes it used to reach for itself.
- The `#[cfg_attr(feature = "clap", …)]` boundary is real: `clap` being an optional dependency creates the implicit feature, `icp-cli` enables it via `features = ["clap"]`, and `cargo check -p icp --no-default-features` passes alongside the full workspace build.
- `FlexibleAccountId` moved verbatim; the `pub(crate) use` re-export keeps `commands::args::FlexibleAccountId` resolving, so no command signature moved.

Everything else is import rewrites and visibility widening. `generate-cli-docs.sh` and `generate-config-schemas.sh` both produce no diff, and `cargo doc -p icp` emits the same 8 warnings as `main` — so nothing broke in the intra-doc links, which was my main worry for a move this size. `clippy --all-targets` and `fmt --check` clean; full test suite green apart from environmental failures (`mitmproxy`, Docker daemon).

Three small things, none blocking:

**1. Stale path reference.** `crates/icp-sync-plugin/src/path.rs:11` still says the bundler lives at `crates/icp-cli/src/operations/bundle.rs`. It's the only reference to the old location left in the tree.

**2. `.claude/CLAUDE.md` overstates the boundary.** "Operations never print or draw" isn't quite true yet — `create.rs`, `settings.rs`, `install.rs`, `recover_cycles.rs` and `binding_env_vars.rs` still emit `tracing` `info!`/`warn!`/`error!` that surface as user-facing lines. That's pre-existing and arguably fine for a library, but as written the sentence reads as a stronger invariant than the code holds, and the next person to add an operation will trip over it. Something like "never draw progress or format results", plus a note that diagnostics still travel on `tracing`, would match reality.

**3. `WasmMemoryPersistenceOpt`'s doc comment** still opens with "CLI-facing choice for …" now that the type lives in the library. Cosmetic, but it's the single place where the `clap` coupling is visible, so it's the comment most worth spending a sentence on explaining *why* it's there.

Agree that the `pub(crate)` → `pub` sweep is right to leave for a follow-up rather than mix into a motion commit — the widening is what makes this diff reviewable as renames, and tightening it later is a separate, mechanical pass.

lwshang (approved):

PR #732, opened 2026-08-26
refactor: generic progress events, and deploy as a composite operation

raymondk:
Draft. An attempt at the design that came out of reviewing #730: instead of moving the task vocabulary from `icp-events` into the CLI, delete it — make the events describe the *shape* of a progress stream and let the operation supply the words. Then make `deploy` an operation that forwards the events of the operations it calls, which is what gives nesting a reason to exist.

Filed against `rk/events`, alongside #729 and #730. It supersedes #730 rather than building on it.

## Why

#730 makes `icp-events` generic in the task payload and moves `TaskKind` to `icp-cli`. That fixes the real defects — the two `_ =>` arms where a new phase silently rendered as a build step, and the two unreachable arms where a byte bar had no message slot. But it keeps a vocabulary and pays a type parameter for it, and it collides with #729: after operations move into `icp`, they can no longer name an `icp-cli` type.

Composing deploy settles that argument. Once every operation in the process shares one event stream, `Event<T>` has exactly one instantiation, forever — the generic buys nothing. So the choice is a vocabulary somewhere, or no vocabulary at all.

This takes the second. A task announces itself:

```rust
reporter.task(Task::new("Building").on(&canister.name))
reporter.task(Task::counter("WASM module", total_bytes).on(name))
reporter.task(Task::group("Installing canisters:"))
```

`icp-events` holds a title, an optional subject, a `Shape` (group / spinner / counter), and a `parent`. Nothing enumerates builds, syncs or installs; `candid` drops out of the crate. The renderer composes every line from those facts, so there is no dispatch left to forget to wire up.

|                           | today                            | #730                     | here                   |
| ------------------------- | -------------------------------- | ------------------------ | ---------------------- |
| add a task kind           | 6 `match` blocks, 2 crates       | 1 impl block, 1 file     | nothing                |
| forget to wire one up     | renders as a build step          | does not compile         | nothing to wire        |
| `icp-events` deps         | `candid`, `serde`, `tokio`       | `serde`, `tokio`         | `serde`, `tokio`       |
| generic machinery         | —                                | `Event<T>`, `StepSink`   | —                      |
| where the vocabulary sits | below `icp`                      | `icp-cli` (breaks #729)  | nowhere                |

## The composite

`commands/deploy.rs` was 973 lines opening **seven** renderers in sequence. That sequencing was the mechanism keeping `info!("Installing canisters:")` from tearing through a live progress view — and it didn't fully hold: `println!("Created canister … with ID …")` fires inside the create scope, racing the MultiProgress it prints over.

Now it is one reporter, one stream, one task tree. Each phase is a group task; the reporter handed to `build_many`/`install_many`/`sync_many` is scoped to that phase, so their tasks nest under it. **None of those operations changed** — they still take `&Reporter` and still start what they think are top-level tasks. That's the property worth having: composing an operation costs nothing at the operations it composes.

- Phase headings are events, ordered on the same stream as the bars they title. Nothing has to be awaited to keep them in place.
- Nothing in the operation writes to the terminal. `DeployReport` carries the created canisters and the command prints them, so stdout stays beside `--json`. The println race is gone structurally.
- The command drops to 476 lines: arguments, the `http_request` probe, URL printing, the JSON shape.

## Why this helps #713

#713 hit this problem one layer down and solved it with `trait StepProgress { fn line(&self, line: String) }` threaded as `Option<&dyn StepProgress>` through `PluginExecutor`, `ScriptRunner` and the fetch path — because a deploy library that runs inside a canister cannot own terminal output. That is the `Option<Sender<String>>` shape `rk/events` deleted, reinvented.

A composite operation reporting vocabulary-free events *is* that seam, and it already works for a host with no terminal: drop the receiver and every send is a no-op. The library emits; the CLI renders; a canister host ignores. No `Option<&dyn _>` parameter on every trait method, and no CLI wording inside the library.

## Deliberate output changes

Composing text instead of writing it per kind moves some wording. All of it is chrome:

- `----- Failed to build canister 'x' -----` → `----- Building failed: 'x' -----`
- `[x] Build output:` → `[x] Building output:`
- sync step headers read `Syncing: step 2 of 4 (assets)`, like every other multi-step task, instead of `\nSyncing: (assets) 2 of 4`
- the Candid check is titled `Checking Candid compatibility`
- build and sync show `Building...`/`Syncing...` before their first step reports, where they used to show nothing
- the settings phase gained the heading it never had (its bars used to appear under `Setting environment variables:`)

`build_tests` and one `deploy_tests` assertion cover the first two; everything else passes untouched, including all 24 deploy tests and `canister_snapshot_tests`. `icp-events` gains 7 tests including a pinned wire format — it had none.

## What I'd want a reviewer to push on

1. **Is uniform wording acceptable?** This is the premise, not a detail. Per-kind bespoke phrasing is exactly what a typed payload buys; if `Synced successfully: <id>` and friends should stay tunable per operation, #730 is the honest way to keep them and this PR is wrong.
2. **The two strings operations still supply**: `Succeeded { message }` and the Candid check's breaking-change report. The success message is there because only sync knows the canister id it synced to; the Candid report is diagnostic content, not chrome. Both are defensible, and both are prose in an operation.
3. **Interactive ordering.** The renderer closes the live view before each top-level heading, reproducing what tearing down a per-phase renderer used to do. Verified in non-TTY (where bars are hidden) by the deploy tests; the TTY behaviour I reasoned about rather than watched. Worth someone running `icp deploy` in a terminal.
4. **`FailureReport` has three variants and an epilogue.** Each maps to a real special case in the old `match` blocks, but it is the part of this design most likely to be over-fitted.

https://claude.ai/code/session_01EjqaKcyQKsUQN838QWnyRY


raymondk:
closed this in favor of:
- https://github.com/dfinity/icp-cli/pull/729
- https://github.com/dfinity/icp-cli/pull/730
- https://github.com/dfinity/icp-cli/pull/739

PR #734, opened 2026-08-26
feat: prune env-specific canisters from deploys even in workspaces

adamspofford-dfinity:
`icp deploy -e foo` should not deploy any canisters not listed in environment `foo`, but this does not work in workspaces. This PR fixes that.


crates/icp/src/project.rs
```diff
@@ -1063,39 +1231,93 @@ pub fn member_scoped_canisters(
...
             }
-            cs
         }
-    };
+    }
+
+    let mut cs: IndexMap<String, (PathBuf, Canister)> = canisters
+        .iter()
+        .filter(|(key, _)| match deciders.get(owning_prefix(key)) {
+            Some(sel) => sel.keys.contains(*key),
```
lwshang: The `None => true` arm just below means a project no list names contributes *all* of its canisters — so the root's own list no longer prunes a dependency that writes no list.

Repro: root declares `staging` with `canisters: []`, dependency `openemail` declares a bare `- name: staging`. Before this PR: empty environment. Now: `["openemail:backend", "openemail:frontend"]` — `icp deploy -e staging` creates and installs two canisters it used to skip.

There's also no syntax for the root to exclude such a dependency: naming canisters only adds them, and omitting them is a no-op. That makes the PR's stated goal — "`icp deploy -e foo` should not deploy any canisters not listed in environment `foo`" — unreachable from the root, which is the one manifest a workspace owner can always edit.

crates/icp/src/project.rs
```diff
@@ -1063,39 +1231,93 @@ pub fn member_scoped_canisters(
...
+        .unwrap_or_default();
+    selections.extend(&root_selection);
+    // Sorted by writer so a workspace with several colliding lists always reports
+    // the same pair.
+    selections.sort_by(|a, b| a.owner.cmp(&b.owner));
+
+    let mut deciders: BTreeMap<&str, &EnvSelection> = BTreeMap::new();
+    for sel in &selections {
+        for target in sel.targets() {
+            if let Some(first) = deciders.insert(target, sel) {
```
lwshang: This collides on any second list naming a project, including when both lists agree — and it fails `consolidate_manifest`, so the whole workspace stops loading for *every* command, not just deploy.

Repro: `a` and `b` both depend on `openemail`, and both write `canisters: [svc, "../openemail:backend"]` for `local`. `icp canister list`, `icp build`, and `icp deploy -e local` all fail. The error asks the user to drop one of the lists, but in a diamond over a vendored dependency that means editing a third-party manifest.

Two separable fixes: let identical selections through, and defer the error to the environment actually being used rather than manifest load.

crates/icp-cli/src/operations/settings.rs
```diff
@@ -245,11 +253,19 @@ pub(crate) async fn sync_settings_many(
...
-                        info.name
-                    );
+                    if env_canisters.contains(name) {
+                        warn!(
+                            "Controller canister '{name}' for '{}' has not been created yet; \
+                             it will be set as a controller once created.",
+                            info.name
+                        );
+                    } else {
+                        warn!(
```
lwshang: `sync_controller_dependents` didn't get the same treatment — its warning at settings.rs:366 is still the unconditional one. A controller pruned out of the environment is reported there as "has not been created yet; it will be set as a controller once created", which will never happen. The same `env_canisters` check applies.

lwshang (commented):
Read every hunk and verified the first two findings empirically in a throwaway worktree. The partition model reads well and the nested `<path>:<canister>` resolution is careful, but two cases behave worse than before the change. Details inline.

crates/icp/src/project.rs
adamspofford-dfinity: ~~Fixed lower in the stack, #737.~~

Working as intended per meeting, see 0e3d03c

crates/icp/src/project.rs
adamspofford-dfinity: Fixed in e41d0ed

crates/icp-cli/src/operations/settings.rs
adamspofford-dfinity: Fixed in 7ae2e61

raymondk (approved):

PR #742, opened 2026-08-27
feat: Add an upgrade-arg field

adamspofford-dfinity:


crates/icp-cli/tests/assets/echo_install_arg_canister/main.mo
```diff
@@ -0,0 +1,15 @@
+import Nat8 "mo:base/Nat8";
```
raymondk, line 1: Should be using the core library, it replaces base and we should stick to best practices

crates/icp-cli/src/commands/deploy.rs
raymondk, line 106: probably oversight from the refactor but it probably makes sense for this error to also be raised from the operation.

crates/icp/src/operations/deploy.rs
```diff
@@ -331,18 +334,36 @@ pub async fn deploy(
...
+                (Some(bytes), _) => Some(bytes.clone()),
+                (None, Some((field, a))) => Some(a.to_bytes().context(InstallArgsSnafu {
+                    canister: name,
+                    field,
+                })?),
+                (None, None) => None,
             };
 
-            Ok::<_, DeployError>((name.clone(), cid, mode, status, init_args_bytes))
+            Ok::<_, DeployError>((name.clone(), cid, mode, status, args_bytes))
```
raymondk, line 366: probably something for another PR but it would be nice to log for debug purposes, the canister, mode and which args were used (not necessarily the value, just where they came from)

raymondk (approved):

crates/icp/src/operations/bundle.rs
```diff
@@ -675,7 +675,8 @@ async fn prepare_canister(
     Ok(Item::Manifest(CanisterManifest {
         name: local.to_owned(),
         settings: localize_controllers(canister.settings.clone().into(), local_names),
-        init_args: canister.init_args.as_ref().map(convert_init_args),
+        init_args: canister.init_args.as_ref().map(convert_args),
+        upgrade_args: canister.upgrade_args.as_ref().map(convert_args),
```
lwshang, line 679: A canister-manifest `init_args`/`upgrade_args` `path:` isn't containment-checked, so bundling inlines a file from outside the project. Confirmed: `upgrade_args: {path: ../secret.txt, format: hex}` bundles fine, with the secret's contents in the bundled `icp.yaml` as `value:`.

The *environment*-override path is checked (`relocate_args_overrides` → `canonicalize_within_project`), and the canister-level env-var analogue is checked by `validate_env_var_files` — there's just no `validate_args_files` counterpart for these two fields.

Pre-existing for `init_args`; this line widens it to a second field. Fine to fix separately, but worth covering both while you're here.

raymondk: We discussed this:
- during bundling your build process is running arbitrary code so you can't really protect from reaching outside of the project directory.
- after the bundling the files will always be moved into the project and the sandbox is much tighter.

So we're keeping things as is for now.

PR #747, opened 2026-08-29
feat: new canister setting snapshot_visibility

raymondk:
> Stacked on #745. Review that one first — this sits on top of its shared visibility machinery and is deliberately small because of it.

## What

Adds the third and last of the management canister's visibility settings, `snapshot_visibility`, which controls who may read a canister's snapshots. It gates three endpoints — `list_canister_snapshots`, `read_canister_snapshot_metadata`, and `read_canister_snapshot_data` — which are what `icp canister snapshot list` and `icp canister snapshot download` call.

It takes the same three forms as the other two:

```yaml
settings:
  snapshot_visibility: controllers      # default
  # snapshot_visibility: public
  # snapshot_visibility:
  #   allowed_viewers: ["aaaaa-aa", "2vxsx-fae"]
```

and the same flag shapes on `icp canister settings update`:

```
--snapshot-visibility <controllers|public>
--add-snapshot-viewer / --remove-snapshot-viewer / --set-snapshot-viewer <PRINCIPAL>
```

including the rule from #745 that a relative viewer edit (`--add-*-viewer`, `--remove-*-viewer`) is refused while the setting is `public`, since `public` carries no allowed viewers list for it to be relative to.

`icp canister status` and `icp canister settings show` report it in both their default and `--json` output, it is applied at creation, and `icp canister settings sync` keeps it in sync. That replaces the two `snapshot_visibility: None` placeholders and their TODOs.

**Taking, restoring, and deleting snapshots stays controller-only** whatever this is set to. The setting governs reading alone.

## How

Nothing new was built. #745 left one `Visibility` enum, one manifest-newtype macro, one flag-resolution routine, one set-wise comparison, and one renderer, so a third setting is:

- one `candid_conversions!(SnapshotVisibility)` line;
- one `visibility_setting!` invocation for the manifest newtype;
- a third `VisibilityFlags` group, plus a field at each of the call sites that already handle the other two.

Manifest forms, parse-error wording, the sync comparison, the rendering, the `--json` shape, and the `public` refusal all come from the shared code unchanged.

## Access policy

Its rule follows [`log_visibility`](https://docs.internetcomputer.org/references/ic-interface-spec/management-canister/) rather than `status_visibility`: **`controllers` is exhaustive**, with no always-allowed callers on top — no subnet-administrator or read-your-own-state carve-out of the kind `status_visibility` has. The schema description and the reference say so per setting, as they already did for the other two.

The reference also flags something the other two settings do not carry: reading a snapshot exposes the canister's **full state** — WASM module, WASM memory, stable memory, and chunk store. Granting snapshot access is closer to granting a state dump than to granting the read-only report `status_visibility` covers, and the docs say that where someone choosing a value will see it.

## Testing

Unit tests aimed at what a mirror actually risks — a copy-paste leaving one group driving another setting:

- `each_group_reads_its_own_flags` — each of the three `*VisibilityOpt` groups reads its own four flags and names itself (`label`, `stem`), and an empty group drives nothing.
- `the_public_refusal_names_the_settings_own_flags` — each group's refusal names its own `--add-*-viewer`, `--set-*-viewer`, and `--*-visibility controllers`.

Both fail if a group is wired to another setting's fields, which is the mistake this shape invites. (A textual guard while writing it caught exactly that: the copied `--remove` help still pointed at `--status-visibility controllers`.)

Also: `icp::canister::visibility` covers `snapshot_visibility`'s parse error naming itself and its round trip through the Candid type, alongside the existing two.

Against a local network:

- `canister_settings_update_snapshot_visibility` (new) — every flag, checked both in `settings show` and in what it actually grants: a non-controller is refused `snapshot list`, then succeeds once added as a viewer or once the setting is public, and is refused again when it is revoked. The replica does enforce the setting, so these are real assertions rather than a stand-in. Also the `public` refusal, both warnings, and that an update naming no visibility group leaves all three untouched.
- `canister_settings_sync_log_visibility` → **renamed** `canister_settings_sync_visibility`, and extended: one manifest now drives all three settings through a single sync. That is the case that matters — any one change alone would satisfy the "settings already match" check that decides whether to send an update at all. It also drives `snapshot_visibility` through the `allowed_viewers` map form.
- `canister_settings_show` (extended) — `--json` carries the new field.

Full suites run locally: `canister_settings_tests` (14), `canister_snapshot_tests` (8), `canister_create_tests` (17), `canister_status_tests` (3), plus unit tests for both crates. `cargo fmt` and `cargo clippy --all-targets --workspace` clean; CLI docs and schemas regenerated.

## Not in this PR

`minimum_incoming_canister_call_cycles` is still passed as `None`. It arrived in the same dependency bump as `status_visibility` but is not a visibility setting and shares none of this machinery.


crates/icp-cli/src/commands/canister/settings/update.rs
```diff
@@ -429,6 +482,16 @@ pub(crate) async fn exec(ctx: &Context, args: &UpdateArgs) -> Result<(), anyhow:
             resolve_visibility(opt.flags(), current)
         })
         .transpose()?;
+    let snapshot_visibility = args
```
lwshang, line 486: Nit: the section comment eight lines up still reads `// Handle log and status visibility.` — it now heads three blocks while naming two.

lwshang (approved):
Clean mirror of the machinery #745 built — the parity holds everywhere I checked. `git grep -l` returns identical file sets for `status_visibility` and `snapshot_visibility`, and the sync path carries all three in both the `&&`-chained early-return guard and the update payload, so a snapshot-only change still triggers an update.

I cross-wired `stem: "snapshot"` to `"status"` to see whether the two new unit tests catch the copy-paste mistake this shape invites — both fail, as intended. Good guards.

Also confirmed the remaining `snapshot_visibility: None` in `create.rs` isn't a miss: that's the CLI-flag path where all three visibility settings are `None` under the pre-existing "should be configurable from the CLI" TODO. Manifest-driven creation does apply it.

The docs note that reading a snapshot exposes WASM module, WASM memory, stable memory, and chunk store is a genuinely useful addition — that's the thing someone choosing `public` here needs to know, and it doesn't carry over from the other two settings.

Verified locally on `82f3dc2a`: `check --workspace --all-targets`, `clippy --all-targets`, `fmt --check` clean; 302 `icp` lib tests and 54 bin tests pass; CLI docs and schemas regenerate with no drift.

One cosmetic nit inline — approving regardless. Note this is based on `rk/feat-status-visibility`, so #745 needs to land first.

PR #760, opened 2026-09-04
chore: Use the correct target dir in tests

adamspofford-dfinity:


marc0olo (approved):
Verified the mechanism: `cargo metadata` in `examples/icp-sync-plugin` (its own workspace, excluded from the root) resolves the target dir correctly, the default path is identical to the previous hardcoded one, and `--no-deps` keeps it offline and lockfile-free. Consistent with how the official rust recipe locates the artifact.

Follow-up in a separate PR: the same hardcoded assumption is still in `examples/icp-rust/icp.yaml`, `examples/icp-sync-plugin/icp.yaml` and the docs snippets, and our docs pin `@dfinity/rust@v3.0.0`, which predates the equivalent fix in the recipe (landed in v3.1.0).

raymondk (approved):

PR #774, opened 2026-09-11
refactor: Abstract out canister calling from icp-project

adamspofford:
---


crates/icp-project/src/operations/deploy.rs
lwshang, line 1: 586401dd drops `{source}` on the grounds that the error still reaches `main` for a full chain, "because the concurrent callers short-circuit on it with `try_join_all`". `create_canisters` is the one that doesn't: it drains `FuturesOrdered` and keeps only the first error via `error.get_or_insert`. So when two canisters fail, the second's cause is printed nowhere — its only trace is the Display-only task line at L458.

`TaskOutcome::Failed` already carries `causes`, and `crate::error::causes` (added in #775) would fill it. `sync.rs` is the only call site populating it today.

raymondk (approved):

crates/icp-project/src/operations/deploy.rs
raymondk: as discussed we will fixed this after the pr stack is merged

PR #778, opened 2026-09-11
fix(deps): bump rust_decimal to drop the vulnerable rkyv 0.7 transitive dep

SafraNako:
## What

`Cargo.lock` carried `rkyv 0.7.46` (via `byte-unit` → `rust_decimal` →
`rkyv`), flagged by [RUSTSEC-2026-0235](https://rustsec.org/advisories/RUSTSEC-2026-0235.html): insufficient archive validation can cause out-of-bounds
reads in archives containing `Rc`/`Arc`.

`rust_decimal 1.42.1` pinned `rkyv = "^0.7.46"` exactly, so `rkyv` couldn't
be bumped on its own (`cargo update -p rkyv` fails to resolve). `rust_decimal
1.43.0` moved to `rkyv ^0.8.13` (well past the 0.8.17 fix) and, in this
workspace, doesn't pull `rkyv` in as a dependency at all anymore —
`cargo update -p rust_decimal --precise 1.43.0` removes `rkyv`, `bytecheck`,
`ptr_meta`, and their supporting crates from the lockfile entirely.
`byte-unit`'s `^1` requirement on `rust_decimal` already allows `1.43.0`, so
this is a lockfile-only change.

## Verified locally

`cargo test --workspace --lib --bins` → **55 tests passed**, 0 failed.
(Also had to install the `wasm32-unknown-unknown` target this workspace's
`build.rs` needs for a canister crate — unrelated to this fix, just missing
in my environment.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,17 +30,6 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.7.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
-dependencies = [
- "getrandom 0.2.17",
- "once_cell",
- "version_check",
-]
-
-[[package]]
-name = "ahash"
 version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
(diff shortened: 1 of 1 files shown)


SafraNako:
Hi, checking in on this one. The fix removes a vulnerable rkyv 0.7 dependency (out of bounds read) by bumping rust_decimal instead of touching rkyv directly, since that pulled rkyv out of the graph entirely. I ran the full workspace test suite locally and 55 tests passed. It has been quiet for about a week, so wanted to see if you had a chance to look at it yet, or if there is anything I should change. Thanks in advance.

lwshang (dismissed):

